PR #13, opened 2026-04-09
Intermediate commit. In progress

slava-vasianovych:
Added basic layout. 
This implementation gives Login and Session commands only. 
Login gives ability to get the token and set Session context as authorized. While Session command gives ability to see the session state - logged in or not, and if logged in - with what user name. 
Still in progress.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,2 @@
+.idea
+target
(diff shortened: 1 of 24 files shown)


andrew-appko (commented):
Looks good, have you considered porting over the openapi yml to autogenerate the REST client?

PR #39, opened 2026-08-10
Declare jackson-datatype-jsr310 in connect-library

shawn-tsanet:
One dependency, proposed for Slava's review: `connect-library` declares `jackson-datatype-jsr310` so the client is self-sufficient outside Spring Boot. Alignment-agenda condition 5.

**Why.** The client relies on `ObjectMapper.findAndRegisterModules()`, which registers only what is on the classpath. Inside Boot the starter supplies the java.time module invisibly; in any other embedding nothing does, and serialization of `java.time` types fails. Found live during the Connect_Gateway engine extraction: 3 of 9 BETA contract tests failed outside Boot for exactly this, and passed 9/9 once the module was on the classpath.

**Receipts.**
- Red: a scratch consumer on the library's exact pre-fix runtime classpath (`dependency:build-classpath`, zero jsr310 entries) throws `InvalidDefinitionException: Java 8 date/time type java.time.OffsetDateTime not supported by default: add Module "com.fasterxml.jackson.datatype:jackson-datatype-jsr310"` - Jackson names this module verbatim.
- Green: same probe on the post-fix classpath serializes; the module arrives transitively.
- Full library build with codegen: exit 0. Module suite: 81 tests, 0 failures, 0 skips.
- Wire format is unaffected: the generated `ApiClient` pins `RFC3339DateFormat` (`invoker/ApiClient.java`, "Use RFC3339 format for date and datetime"), so dates on the wire stay RFC3339 regardless of mapper defaults; this change only fixes mapper construction on plain classpaths. End-to-end receipt: the extracted engine ran the live BETA suite 9/9 with this module declared.

**Version** is deliberately omitted: managed by the Spring Boot 3.3.4 BOM via `tsanet-client-parent`, matching the version-less `jackson-databind` declaration above it. Compile scope is deliberate too - it must flow transitively to consumers.

**Base is `oauth`**, the validated stack (`main@332e5c7` does not build standalone; the oauth tip is the pin the Connect_Gateway v0.1.0 certification ran against).

### DIFF
--- a/connect-library/pom.xml
+++ b/connect-library/pom.xml
@@ -32,6 +32,14 @@
             <artifactId>jackson-databind</artifactId>
         </dependency>
         <dependency>
+            <!-- The client relies on findAndRegisterModules(); outside Spring Boot nothing
+                 else puts the java.time module on the classpath, and date serialization
+                 fails with InvalidDefinitionException. Declared here so every consumer
+                 gets it, not only Boot apps. -->
+            <groupId>com.fasterxml.jackson.datatype</groupId>
+            <artifactId>jackson-datatype-jsr310</artifactId>
+        </dependency>
+        <dependency>
             <groupId>org.springframework</groupId>
             <artifactId>spring-jdbc</artifactId>
         </dependency>


shawn-tsanet (commented):
**Verdict:** `LOOKS GOOD`

**Minimal-Diff Rule check:** Single hunk, `connect-library/pom.xml` lines 34–41: adds a `<dependency>` block for `jackson-datatype-jsr310` with an explanatory comment. Traces directly to the PR title and body ("`connect-library` declares `jackson-datatype-jsr310`"). No incidental hunks, no reformatting, +8/−0. Clean.

**Correctness:**
- The dependency is declared without a `<version>`, relying on BOM management. The PR asserts `tsanet-client-parent` imports the Spring Boot 3.3.4 BOM and that the version-less `jackson-databind` above it works the same way. The diff shows the version-less `jackson-databind` sibling (lines 31–33), which corroborates that version management is in effect for this groupId family — jsr310 is BOM-managed by the same Spring Boot BOM. I have not opened the parent POM this session to confirm the BOM import, but the adjacent version-less `jackson-databind` in the same file is strong in-diff evidence.
- Trace: a non-Boot consumer builds `ObjectMapper`, calls `findAndRegisterModules()`, now finds `JavaTimeModule` on the classpath, and serializes `OffsetDateTime` instead of throwing `InvalidDefinitionException`. The red/green probe pair in the PR (Jackson naming the module verbatim in the error, then serializing post-fix) exercises exactly this path.
- Scope of the dependency defaults to `compile` (none specified), which the PR intends so it flows transitively. Correct for the stated goal — a `provided`/`test` scope would defeat the purpose. Good that the PR called scope out explicitly.
- Wire-format regression concern is pre-empted: `ApiClient` pins `RFC3339DateFormat`, so adding the module changes mapper construction on bare classpaths but not the on-wire date format. This is the right thing to have checked, since adding `JavaTimeModule` can otherwise flip serialization to numeric timestamps by default.

**Security:** None. No secrets, no auth, no external input handling. Adding a well-known Jackson module from the same coordinate family already present.

**Regression risk:** Low. The only behavioral change is inside Boot apps that already had this module transitively — the declaration is now explicit rather than implicit, which BOM management keeps at the same version, so no version skew. Consumers outside Boot gain the module (the intended fix). The `findAndRegisterModules()` call is idempotent, so a module already present via the Boot starter registered once and still registers once. The build receipt (81 tests, 0 failures; full library build exit 0; live BETA 9/9) covers the consumer surface. One edge worth noting but not holding: any consumer that deliberately avoided `JavaTimeModule` and relied on its absence would now get it transitively — but that would already have been the case under Boot, so the exposure is only new for bare-classpath embeddings, which is precisely the population this fix targets.

**Recommendation:** Merge — the one hunk traces cleanly, the fix is minimal and well-receipted, and the wire-format edge was checked. (Draft against `oauth`; that's Shawn's/Slava's merge call, not mine.)

---

shawn-tsanet:
Good to merge

Issue #37, opened 2026-08-10
testSubmission hardcoded true in DefaultTsaNetApiSession; production consumers cannot create real cases


shawn-tsanet:
> Re-filed from `shawn-tsanet/connect-sdk-demo#6`, where it was parked as an upstream issue per that repo's no-fork-fixes rule. Observed on the `oauth` branch at `0f57facd3326`, the stack the demo consumes verbatim.

## Problem

`DefaultTsaNetApiSession` sets the test flag unconditionally just before the gateway call, so every collaboration case created through the SDK facade is a test submission:

`connect-library/src/main/java/com/tsanet/api/internal/DefaultTsaNetApiSession.java:361`

```java
form.setTestSubmission(true);
```

Correct default for a demo. Wrong for any production consumer: the facade offers no way to create a real case.

## Why it matters

Production consumers of this stack (beyond the demo) are being scoped, and a production consumer must be able to create real member cases.

## Proposed fix

Expose the flag on the create-case path instead of hardcoding it. Preference: an explicit per-call parameter on the facade create API, with no silent default, so a demo passes `true` and a production consumer passes `false` deliberately. A session/config-level option is acceptable if a per-call parameter does not fit the facade design.

## Acceptance

- A facade consumer can create both test-flagged and real cases.
- Demo behavior preserved by passing `true` explicitly.
- Note the read side uses a different field name (`testCase`); surfacing it is the sibling issue (filed alongside this one).

Context: the gateway engine in `tsanetgit/Connect_Gateway` deliberately uses the public generated client (`com.tsanet.api.generated.*`) rather than the facades partly because of this pair; fixing them mainly serves the custom-integration lane.

shawn-tsanet:
Sibling issue (read side): `tsanetgit/Connect_SDK#38`.

shawn-tsanet:
Fixed on the oauth branch by `tsanetgit/Connect_SDK#41` (the test flag is a per-call parameter on the facade create API) and delivered to main in `tsanetgit/Connect_SDK#44` (2026-08-11, `66f44fe`); shipped in `v0.1.0`. Closing as completed. GitHub did not auto-close this because #41 merged into the oauth branch rather than the default branch.

PR #40, opened 2026-08-10
Surface testCase through the facade DTO and cache

### LINKED ISSUE #38: Facade drops testCase; consumers cannot distinguish test cases from real ones
> Re-filed from `shawn-tsanet/connect-sdk-demo#7`, where it was parked as an upstream issue per that repo's no-fork-fixes rule. Observed on the `oauth` branch at `0f57facd3326`, the stack the demo consumes verbatim.

## Problem

The Connect OpenAPI spec returns `testCase: boolean` on `CollaborationRequestStatusDTO`, and the generated client model parses it (`connect-library/target/generated-sources/openapi/connect-api/src/main/java/com/tsanet/api/generated/model/CollaborationRequestStatusDTO.java`). The hand-written facade record `CollaborationRequestStatusDto` (`connect-library/src/main/java/com/tsanet/api/connectapi/dto/CollaborationRequestStatusDto.java`) omits the field, so it is dropped at all three construction sites:

- `connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiCollaborationGateway.java:73`
- `connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiResponsesGateway.java:211`
- `connect-library/src/main/java/com/tsanet/api/storage/CollaborationRequestRepository.java:91` (so the flag is not cached either)

A facade consumer cannot tell a test case from a real one.

## Why it matters

A production consumer must badge and filter test cases. This bug class has bitten before: the Connect API writes `testSubmission` but reads back `testCase`, and (as seen in the `tsanetgit/Zendesk_App` work) filtering on the write-side name silently treats test cases as real.

## Proposed fix

Add `testCase` to the facade DTO, populate it at the three construction sites above, and persist it in the repository cache.

## Acceptance

- `testCase` is present on facade reads (list and detail) and round-trips through the SQLite cache.

Sibling issue: `tsanetgit/Connect_SDK#37` (make `testSubmission` controllable on the write side).

shawn-tsanet:
Implements #38: `testCase` now travels from the generated model through the facade DTO and round-trips the SQLite cache. Base is `oauth` for the same reason as #39. Closes #38.

## Design (two decisions for your sign-off)

1. **Record compatibility.** `CollaborationRequestStatusDto` gains `testCase` as an 11th component plus a `@Deprecated` 10-arg compatibility constructor that delegates null (null = unknown, never "real"). Chosen over a positional-only change because the record has ~42 call sites across three modules; the overload keeps every legacy site compiling and keeps each of the sibling condition-PRs' diffs to production code. All three production construction sites now use the 11-arg form; only tests ride the overload. The deprecation javadoc promises removal at the next release boundary, which naturally rides the first tagged release (alignment-agenda condition 2) - say the word if you want it tracked as its own issue.
2. **Refresh semantics: demote-to-unknown.** The upsert's `DO UPDATE SET` arm sets `test_case = excluded.test_case` unconditionally, so a refresh carrying null overwrites a known value back to "unknown". That is consistent with null-means-unknown, and it is last-write-wins like every other column; a `COALESCE` (sticky) variant would be a semantics change. Flagging it so the choice is yours, not implicit. In practice API rows always carry the flag, so null arrives only from pre-upgrade cache rows and legacy constructors.

## Migration behavior

The cache is a durable per-account SQLite file and `CREATE TABLE IF NOT EXISTS` is a no-op on it, so the column arrives via the repo's existing `ensureColumn` evolution mechanism (same pattern as `webhook_subscription.secret`). Existing rows read null until refetched; a refreshed row backfills. Physical column order differs between fresh (before `fetched_at`) and upgraded (appended) databases - verified no `SELECT *` or column-list-free access exists, everything names columns.

## Receipts

- Full reactor green: 119 tests (connect-library 83, integration-app 30, demo 6), zero failures - which is also the executable proof that all ~42 legacy call sites compile via the overload.
- **Red probe on the subtle half:** removing the `DO UPDATE SET test_case` arm fails exactly the frozen-DDL upgrade test (`failures=1`) - without that arm, a pre-upgrade row refreshed from the API keeps NULL forever, invisible to fresh-DB tests.
- The upgrade test's old DDL is a frozen verbatim copy of the pre-change `COLLABORATION_REQUEST_TABLE` (diffed against `origin/oauth`, normalized: identical) - deriving it from current constants would pass by construction.
- First test run caught a real defect: the SQLite driver's typed `getObject(col, Integer.class)` throws "Bad value for type Integer" on NULL; reads now use `getInt` + `wasNull()`.
- Non-test construction sites enumerated by command: exactly the three in this diff.

Sibling PR: #39 (jsr310). The write-side flag (#37) lands separately.

### DIFF
--- a/connect-library/src/main/java/com/tsanet/api/connectapi/dto/CollaborationRequestStatusDto.java
+++ b/connect-library/src/main/java/com/tsanet/api/connectapi/dto/CollaborationRequestStatusDto.java
@@ -13,6 +13,30 @@ public record CollaborationRequestStatusDto(
     Long receiveCompanyId,
     String token,
     String createdAt,
-    String updatedAt
+    String updatedAt,
+    Boolean testCase
 ) {
+    /**
+     * Compatibility constructor from before {@code testCase} existed; leaves it null,
+     * which readers must treat as "unknown", never as "real case".
+     *
+     * @deprecated pass {@code testCase} explicitly; scheduled for removal at the next
+     *     release boundary.
+     */
+    @Deprecated
+    public CollaborationRequestStatusDto(
+        Long id,
+        String status,
+        String summary,
+        String submitCompanyName,
+        Long submitCompanyId,
+        String receiveCompanyName,
+        Long receiveCompanyId,
+        String token,
+        String createdAt,
+        String updatedAt
+    ) {
+        this(id, status, summary, submitCompanyName, submitCompanyId, receiveCompanyName,
+            receiveCompanyId, token, createdAt, updatedAt, null);
+    }
 }
--- a/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiCollaborationGateway.java
+++ b/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiCollaborationGateway.java
@@ -80,7 +80,8 @@ public class ConnectApiCollaborationGateway {
             dto.getReceiveCompanyId(),
             dto.getToken(),
             formatDateTime(dto.getCreatedAt()),
-            formatDateTime(dto.getUpdatedAt())
+            formatDateTime(dto.getUpdatedAt()),
+            dto.getTestCase()
         );
     }
 
--- a/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiResponsesGateway.java
+++ b/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiResponsesGateway.java
@@ -218,7 +218,8 @@ public class ConnectApiResponsesGateway {
             dto.getReceiveCompanyId(),
             dto.getToken(),
             formatDateTime(dto.getCreatedAt()),
-            formatDateTime(dto.getUpdatedAt())
+            formatDateTime(dto.getUpdatedAt()),
+            dto.getTestCase()
         );
     }
 
--- a/connect-library/src/main/java/com/tsanet/api/storage/CollaborationRequestRepository.java
+++ b/connect-library/src/main/java/com/tsanet/api/storage/CollaborationRequestRepository.java
@@ -28,8 +28,9 @@ public class CollaborationRequestRepository {
             """
             INSERT INTO collaboration_request (
                 id, status, summary, submit_company_name, submit_company_id,
-                receive_company_name, receive_company_id, token, created_at, updated_at, fetched_at
-            ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
+                receive_company_name, receive_company_id, token, created_at, updated_at,
+                test_case, fetched_at
+            ) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)
             ON CONFLICT(id) DO UPDATE SET
                 status = excluded.status,
                 summary = excluded.summary,
@@ -40,6 +41,7 @@ public class CollaborationRequestRepository {
                 token = excluded.token,
                 created_at = excluded.created_at,
                 updated_at = excluded.updated_at,
+                test_case = excluded.test_case,
                 fetched_at = excluded.fetched_at
             """,
             requests,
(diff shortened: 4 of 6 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
- `CollaborationRequestStatusDto.java` — adds `testCase` 11th component + `@Deprecated` 10-arg constructor. Traces to #38 (surface testCase through DTO) and design decision 1. ✓
- `ConnectApiCollaborationGateway.java` / `ConnectApiResponsesGateway.java` — both `toDto` sites now pass `dto.getTestCase()`. These are the production construction sites; traces to "travels from the generated model through the facade DTO." ✓
- `CollaborationRequestRepository.java` — INSERT column list + placeholder, DO UPDATE SET arm, SELECT column lists (×2), mapRow, two new helpers. All trace to "round-trips the SQLite cache." ✓
- `DatabaseInitializer.java` — DDL column + `ensureColumn` call. Traces to migration behavior. ✓
- Test file — two new tests. Trace to the receipts. ✓

**Correctness:**
Traced the upgrade path (the subtle one). A pre-upgrade DB has `collaboration_request` without `test_case`. `createSchema` runs `CREATE TABLE IF NOT EXISTS` (no-op on existing table) then `ensureColumn(..., "test_case", "INTEGER")` which appends the column NULL. Existing rows read NULL → `readStoredBoolean` returns `null` via `wasNull()`. A refresh `saveAll` for row 1 with `testCase=true` hits the `ON CONFLICT DO UPDATE SET test_case = excluded.test_case` arm, storing `1`. `findAll` returns row 1 = TRUE, row 2 (untouched) = null. Matches the test assertion at test lines. ✓

The `getObject(col, Integer.class)` → `getInt` + `wasNull()` fix is correct: SQLite's typed getObject throwing on NULL is a real driver behavior, and `getInt` returns 0 on NULL which `wasNull()` disambiguates. Note: if a stored value were somehow 0 but non-null, `raw != 0` → false, correct. ✓

`toStoredBoolean` maps null→null, true→1, false→0; `setObject(11, null)` binds SQL NULL. Round-trip verified by `itRoundTripsTestCaseThroughTheCache` covering both false and true. ✓

Edge case worth naming (not a bug): the demote-to-unknown refresh semantics (design decision 2) means a null-carrying refresh overwrites a known value back to NULL. The author flagged this explicitly and it's last-write-wins consistent with every other column. The PR notes API rows always carry the flag, so null only arrives from legacy paths. This is a documented design choice, not a defect.

**Security:**
None. No secrets, no injection surface. The frozen-DDL test builds SQL from string literals, not external input. Production INSERT/SELECT use parameterized placeholders throughout.

**Regression risk:**
- Positional contract change: the INSERT went from 11 to 12 placeholders and `fetched_at` moved from index 11 to 12. The setter block was updated in lockstep (`setObject(11, ...)`, `setString(12, fetchedAt)`). ✓
- Physical column order differs between fresh (test_case before fetched_at) and upgraded (test_case appended last) DBs. PR claims no `SELECT *` exists — the two SELECTs in this file name columns explicitly, so ordinal drift is harmless. I did not independently grep the rest of the codebase for `SELECT *` against this table, but both SELECTs modified here are column-named and mapRow reads by column name, not ordinal.
- The `@Deprecated` 10-arg constructor keeps the ~42 legacy call sites compiling; the PR asserts full-reactor-green (119 tests) as the compile proof, and states production sites are exactly the three in this diff. I take the green build as the receipt for compilation; I can't independently confirm the "exactly three production sites" count from the diff alone, but the two gateways + the enumeration claim are consistent.

**Operational:**
None. Cache is a durable per-account SQLite file; column arrives via `ensureColumn` (same pattern as `webhook_subscription.secret`, per PR). No new env var, no dependency, no schema.sql in this repo (this is the Java connect-library). No ops-sidecar restart implicated.

**Recommendation:** Merge — both flagged design decisions (deprecation-removal timing, demote-to-unknown refresh) are yours to confirm, but neither is a defect and neither holds the merge.

---

shawn-tsanet:
Good to merge

PR #41, opened 2026-08-10
Make the test flag a per-call parameter on the facade create API

### LINKED ISSUE #37: testSubmission hardcoded true in DefaultTsaNetApiSession; production consumers cannot create real cases

shawn-tsanet:
Implements #37: the test flag becomes a per-call parameter on the facade create API, replacing the hardcoded `form.setTestSubmission(true)` that made every facade-created case a test submission. Base is `oauth`, same as #39/#40. Closes #37.

## Design

- Both `createRequest` signatures on `CollaborationRequestsFacade` gain `boolean testSubmission` with **no default**: every case that is not test-flagged pages a real engineer at a real partner, so the choice is per-call and deliberate (the issue's stated preference).
- The old signatures remain as `@Deprecated(forRemoval = true)` default methods delegating `true` - the pre-existing always-test behavior, preserved loudly. External consumers (`shawn-tsanet/connect-sdk-demo`) keep compiling; their migration to explicit `true` is the named follow-up, and the removal rides the first tagged release (alignment-agenda condition 2), same schedule as #40's DTO compat constructor.
- Both in-repo callers (CLI `CreateRequestExecutor`, demo Acme scenario) now pass `true` explicitly - zero behavior change. Whether the CLI should grow a production flag (`--production`?) is deliberately NOT in this PR - your call on whether that is wanted at all.

## Receipts

- Full reactor 120 tests green (connect-library 84 incl. 3 new, integration-app 30, demo 6).
- **The flag is asserted on the outgoing wire form** (captured gateway argument), not on a request object proxy: `false` reaches the wire, `true` reaches the wire, and the deprecated overload still sends `true`. Read-back through `testCase` (the read-side name) requires #40's facade change, so the wire-form capture is the strongest check available here; noted rather than stubbed.
- Red probe: re-hardcoding `true` fails exactly `itSendsFalseWhenAProductionConsumerPassesFalse` (Failures: 1), then restored to green.
- Old-decision sweep: `setTestSubmission` appears exactly once in non-generated code (the parameterized line). Caller sweep: both real call sites updated; remaining grep hits are javadoc. Implementor sweep: `DefaultTsaNetApiSession` is the interface's only implementor, and the reactor compile is the mechanical proof the new abstract methods are fully implemented.
- Untested compat path, named: the 4-arg deprecated overload (by receiverCompanyId) has no dedicated test - it routes through `getCreateFormByCompanyId`, which needs materially more mocking; the 5-arg overload test covers the delegation pattern.

Siblings: #39 (jsr310), #40 (testCase read side). Together they close the write/read asymmetry end to end: a consumer can create a real case and can tell it apart afterwards.

### DIFF
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CreateRequestExecutor.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CreateRequestExecutor.java
@@ -37,7 +37,8 @@ public class CreateRequestExecutor {
                 caseNumber,
                 summary,
                 description,
-                customFieldValues
+                customFieldValues,
+                true
             );
             CollaborationRequestPrinter.printList(cliRunContext, "Created collaboration request", List.of(created));
             System.out.println(EntityPrinter.info(
--- a/TSANet-integration-demo/src/main/java/com/tsanet/application/scenarios/AcmeCreateCollaborationRequestsScenario.java
+++ b/TSANet-integration-demo/src/main/java/com/tsanet/application/scenarios/AcmeCreateCollaborationRequestsScenario.java
@@ -124,7 +124,8 @@ public class AcmeCreateCollaborationRequestsScenario implements IntegrationScena
             BETA_COMPANY_ID,
             caseNumber,
             summary,
-            description
+            description,
+            true
         );
         existing.add(created);
         log.info("Created id={} status={} token={}", created.id(), created.status(), created.token());
--- a/connect-library/src/main/java/com/tsanet/api/facade/CollaborationRequestsFacade.java
+++ b/connect-library/src/main/java/com/tsanet/api/facade/CollaborationRequestsFacade.java
@@ -28,20 +28,66 @@ public interface CollaborationRequestsFacade {
 
     List<CollaborationRequestFormDto> listStoredFormsForDocument(long documentId);
 
+    /**
+     * @param testSubmission whether the case is created test-flagged. There is no
+     *     default on purpose: every case that is not test-flagged pages a real
+     *     engineer at a real partner, so the choice is made per call, deliberately.
+     */
     CollaborationRequestStatusDto createRequest(
         long receiverCompanyId,
         String caseNumber,
         String summary,
-        String description
+        String description,
+        boolean testSubmission
     );
 
+    /**
+     * @param testSubmission whether the case is created test-flagged. See
+     *     {@link #createRequest(long, String, String, String, boolean)}.
+     */
     CollaborationRequestStatusDto createRequest(
         CollaborationRequestFormTemplateDto formTemplate,
         String caseNumber,
         String summary,
         String description,
-        Map<Long, String> customFieldValues
+        Map<Long, String> customFieldValues,
+        boolean testSubmission
     );
+
+    /**
+     * Compatibility form from before the flag was per-call; keeps the pre-existing
+     * behavior of always creating a TEST submission.
+     *
+     * @deprecated pass {@code testSubmission} explicitly; scheduled for removal at
+     *     the next release boundary.
+     */
+    @Deprecated(forRemoval = true)
+    default CollaborationRequestStatusDto createRequest(
+        long receiverCompanyId,
+        String caseNumber,
+        String summary,
+        String description
+    ) {
+        return createRequest(receiverCompanyId, caseNumber, summary, description, true);
+    }
+
+    /**
+     * Compatibility form from before the flag was per-call; keeps the pre-existing
+     * behavior of always creating a TEST submission.
+     *
+     * @deprecated pass {@code testSubmission} explicitly; scheduled for removal at
+     *     the next release boundary.
+     */
+    @Deprecated(forRemoval = true)
+    default CollaborationRequestStatusDto createRequest(
+        CollaborationRequestFormTemplateDto formTemplate,
+        String caseNumber,
+        String summary,
+        String description,
+        Map<Long, String> customFieldValues
+    ) {
+        return createRequest(formTemplate, caseNumber, summary, description, customFieldValues, true);
+    }
 
     CollaborationRequestStatusDto fetchRequestByToken(String caseToken);
 
--- a/connect-library/src/main/java/com/tsanet/api/internal/DefaultTsaNetApiSession.java
+++ b/connect-library/src/main/java/com/tsanet/api/internal/DefaultTsaNetApiSession.java
@@ -308,10 +308,11 @@ final class DefaultTsaNetApiSession implements TsaNetApiSession, AuthFacade, Col
         long receiverCompanyId,
         String caseNumber,
         String summary,
-        String description
+        String description,
+        boolean testSubmission
     ) {
         CollaborationRequestFormTemplateDto template = getCreateFormByCompanyId(receiverCompanyId);
-        return createRequest(template, caseNumber, summary, description, Collections.emptyMap());
+        return createRequest(template, caseNumber, summary, description, Collections.emptyMap(), testSubmission);
     }
 
     @Override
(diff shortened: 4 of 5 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `CreateRequestExecutor.java` (+`true`): traces to "both in-repo callers now pass `true` explicitly." ✓
- `AcmeCreateCollaborationRequestsScenario.java` (+`true`): the demo Acme scenario caller. ✓
- `CollaborationRequestsFacade.java` — two signatures gain `boolean testSubmission`, two `@Deprecated(forRemoval=true)` default overloads added delegating `true`, plus javadoc. Traces to the core design (per-call param) and the compat-preservation bullet. ✓
- `DefaultTsaNetApiSession.java` — both public signatures + private `createFromForm` threaded, and `setTestSubmission(true)` → `setTestSubmission(testSubmission)`. This is the actual behavior change. ✓
- New test file — three wire-form assertions. Traces to the Receipts section. ✓

All hunks trace. No incidental tidying.

**Correctness:**

Traced `createRequest(template, "CASE-1", "s", "d", Map.of(), false)`:
- `createFromForm` receives `testSubmission=false` → `form.setTestSubmission(false)` (line 362 in diff) → form captured by the gateway mock → `itSendsFalseWhenAProductionConsumerPassesFalse` asserts `.getTestSubmission()).isFalse()`. Correct.

The 4-arg `receiverCompanyId` overload path: `createRequest(companyId, ...)` now threads `testSubmission` into the 6-arg delegate (`DefaultTsaNetApiSession` line ~314). The deprecated default in the facade delegates `true`, preserving old behavior. The PR correctly names this compat path as untested (routes through `getCreateFormByCompanyId`, more mocking needed) — the 5-arg overload test covers the delegation pattern. Acceptable gap, honestly disclosed.

Red probe cited (re-hardcode `true` fails exactly `itSendsFalseWhenAProductionConsumerPassesFalse`) — this is the assertion that would catch a regression, and it's the one that fails on the mutation. Good.

One note, not a defect: `getTestSubmission()` returning `Boolean` (boxed) vs primitive — `isFalse()`/`isTrue()` on an AssertJ `BooleanAssert` will NPE if the getter returns null, but the code sets it unconditionally on every path, so it's never null here. Fine.

**Security:** None. No secrets, no external input reaching a query/shell/path. The new param is a `boolean` — no injection surface.

**Regression risk:**

- **Consumers of the two facade signatures:** in-repo, `CreateRequestExecutor` and the Acme scenario (both updated). External: `shawn-tsanet/connect-sdk-demo` — the PR names this and the deprecated default overloads keep it compiling. I have not independently verified that repo's call sites, but the compat overloads make source compatibility hold by construction (old 4-/5-arg calls still resolve to the defaults).
- **Implementor sweep:** PR states `DefaultTsaNetApiSession` is the only implementor and reactor compile is the proof the new abstract methods are implemented. Consistent with the diff (only one implementor touched). New abstract methods on an interface would break any *other* implementor — the claim that there's only one is the load-bearing fact here, and it's asserted via compile, which is the right proof.
- **Behavior change:** the meaningful one is `setTestSubmission(true)` → `setTestSubmission(testSubmission)`. Every existing in-repo caller now passes `true`, and both deprecated overloads pass `true`, so no path silently flips to production. The only way `false` reaches the wire is an explicit new-signature call. Correct and safe.

**Recommendation:** Merge — the behavior change is minimal, every caller is either updated or covered by a compat default, and the false-path is asserted on the wire form; the untested 4-arg compat path is disclosed and low-risk.

---

shawn-tsanet:
Good to merge

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `CreateRequestExecutor.java` (+1): CLI caller passes `true` explicitly. Traces to "both in-repo callers now pass `true`."
- `AcmeCreateCollaborationRequestsScenario.java` (+1): demo caller passes `true`. Same trace.
- `CollaborationRequestsFacade.java` (~40 lines): both `createRequest` signatures gain `boolean testSubmission`, plus two `@Deprecated(forRemoval=true)` default overloads delegating `true`. Traces directly to the design section.
- `DefaultTsaNetApiSession.java` (5 hunks): parameter threaded through both public signatures → `createFromForm` → `form.setTestSubmission(testSubmission)`. Traces to "replacing the hardcoded `setTestSubmission(true)`."
- New test file (+121): three assertions on the wire form. Traces to the receipts.

**Correctness:**

Traced `createRequest(receiverCompanyId,...)` 4-arg deprecated overload → default method delegates `true` → 5-arg abstract → `DefaultTsaNetApiSession.createRequest(long,...)` now takes `testSubmission`, calls `getCreateFormByCompanyId` then delegates to the template overload with `testSubmission` → `createFromForm` → `form.setTestSubmission(testSubmission)`. Correct; the flag reaches the wire form unchanged.

The `false` path (`itSendsFalseWhenAProductionConsumerPassesFalse`) is the real behavioral change and it's asserted on the captured gateway argument — a genuine wire-form check, not a proxy. Test setup mocks `formGateway` and `collaborationGateway` and captures the actual DTO. Good.

Edge/blast-radius notes:
- **Guard/representation:** N/A — this is a plumbed parameter, not a guard.
- **Positional contract:** the new parameter is appended last on both signatures, so no existing positional arg was renumbered. The caller sweep in the PR body claims "both real call sites updated; remaining grep hits are javadoc" and the reactor compile is cited as mechanical proof the two abstract methods are implemented (single implementor `DefaultTsaNetApiSession`). Since the old 4-arg/5-arg call shapes still resolve to the deprecated defaults, any missed caller would compile silently and keep the old `true` behavior — that's the safe direction, and the sweep claim covers it.
- **Untested compat path:** the 4-arg deprecated overload (by companyId) has no dedicated test — the PR names this explicitly. The 5-arg deprecated overload (by template) IS covered by `theDeprecatedOverloadKeepsTheOldAlwaysTestBehavior`, which asserts `true`. The delegation of the 4-arg path is identical in shape, so this is an acceptable named gap, not a blocking one.

**Security:** None. No secrets, no injection surface, no auth/credential handling. Test fixtures use throwaway literals.

**Regression risk:**

Low. The change is additive at the signature level: old call shapes bind to `@Deprecated` defaults that preserve the exact prior `true` behavior. In-repo callers were both updated to pass `true` — zero behavior change for CLI and demo. External consumer `connect-sdk-demo` keeps compiling via the deprecated defaults; its migration is the named follow-up. The single interface implementor is `DefaultTsaNetApiSession`, and both new abstract methods are implemented there. Only new runtime behavior is that a consumer can now pass `false` and reach a real partner — that's the intended feature, gated behind an explicit, no-default choice.

One thing worth tracking (not blocking): the deprecated defaults silently preserve `true`, so a consumer who forgets to migrate never gets a compile error — they just keep sending test submissions. That's the safe failure direction and matches the issue's intent, but the removal at the next release boundary is what forces the migration. That schedule is asserted only in the `@deprecated` javadoc text, not by any check — consistent with #40's approach per the PR.

**Recommendation:** Merge — the change is minimal, correctly threaded, and the behavioral `false` path is asserted on the wire form; the one untested compat overload is explicitly named and fails safe.

---

PR #42, opened 2026-08-10
README: document the sibling-spec build requirement and branch state

shawn-tsanet:
The README was one line, and the build's hardest fact was recorded nowhere: codegen reads the Connect OpenAPI spec from a sibling clone of `Connect-API-Code`, so a lone clone of this repository does not build, and the sibling's checked-out branch decides which symbols get generated. Alignment-agenda conditions 4 and 6.

Every documented step was run before being written, today:

- Sibling on `beta` + this repo at `0f57facd3326`: `mvn install` green, and that stack passes the live BETA contract suite (the Connect_Gateway `v0.1.0` certification pin).
- Same sibling + `main` at `332e5c7`: compile FAILS with cannot-find-symbol in `ConnectApiWebhooksGateway` (three sites) - documented as a known release-path condition so nobody debugs their environment over it.
- The `connect.openapi.spec` path quoted in the README was checked against the root pom on both `main` and `oauth` (`pom.xml:31`, identical) - the pre-PR gate flagged this as its one confirm-before-merge item; confirmed.

Deliberately NOT in this PR: vendoring the spec (surfaced as a maintainer option). One consequence worth deciding at the alignment session: `Connect-API-Code` is private, so an external engineer building this library from source needs either access to it, a vendored spec, or a published artifact - the sibling requirement is an access question, not just a build step.

Siblings: #39 (jsr310), #40 (testCase), #41 (per-call test flag).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #43, opened 2026-08-10
Publish connect-library to GitHub Packages on release

shawn-tsanet:
The machinery for alignment-agenda condition 2's second half: on a published GitHub Release (or manual dispatch), build, test and publish the root pom plus `connect-library` to GitHub Packages. Mirrors `Connect_Gateway`'s connector-core publish (release-cuts-the-version, SNAPSHOT refusal, x.y.z validation, `versions:set`, deploy runs the full lifecycle so the 84 tests gate the release). Base is `oauth`, same as #39-#42. Harmless to merge before any tag exists - the workflow sits dormant until you publish a release.

## What this repo forces, and the one deliberate deviation

- **The spec sibling**: codegen reads the Connect OpenAPI spec from a sibling `Connect-API-Code` checkout and that repo is private, so the workflow checks it out (`beta`) with a `CONNECT_API_CODE_TOKEN` secret (a PAT with read on Connect-API-Code; the secret does not exist yet - configuring it is part of activation).
- **Deviation, recorded**: the task brief said skip-politely when the secret is missing. This workflow FAILS LOUDLY instead, with the secret named in the error. Reason: Fin-Intercom_App's CI carries a comment documenting exactly that token-gated polite skip "skipping every real step while reporting success" - and for a publish, a green run that shipped no artifact is a release that silently does not exist. Loud and actionable beats green and empty.
- **Scoping**: reactor selection (`-pl .,connect-library`) publishes the parent pom (consumers must resolve it) and the library only; the app and demo modules never enter the reactor and their poms are untouched. Receipt: `connect-library` has no dependency on any sibling module (pom grep: only the parent reference), and the local dry-run deploy yields exactly `tsanet-client-parent.pom` + `connect-library.{jar,pom}` and nothing else.

## Receipts

- Dry-run deploy to a local file repository (`-DaltDeploymentRepository`): correct artifact set, nothing extra. Workflow YAML parse-validated. No workflow-name collision (this repo has no workflows directory at the base).
- Credential wiring: `server-id: github` matches `distributionManagement/<id>github</id>`; `packages: write` + the run's `GITHUB_TOKEN` suffices because the registry URL is this repo. A fork run fails loudly (fork tokens cannot write upstream packages) - acceptable.

## Behavior worth agreeing to explicitly

- **Prereleases cannot publish**: a prerelease fires `release: published`, and a tag like `v1.2.3-rc.1` fails the x.y.z gate with a loud error. Intended - say the word if you want an rc scheme supported.
- **`ref: beta` on the spec checkout** means a version's published content depends on when the release is cut; rebuilding the same version later could differ if beta moved. Connect_Gateway pins its spec the same way today, so consistency wins - pinning to a spec SHA per release is a reasonable follow-up once the release cadence exists.
- **Consumers need a one-time token**: GitHub Packages requires a `read:packages` PAT for Maven downloads even on public packages. Consumer `settings.xml`:

```xml
<servers>
  <server>
    <id>github</id>
    <username>YOUR_GITHUB_USERNAME</username>
    <password>YOUR_PAT_WITH_READ_PACKAGES</password>
  </server>
</servers>
```

and in the consumer pom: `<repository><id>github</id><url>https://maven.pkg.github.com/tsanetgit/Connect_SDK</url></repository>`.

**Sequencing note**: publishing before #39 (jsr310) merges would ship the embedability defect to every non-Boot consumer - land #39 first.

Siblings: #39 jsr310, #40 testCase, #41 per-call test flag, #42 build README.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,103 @@
+name: publish
+
+# A published GitHub Release is what cuts a version, deliberately rather than on
+# every push: consumers pin a version, so a library change reaches them only when
+# someone decides it should. Same pattern as Connect_Gateway's connector-core
+# publish. Publishes the root pom (connect-library's parent, which consumers must
+# be able to resolve) and connect-library ONLY; the app and demo modules are
+# never published.
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to publish, no leading v (e.g. 0.1.0)"
+        required: true
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      # connect-library's codegen reads the Connect OpenAPI spec from a sibling
+      # checkout of Connect-API-Code (the connect.openapi.spec property), which is
+      # a PRIVATE repository, so this checkout needs a token with read access to
+      # it. If the secret is missing, FAIL LOUDLY rather than skip: a green
+      # publish run that shipped no artifact is the skip-reads-as-success trap
+      # Fin-Intercom_App's CI documented after living it. The two checkout paths
+      # below place the repos as siblings, which is the layout the pom expects.
+      - name: Require the spec-repo token
+        env:
+          HAVE_TOKEN: ${{ secrets.CONNECT_API_CODE_TOKEN != '' }}
+        run: |
+          set -euo pipefail
+          if [ "$HAVE_TOKEN" != "true" ]; then
+            echo "::error::The CONNECT_API_CODE_TOKEN secret is not configured. Publishing builds connect-library from the Connect-API-Code spec (a private repo), so a read token for it is required. Configure the secret and re-run."
+            exit 1
+          fi
+
+      - uses: actions/checkout@v4
+        with:
+          path: Connect_SDK
+          persist-credentials: false
+
+      - uses: actions/checkout@v4
+        with:
+          repository: tsanetgit/Connect-API-Code
+          ref: beta
+          token: ${{ secrets.CONNECT_API_CODE_TOKEN }}
+          path: Connect-API-Code
+          persist-credentials: false
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "21"
+          cache: maven
+          # Writes the ~/.m2/settings.xml server entry that the root pom's
+          # distributionManagement <id>github</id> resolves against.
+          server-id: github
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_TOKEN
+
+      - name: Resolve the version to publish
+        id: v
+        env:
+          EVENT: ${{ github.event_name }}
+          TAG: ${{ github.ref_name }}
+          INPUT: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT" = "release" ]; then V="${TAG#v}"; else V="$INPUT"; fi
+
+          # Refuse a SNAPSHOT: a consumer pinned to a SNAPSHOT is pinned to
+          # nothing, which reintroduces the silent drift publishing removes.
+          case "$V" in
+            *-SNAPSHOT) echo "::error::Refusing to publish a SNAPSHOT ($V)."; exit 1 ;;
+          esac
+          if ! printf '%s' "$V" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Version '$V' is not x.y.z."; exit 1
+          fi
+          echo "version=$V" >> "$GITHUB_OUTPUT"
+          echo "publishing $V"
+
+      - name: Set the project version
+        working-directory: Connect_SDK
+        run: |
+          mvn -B --no-transfer-progress versions:set \
+            -DnewVersion="${{ steps.v.outputs.version }}" -DgenerateBackupPoms=false
+
+      # `deploy` runs the full lifecycle, so connect-library's tests gate the
+      # release: a broken client cannot be published. Reactor selection publishes
+      # the parent pom and the library only.
+      - name: Build, test and publish connect-library
+        working-directory: Connect_SDK
+        env:
+          MAVEN_USERNAME: ${{ github.actor }}
+          MAVEN_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: mvn -B --no-transfer-progress -pl .,connect-library deploy
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -100,4 +100,6 @@ jobs:
         env:
           MAVEN_USERNAME: ${{ github.actor }}
           MAVEN_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-        run: mvn -B --no-transfer-progress -pl .,connect-library deploy
+        # -Dmaven.deploy.skip=false arms the deploy that the root pom disarms by
+        # default, so this workflow is the only path that publishes.
+        run: mvn -B --no-transfer-progress -pl .,connect-library deploy -Dmaven.deploy.skip=false
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -12,7 +12,7 @@ on:
   workflow_dispatch:
     inputs:
       version:
-        description: "Version to publish, no leading v (e.g. 0.1.0)"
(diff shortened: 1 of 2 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**

- `publish.yml` `on:` / `permissions:` / job scaffold (lines 6–24) — traces to "on a published GitHub Release (or manual dispatch), build, test and publish". ✓
- "Require the spec-repo token" step (lines 33–42) — traces to the recorded deviation ("FAILS LOUDLY instead"). ✓
- Two `checkout` steps (lines 44–57) — traces to "checks out Connect-API-Code sibling". ✓
- `setup-java` with `server-id: github` (lines 59–71) — traces to credential wiring. ✓
- "Resolve the version" step (lines 73–92) — traces to SNAPSHOT refusal + x.y.z gate. ✓
- `versions:set` + `deploy` steps (lines 94–103) — traces to "versions:set, deploy runs the full lifecycle". ✓
- `pom.xml` `distributionManagement` block (lines +34–47) — traces to "publish … to GitHub Packages". ✓

All hunks trace. No scope creep.

**Correctness:**

- Version resolution trace — release event with tag `v1.2.3`: `V="${TAG#v}"` → `1.2.3`, passes SNAPSHOT case and the `^[0-9]+\.[0-9]+\.[0-9]+$` grep, sets output. ✓ Prerelease `v1.2.3-rc.1` fails the grep loudly — matches the documented intent. ✓ `workflow_dispatch` with input `0.1.0` → passes. ✓
- **Edge worth naming (probe not shown):** the x.y.z gate rejects a manual-dispatch input that already carries a leading `v` (e.g. `v0.1.0`) — for `workflow_dispatch` the `${TAG#v}` strip does not run, so `v0.1.0` fails the regex. The input description says "no leading v", so this is a well-formed-but-rejected input handled by a loud error, which is acceptable. Non-blocking, but the PR's "names the wrong-but-well-formed input it rejects" edge is only partially swept — the rc case is probed, the leading-v dispatch case is not.
- `checkout` step for `Connect_SDK` uses `path: Connect_SDK` but the token-require step (lines 33–42) runs with no `working-directory` and no checkout dependency — it runs first and only reads an env var, so ordering is fine. ✓
- `versions:set` runs on the parent only (`working-directory: Connect_SDK`, no `-pl`), which reactor-recurses to child modules by default — so `connect-library`'s version is set even though app/demo also get set. Since app/demo are never deployed, harmless. The parent+library deploy will carry the new version consistently. ✓
- One real concern: **the `deploy` step re-derives nothing from `steps.v` — it deploys whatever `versions:set` wrote.** That is correct, but there's no guard that `versions:set` actually succeeded before `deploy`; a `versions:set` failure would fail the step and stop the job (default `fail-fast`), so this is fine. ✓

**Security:**

- No secrets in the diff. `CONNECT_API_CODE_TOKEN` and `GITHUB_TOKEN` are referenced via `secrets.`/env, never printed. ✓
- `HAVE_TOKEN: ${{ secrets.CONNECT_API_CODE_TOKEN != '' }}` evaluates to a boolean string, not the token value — no leak. ✓
- Both checkouts set `persist-credentials: false`. ✓
- `set -euo pipefail` in every `run` block. ✓
- No injection: `TAG`/`INPUT` flow through env vars into shell, and are validated before use in `versions:set`. The `versions:set` step interpolates `${{ steps.v.outputs.version }}` directly into the `mvn` line — but that value already passed the `^[0-9]+\.[0-9]+\.[0-9]+$` gate, so it cannot contain shell metacharacters. ✓

**Regression risk:**

- New workflow file, no existing workflow to collide with (PR asserts "this repo has no workflows directory at the base" — I have not independently verified the base tree this session, but a `new file mode` diff on `.github/workflows/publish.yml` is consistent with that).
- `pom.xml`: adding `distributionManagement` changes `mvn deploy` behavior repo-wide — any human running a bare `mvn deploy` locally now targets GitHub Packages. Previously `deploy` had no target and would fail. This is a behavior change to a shared build file, but `deploy` is not a routine local command and the workflow scopes it with `-pl`. Low risk, worth Shawn knowing.
- **Sequencing dependency stated in the PR body (land #39 first) is a real cross-PR ordering constraint** — not a code regression here, but merging/activating this before #39 ships the jsr310 defect to consumers. This PR is a draft on base `oauth`, so dormant-until-tagged holds.

**Operational concerns:**

- New secret `CONNECT_API_CODE_TOKEN` required for the workflow to do anything — PR states it does not exist yet and configuring it is part of activation. No `.env`/`.env.example` impact (this is a GitHub Actions secret, not sidecar env). ✓
- No sidecar/LaunchAgent restart, no schema change, no new Python dependency. Nothing touches the ops-sidecar. ✓
- `ref: beta` non-reproducibility is acknowledged as a follow-up — agreed, track it but don't hold merge.

**Recommendation:** Merge — findings (leading-`v` dispatch input rejection unprobed, shared-pom `deploy` behavior change, #39 sequencing) are follow-ups, none is a broken contract or security exposure on this dormant draft.

---

shawn-tsanet:
Disposition on the one CONCERNS item (leading-`v` manual-dispatch input): intended behavior, kept. The input description says "no leading v", and the rejection is loud with the offending value named, so the operator fix is immediate. Making the input forgiving is a one-line change (`V="${INPUT#v}"`); happy to fold it in at merge if you would rather the dispatch path match the release path

shawn-tsanet:
Per-finding response to the auto-QA round (verdict CONCERNS, recommendation Merge):

1. **Leading-`v` dispatch input**: dispositioned above - intended, loud, one-line fix available at merge if preferred.
2. **Bare local `mvn deploy` behavior change**: FIXED rather than dispositioned, in the follow-up commit. The root pom now disarms deploy by default (`maven.deploy.skip=true`) and the workflow is the only path that arms it (`-Dmaven.deploy.skip=false`) - the same toggle pattern Connect_Gateway uses. Probed both ways: bare deploy to a local file repository publishes nothing; the armed form publishes exactly `tsanet-client-parent.pom` + `connect-library.{jar,pom}`. This also closes the accidental-SNAPSHOT-publish hole the review implied, since the SNAPSHOT gate only exists on the workflow path.
3. **#39-before-publish sequencing**: already the PR body's own bolded warning; it stands.

shawn-tsanet:
Finding 1 is now also FIXED rather than dispositioned, per Shawn: the dispatch path strips a leading `v` exactly as the release path does, and the input description matches. Probe across all four input shapes: `0.1.0` and `v0.1.0` both accept (resolving to `0.1.0`), `v1.2.3-rc.1` rejects as not x.y.z, `0.1.0-SNAPSHOT` rejects as SNAPSHOT. All three findings from the auto-QA round are now closed.

shawn-tsanet:
Good to Merge

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass.

**Correctness:**
- Leading-`v` dispatch fix (line 76): `V="${INPUT#v}"`. Trace `v0.1.0` → `0.1.0`, passes SNAPSHOT case and `^[0-9]+\.[0-9]+\.[0-9]+$` → accepts. `0.1.0` → unchanged → accepts. `v1.2.3-rc.1` → `1.2.3-rc.1` → fails regex loudly. `0.1.0-SNAPSHOT` → fails SNAPSHOT gate. All four shapes match the claimed probe. ✓
- `maven.deploy.skip=true` in pom (line +34) + `-Dmaven.deploy.skip=false` on the workflow deploy (line 105): CLI `-D` overrides the pom property, so the workflow arms deploy while a bare local `mvn deploy` is a no-op. Standard Maven property-precedence — correct. ✓ Note: `versions:set` (unchanged step) runs before deploy and is unaffected by `deploy.skip`; only the deploy phase is gated, which is the intent.

**Security:**
None.

**Regression risk:**
- `maven.deploy.skip=true` is a repo-wide pom property. Any other consumer of `mvn deploy` on this pom is now a no-op unless it also passes `-Dmaven.deploy.skip=false`. This workflow is the only deploy path (PR asserts, consistent with #39–#42 all targeting `oauth`), so no live consumer regresses. This is the intended hardening, not a defect.

**Carried forward:**
- Leading-`v` dispatch input rejection — RESOLVED (`${INPUT#v}` strip, four-shape probe cited)
- Bare local `mvn deploy` behavior change — RESOLVED (`deploy.skip=true` default + workflow-only arm)
- #39-before-publish sequencing — FOLLOW-UP (cross-PR ordering, PR body warning stands)

**Recommendation:** Merge — both prior CONCERNS items are fixed with cited probes and the new commits introduce nothing that holds the merge.

---

PR #44, opened 2026-08-11
Merge the oauth branch: OAuth client-credentials auth plus the facade and packaging work

shawn-tsanet:
Brings the \`oauth\` integration branch into main. Contents:

- **OAuth basic implementation** (Slava, 2026-07-30): client-credentials auth mode alongside password auth. \`AuthMode\`, \`ClientCredentialsAuthConfig\`, \`OAuthTokenGateway\`, \`TokenManager\`, \`OAuth401RetryInterceptor\`, and the session/registry plumbing to select auth per account.
  - The same commit (\`0f57fac\`) also migrates the webhook surface to the newer spec: \`WebhookDeliveryDto.integrationId\` (Long) becomes \`cloudEventId\` (String) and the gateway calls \`WebhooksV1Api\` for list/create. This is the change that lets the branch compile against the sibling \`beta\` spec; main at \`332e5c7\` does not (see the README branch notes from \`tsanetgit/Connect_SDK#42\`).
- Five PRs squash-merged onto the branch since (2026-08-10):
  - \`tsanetgit/Connect_SDK#39\`: declare jackson-datatype-jsr310 so the client is self-sufficient outside Boot
  - \`tsanetgit/Connect_SDK#40\`: surface testCase through the facade DTO and cache (\`tsanetgit/Connect_SDK#38\`)
  - \`tsanetgit/Connect_SDK#41\`: make the test flag a per-call parameter on the facade create API (\`tsanetgit/Connect_SDK#37\`)
  - \`tsanetgit/Connect_SDK#42\`: README: document the sibling-spec build requirement and branch state
  - \`tsanetgit/Connect_SDK#43\`: publish connect-library to GitHub Packages on release
- Incidental tidying carried by the branch: \`tsaet\` to \`tsanet\` typo fixes in three log/error strings, and two stale comments removed in \`ConnectApiWebhooksGateway\`.

No overlap with main's \`tsanetgit/Connect_SDK#35\` (credential removal + security-audit CI): the branch touches neither \`application.yml\` nor the audit tooling, so nothing removed there comes back.

Build receipt for the head (\`f747bc6\`), sibling \`Connect-API-Code\` on \`beta\` (\`0691e1f\`), JDK 21: \`mvn install\` BUILD SUCCESS, 122 tests (connect-library 86, demo 6, app 30), 0 failures.

Merging as a merge commit to keep the individual commit trail.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -1,0 +1,105 @@
+name: publish
+
+# A published GitHub Release is what cuts a version, deliberately rather than on
+# every push: consumers pin a version, so a library change reaches them only when
+# someone decides it should. Same pattern as Connect_Gateway's connector-core
+# publish. Publishes the root pom (connect-library's parent, which consumers must
+# be able to resolve) and connect-library ONLY; the app and demo modules are
+# never published.
+on:
+  release:
+    types: [published]
+  workflow_dispatch:
+    inputs:
+      version:
+        description: "Version to publish (e.g. 0.1.0; a leading v is tolerated)"
+        required: true
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  publish:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      # connect-library's codegen reads the Connect OpenAPI spec from a sibling
+      # checkout of Connect-API-Code (the connect.openapi.spec property), which is
+      # a PRIVATE repository, so this checkout needs a token with read access to
+      # it. If the secret is missing, FAIL LOUDLY rather than skip: a green
+      # publish run that shipped no artifact is the skip-reads-as-success trap
+      # Fin-Intercom_App's CI documented after living it. The two checkout paths
+      # below place the repos as siblings, which is the layout the pom expects.
+      - name: Require the spec-repo token
+        env:
+          HAVE_TOKEN: ${{ secrets.CONNECT_API_CODE_TOKEN != '' }}
+        run: |
+          set -euo pipefail
+          if [ "$HAVE_TOKEN" != "true" ]; then
+            echo "::error::The CONNECT_API_CODE_TOKEN secret is not configured. Publishing builds connect-library from the Connect-API-Code spec (a private repo), so a read token for it is required. Configure the secret and re-run."
+            exit 1
+          fi
+
+      - uses: actions/checkout@v4
+        with:
+          path: Connect_SDK
+          persist-credentials: false
+
+      - uses: actions/checkout@v4
+        with:
+          repository: tsanetgit/Connect-API-Code
+          ref: beta
+          token: ${{ secrets.CONNECT_API_CODE_TOKEN }}
+          path: Connect-API-Code
+          persist-credentials: false
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "21"
+          cache: maven
+          # Writes the ~/.m2/settings.xml server entry that the root pom's
+          # distributionManagement <id>github</id> resolves against.
+          server-id: github
+          server-username: MAVEN_USERNAME
+          server-password: MAVEN_TOKEN
+
+      - name: Resolve the version to publish
+        id: v
+        env:
+          EVENT: ${{ github.event_name }}
+          TAG: ${{ github.ref_name }}
+          INPUT: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT" = "release" ]; then V="${TAG#v}"; else V="${INPUT#v}"; fi
+
+          # Refuse a SNAPSHOT: a consumer pinned to a SNAPSHOT is pinned to
+          # nothing, which reintroduces the silent drift publishing removes.
+          case "$V" in
+            *-SNAPSHOT) echo "::error::Refusing to publish a SNAPSHOT ($V)."; exit 1 ;;
+          esac
+          if ! printf '%s' "$V" | grep -qE '^[0-9]+\.[0-9]+\.[0-9]+$'; then
+            echo "::error::Version '$V' is not x.y.z."; exit 1
+          fi
+          echo "version=$V" >> "$GITHUB_OUTPUT"
+          echo "publishing $V"
+
+      - name: Set the project version
+        working-directory: Connect_SDK
+        run: |
+          mvn -B --no-transfer-progress versions:set \
+            -DnewVersion="${{ steps.v.outputs.version }}" -DgenerateBackupPoms=false
+
+      # `deploy` runs the full lifecycle, so connect-library's tests gate the
+      # release: a broken client cannot be published. Reactor selection publishes
+      # the parent pom and the library only.
+      - name: Build, test and publish connect-library
+        working-directory: Connect_SDK
+        env:
+          MAVEN_USERNAME: ${{ github.actor }}
+          MAVEN_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        # -Dmaven.deploy.skip=false arms the deploy that the root pom disarms by
+        # default, so this workflow is the only path that publishes.
+        run: mvn -B --no-transfer-progress -pl .,connect-library deploy -Dmaven.deploy.skip=false
--- a/README.md
+++ b/README.md
@@ -1,1 +1,42 @@
-Demonstration of Integration with TSANet API - In Java
+# Connect_SDK
+
+Demonstration of Integration with TSANet API - In Java.
+
+Modules: `connect-library` (the Java client for the TSANet Connect API),
+`TSANet-integration-app` (console app), `TSANet-integration-demo` (demo scenarios).
+
+## Building
+
+The client code in `connect-library` is generated from the Connect API's OpenAPI
+specification at build time, and the build reads that specification from a **sibling
+clone** of `tsanetgit/Connect-API-Code` (the `connect.openapi.spec` property in the
+root pom resolves to `../Connect-API-Code/connect-core-api/src/main/resources/openapi.yaml`).
+A clone of this repository alone does not build.
+
+The working steps, in order:
+
+1. Clone `tsanetgit/Connect-API-Code` beside this repository (the directory must be
+   named `Connect-API-Code`) and check out the `beta` branch. The generated symbols
+   depend on which specification the sibling provides, so the branch matters.
+2. In this repository, check out the commit you are building (see the branch notes
+   below).
+3. `mvn install` from the root, or `mvn -pl connect-library -am install` for the
+   library alone. JDK 21.
+
+## Branch notes: what builds today
+
+- **`oauth` at `0f57facd3326` builds against the beta specification and passes the
+  live BETA contract suite.** This is the commit the Connect Gateway's `v0.1.0`
+  certifies against, and the pin to use until `oauth` merges and a release is
+  tagged.
+- **`main` at `332e5c7` does not compile against the beta specification**:
+  `ConnectApiWebhooksGateway` references generated symbols that specification does
+  not produce (cannot-find-symbol at its webhook-gateway call sites), because that
+  code tracks a newer specification than the sibling's `beta` branch carries. This
+  is a known condition on the release path, not a local setup problem; do not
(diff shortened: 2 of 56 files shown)


shawn-tsanet (commented):
This is an integration merge of a feature branch with 6 logical changes. Let me note upfront: I'm reviewing the diff text only — I have not run the build, cannot verify generated symbols (`WebhooksV1Api`, `dto.getTestCase()`, `dto.getCloudEventId()`) exist against the `beta` spec, and cannot verify tests pass. Those are receipts I don't have this session.

**Verdict:** CONCERNS

**Minimal-Diff Rule check:**

Walking the hunks against the six stated components:

- `.github/workflows/publish.yml` — traces to PR #43 (publish to GitHub Packages). ✓
- `pom.xml` (distributionManagement, `maven.deploy.skip`) — PR #43. ✓
- Root `README.md`, `connect-library/README.md` — PR #42 (build docs) + #40/#39 doc surfacing. ✓
- `jackson-datatype-jsr310` dep in `connect-library/pom.xml` — PR #39. ✓
- `testCase` on `CollaborationRequestStatusDto`, gateways, repository, `DatabaseInitializer`, cache tests — PR #40 (surface testCase). ✓
- `testSubmission` per-call param on facade `createRequest`, CLI/demo call sites, session test — PR #41. ✓
- `AuthMode`/`ClientCredentialsAuthConfig`/`OAuthTokenGateway`/`TokenManager`/`OAuth401RetryInterceptor`, session-store, config plumbing, `AccountScopedTsaNetApiSession` — the OAuth base work. ✓

**Untraced hunks:**
- `WebhookDeliveryDto`: `Long integrationId` → `String cloudEventId` (WebhookDeliveryDto.java, and the mapper in `ConnectApiWebhooksGateway.toDeliveryDto`), plus the swap of `webhooksApi` → `webhooksV1Api` for list/create. This is a real behavior/contract change and does **not** trace to any of the six bullets in the PR body. It's the kind of "tracks a newer spec" change the README hunk alludes to, but the PR body never mentions a webhook API migration. **Flag: untraced scope.** Not blocking on its own, but it should be named in the PR body since it changes a DTO field type and switches which generated API is called.
- The `tsaet` → `tsanet` typo fixes (ConnectApiPollingCoordinator log, two `requireBy*` error strings) — incidental tidying, not in the request. Minor, but by the rule these are untraced. Nit-level.
- Removed `// Refresh stored list after mutation.` comments in `ConnectApiWebhooksGateway` — incidental. Nit-level.

**Correctness:**

- OAuth refresh path traced: client-credentials account → `bearerTokenForRequest` (TsaNetApiRuntime.java:196) → `tokenManager.ensureValidAccessToken()` → if `sessionStore.isExpired` refresh, else return stored token. `isExpired` uses `EXPIRY_SKEW` 60s. `TokenManagerTest.itRefreshesExpiredOAuthTokenBeforeUse` covers the expired case with a fixed clock. Looks correct.
- **`bearerTokenForRequest` gate on first call:** at line ~196 it returns `null` when `getBearerToken().isEmpty()`. For a client-credentials session that has never authenticated, the very first API call gets a null bearer even though `ensureValidAccessToken` could mint one. That's intentional-ish (caller must call `authenticate()`), and the `OAuth401RetryInterceptor` will catch the resulting 401 and refresh+retry. So a first call without explicit `authenticate()` costs one wasted 401 round-trip but still succeeds. Worth confirming that's the intended flow, not a bug — it's benign either way.
- **401 retry infinite-loop risk:** `OAuth401RetryInterceptor.intercept` refreshes and re-executes once. The re-execution is not itself wrapped, so a persistent 401 returns after one retry — no loop. Good. But if the token endpoint itself is what's failing, `refreshAccessToken` throws `IllegalStateException` out of the interceptor; the receiving handler is whatever called the API. `ensureAuthenticatedForWebhook` and the polling coordinator catch broadly, but ad-hoc facade callers get a raw `IllegalStateException`. Acceptable, but that's the user-visible outcome for a bad secret: an exception, not a clean auth error.
- **`OAuthAccessToken` compact constructor mutates `expiresInSeconds`** (sets 3600 when ≤0) but the record component reassignment in a compact constructor is valid Java. Fine.
- `readStoredBoolean`/`toStoredBoolean` (CollaborationRequestRepository) — the NULL-via-`wasNull` read is the correct JDBC idiom and the comment explains why `getObject(Integer.class)` was avoided. The upgrade test (`itUpgradesAPreExistingDatabaseAndBackfillsTestCaseOnRefresh`) uses a frozen pre-column DDL and asserts the `ensureColumn` migration adds the column and the upsert backfills. This is exactly the "CREATE TABLE IF NOT EXISTS is a no-op on existing data" edge, and it's asserted, not commented. Well done.
- **`testSubmission` default flip — this is the one to eyeball.** The deprecated no-flag overloads (`CollaborationRequestsFacade` and the DTO compat constructor) default to `true` (always TEST). The CLI (`CreateRequestExecutor.java:40`) and demo (`AcmeCreateCollaborationRequestsScenario.java:127`) were updated to pass `true` explicitly. So current callers preserve TEST behavior. Correct and conservative. But note: the *point* of PR #41 was to make production submissions possible by passing `false` — no in-repo caller passes `false` yet except tests. That's fine, it's a library capability, but confirm no consumer expected the default to become `false`.
- `authMode()`/`currentAccountId()` on `DefaultTsaNetApiSession` fall back to `configuration` values via `.or(...)` even before authentication — so `authMode()` reports the *configured* mode pre-login. Consistent with the `AccountScoped` variant returning empty when `delegate == null`. Slight asymmetry between the two session impls (one falls back to config, one returns empty), but not wrong.

**Security:**

- No secrets in the diff. `client-secret` in README examples uses `${TSANET_CLIENT_SECRET}` / `System.getenv` — good.
- `OAuthTokenGateway.fetchClientCredentialsToken` posts `client_secret` as a form field over the resolved token URL. On a `RestClientResponseException` it logs `HTTP <status>` only — does **not** log the response body or the secret. Good; that's the right level of restraint given error bodies from token endpoints can echo request context.
- `resolvedTokenUrl` interpolates `tenantId` into the Azure URL. `tenantId` comes from config, not external request input, and there's no shell/SQL sink — no injection concern.
- **Bearer token now flows through `RestTemplate` interceptor and `setBearerToken` supplier.** Confirm no logging interceptor prints headers. `BufferingClientHttpRequestFactory` was added (for the 401 retry to re-read the body) — this buffers request/response in memory; not a secret exposure, just a memory note for large payloads.
- SQL in `CollaborationRequestRepository` is fully parameterized (`?` placeholders). No injection.

**Regression risk:**

- **Constructor signature change on `ApplicationUserAccount`** (`(id, username, password, sqlitePath)` → `(id, sqlitePath, auth)`) is a positional-contract change. The PR body does not include a grep of consumers. I can see updated call sites: `ApplicationUserAccountRegistry.fromLegacyAuth`, both `*ApplicationProperties.toAccount`, `TsaNetApiSessionFactory`, and three test files. Whether that's *all* of them, I can't verify from the diff — **the enumeration artifact is missing.** Given `ApplicationUserAccount` is a public API type in `com.tsanet.api`, any external consumer constructing it directly breaks. This is the highest-value thing to confirm before merge: run the build (it won't compile if a call site was missed) and, if there are downstream repos, grep them.
- **`ConnectApiWebhooksGateway` constructor gained a parameter** (`webhooksV1Api` first). Sole caller is `TsaNetApiRuntime.create`, updated. ✓ within this repo.
- **`WebhookDeliveryDto.integrationId` → `cloudEventId`** — any consumer reading `.integrationId()` off this DTO breaks. No enumeration in PR body. This DTO is public (`com.tsanet.api.connectapi.dto`). The test asserts the new field maps from `dto.getCloudEventId()`, so the generated model supports it — but I can't confirm the generated `WebhookDeliveryLogDTO` actually has `getCloudEventId()` against the beta spec (the README explicitly warns main doesn't compile against beta because of webhook-gateway symbols). **This is the compile-risk to verify.** If `getCloudEventId()`/`WebhooksV1Api` aren't in the beta-generated code, the branch doesn't build — and the README's own note about main's webhook symbols not existing in beta is a yellow flag that webhook code and spec are drifting.
- `save` → `savePassword` rename on `ConnectApiSessionStore` — all callers updated (`DefaultTsaNetApiSession`, tests). ✓
- `isAuthorized()` now returns false when a token is *expired* (previously: token present = authorized). Consumers: `ensureAuthenticatedForWebhook` (re-auths, fine), `currentBearerToken` (now returns empty on expiry — behavior change, arguably a fix), CLI `SessionStatusCommand`. A password session has `expiresAt == null` so `isExpired` returns false — password behavior unchanged. Reasonable.

**Operational:**
- New `.github/workflows/publish.yml` requires repo secret `CONNECT_API_CODE_TOKEN` (fails loudly if absent — good) and relies on `GITHUB_TOKEN` for packages. Shawn needs to configure `CONNECT_API_CODE_TOKEN` before the first release, or every publish run red-fails. Call this out.
- New dependency `jackson-datatype-jsr310` — declared in `connect-library/pom.xml`. Version presumably managed by the Boot BOM; no explicit version, so confirm it resolves.
- No `.env`, no ops-sidecar, no schema.sql, no LaunchAgent touched — this is the SDK repo, not the ops runtime. No sidecar restart implication.
- SQLite `test_case` column added via `ensureColumn` migration — handled in-code, no separate migration file needed for this repo's pattern.

**Recommendation:** Merge is fine after two verifications you can do in one build run — (1) the branch actually compiles against the `beta` spec so `WebhooksV1Api`/`getCloudEventId`/`getTestCase` resolve, and (2) no `ApplicationUserAccount`/`WebhookDeliveryDto` consumer was missed by the positional/field-name changes; both are compile-enforced, so a green build is the receipt. The untraced webhook-DTO migration and the missing consumer-enumeration artifacts are the reason this is CONCERNS not LOOKS GOOD, but none of it holds the merge if the build is green and no external repo consumes those two public types directly.

---

shawn-tsanet:
Disposition on the auto-QA review (verdict CONCERNS, no blockers). Per-finding, with receipts:

1. **Untraced webhook hunk** (\`WebhookDeliveryDto.integrationId\` Long → \`cloudEventId\` String, \`webhooksApi\` → \`webhooksV1Api\`): traced. It lands in \`0f57fac\` ("OAuth basic implementation", Slava, 2026-07-30), not in any of the five squashed PRs. It is also what makes this branch compile against the sibling \`beta\` spec while main at \`332e5c7\` does not, exactly as the README branch notes (\`tsanetgit/Connect_SDK#42\`) describe. The PR body now names it.

2. **Compile + test receipt**: \`mvn install\` on the PR head \`f747bc6\`, sibling \`Connect-API-Code\` clone on \`beta\` (\`0691e1f\`), JDK 21. BUILD SUCCESS on all three modules; 122 tests, 0 failures (connect-library 86, demo 6, app 30); zero cannot-find-symbol. \`WebhooksV1Api\`, \`getCloudEventId()\`, and \`getTestCase()\` all resolve; \`ConnectApiWebhooksGatewayTest\` ran green.

3. **External consumers of the changed public types** (\`ApplicationUserAccount\` constructor, \`WebhookDeliveryDto\` field): GitHub code search over \`shawn-tsanet/connect-sdk-demo\` returns zero hits for \`ApplicationUserAccount\`, \`integrationId\`, and \`WebhookDeliveryDto\`; the local \`tsanetgit/Connect_Gateway\` clone (main) has no Java references either. In-repo call sites are compile-enforced by the green build above.

4. **\`CONNECT_API_CODE_TOKEN\` repo secret**: confirmed missing (the repo currently has zero Actions secrets). The publish workflow fires only on release, so nothing breaks at merge time; the secret must be configured before the first release is cut. Open item for Shawn.

5. **Header logging**: no logging interceptor prints headers. \`OAuth401RetryInterceptor\` contains no log statements, and no \`Authorization\` string appears in library main sources outside generated code.

6. **Accepted as-is**: the first-call null bearer for a never-authenticated client-credentials session (costs one 401 round trip, then the interceptor refreshes and retries once, no loop) and the \`testSubmission=true\` default on the deprecated overloads (conservative; no in-repo caller passes \`false\` outside tests, per the write-\`testSubmission\`/read-\`testCase\` asymmetry). The \`tsaet\` typo fixes and comment removals are nit-level tidying, now named in the PR body.

Issue #52, opened 2026-08-12
attachment-receiver: hardening (filename safety, size caps, TLS, rotation)


shawn-tsanet:
Part of the P4 attachment-receive epic: #46.

Filename sanitization (the backend concatenates raw filenames into the URL path with no encoding: reject or normalize traversal sequences, control characters, and empty names), configurable size cap enforced during streaming (not after), TLS-only listener, password rotation without downtime (accept current + next during a rotation window). Filename normalization rules for spaces/non-ASCII await tsanetgit/Connect-API-Code#140 Q2; everything else is buildable now.

Acceptance: traversal and oversize attempts rejected with tests; rotation exercised in a test.

shawn-tsanet:
Scope addition from PR #57's auto-QA review (log hygiene): remote-controlled attachment names currently echo verbatim inside `AttachmentStorageException` messages (e.g. the oversized-name failure includes the full hostile token). Not an injection risk, but exception messages propagate to logs. When this issue lands the filename-handling policy, include a decision on truncating/escaping hostile names in exception and log text across both adapters and the ingest endpoint.

shawn-tsanet:
Moved to `tsanetgit/Connect_Gateway#36` on 2026-09-08, together with the ingest endpoint it hardens (`tsanetgit/Connect_SDK#51` → `tsanetgit/Connect_Gateway#35`). The log-hygiene addition from the #57 review travels with it. Nothing from this issue applies to the storage adapters that stay here; their own name handling shipped with them.

Issue #51, opened 2026-08-12
attachment-receiver: normalized HTTPS ingest endpoint (gated: Connect-API-Code#140 Q1)


shawn-tsanet:
Part of the P4 attachment-receive epic: #46.

The wire surface: accept `PUT {path}/{fileName}` raw bytes, HTTP Basic (username = case number = the only correlation metadata, password = tenant password), resolve tenant from path, stream to the tenant's `AttachmentStorage`, answer 2xx only after storage confirms. Content-Type recorded as metadata. Same-name policy per the epic's design decision (context: Cisco silently ignores same-name uploads, tsanetgit/SFDC_App#76).

**Do not start until tsanetgit/Connect-API-Code#140 Q1 confirms the contract.** If the September common method turns out backend-native (S3/SigV4), this issue is re-scoped, not started.

Acceptance: end-to-end against a real backend push on DEV/BETA; wrong password → 401; unknown tenant path → 404; storage failure → non-2xx (the backend reports FORWARD_FAILED, never a false success).

shawn-tsanet:
Re-scope note (from tsanetgit/Connect-API-Code#140): September's receiver enablement is core-API-native per Paul, so this endpoint's first consumer is NOT a legacy-connector member — it is the gateway runtime (the Fin Phase 4 attachment bridge registers a normalized HTTPS endpoint and receives via the same Cisco-shaped push). Still gated on #140 Q1 (contract stability) and paced by the Fin bridge window (Oct-Nov). The gate holds: do not build the wire surface until Q1 is answered.

shawn-tsanet:
Moved to `tsanetgit/Connect_Gateway#35` on 2026-09-08: the relay-path receiver, this HTTPS ingest endpoint, is built in the gateway's attachments-only profile rather than here. What stays upstream in this repo's `attachment-receiver` module is everything the endpoint consumes: the `AttachmentStorage` SPI, the S3, Azure Files, Azure Blob and GCS adapters, the tenant config model and the go-live verifier, all merged. The gateway consumes them as the published artifact. Epic #46 records the decision.

Issue #49, opened 2026-08-12
attachment-receiver: Azure Files storage adapter


shawn-tsanet:
Part of the P4 attachment-receive epic: #46.

`AttachmentStorage` implementation on the Azure Files SDK. Auth: SAS-scoped to the share, or service principal; both must work credential-supplied (gateway-hosted, member-owned share) and identity-based (member-deployed with managed identity). Streaming upload. Buildable now.

Acceptance: passes the shared adapter contract test against a real share (test subscription); both auth shapes documented.

shawn-tsanet:
Acceptance criterion added from the #47 gate review (PR #54, disposition 1): the shared contract test cannot force rollback-after-partial-commit, because the in-memory reference never commits incrementally. This adapter commits incrementally against the real backend, so its own tests must exercise an abort after a partial commit (fail the stream after at least one part/chunk is written) and prove no partial object stays visible.

Issue #48, opened 2026-08-12
attachment-receiver: AWS S3 storage adapter


shawn-tsanet:
Part of the P4 attachment-receive epic: #46.

`AttachmentStorage` implementation on the AWS SDK v2 S3 client. Auth via the default provider chain so both deployment modes work unchanged: instance/task role when member-deployed, cross-account role assumption (with external ID) when gateway-hosted against a member-owned bucket. Streaming upload, no local spooling. Buildable now.

Acceptance: passes the shared adapter contract test against a real bucket (test account); cross-account path exercised at least once manually and documented.

shawn-tsanet:
Acceptance criterion added from the #47 gate review (PR #54, disposition 1): the shared contract test cannot force rollback-after-partial-commit, because the in-memory reference never commits incrementally. This adapter commits incrementally against the real backend, so its own tests must exercise an abort after a partial commit (fail the stream after at least one part/chunk is written) and prove no partial object stays visible.

PR #54, opened 2026-08-12
attachment-receiver: module skeleton and the AttachmentStorage SPI

### LINKED ISSUE #47: attachment-receiver: module skeleton and the AttachmentStorage SPI

New Maven module `attachment-receiver` alongside `connect-library`. Defines the `AttachmentStorage` SPI: store (stream, not buffer: the backend buffers the whole file in memory on its side, the receiver must not repeat that), object-key layout (case number / filename, exact layout is a design decision recorded on the epic), existence check for the same-name policy, and a verification probe (write + read + delete). No wire surface in this issue. Buildable now; the SPI must not leak ingest-contract assumptions.

Acceptance: module builds in the reactor; SPI has an in-memory test double; adapter contract test exists and runs against the double.

shawn-tsanet:
Closes #47, the first sub-issue of the P4 attachment-receive epic #46.

## What this adds

- New Maven module `attachment-receiver` (registered in the parent reactor; the only change to an existing file is that one `<module>` line).
- The `AttachmentStorage` SPI (`com.tsanet.receiver.storage`, distinct from `connect-library`'s existing SQLite `storage` package): `store` (streaming), `exists` (throwing), `verifyAccess` (go-live probe), with value types `IncomingAttachment` / `StoredAttachment` and checked `AttachmentStorageException`.
- `AttachmentStorageContractTest`, the abstract acceptance bar every adapter (#48 S3, #49 Azure Files) must extend, plus the `InMemoryAttachmentStorage` reference double bound to it.
- Nothing wire-shaped: the ingest endpoint is gated on tsanetgit/Connect-API-Code#140 and no HTTP concept appears in this module.

## Binding rules the SPI encodes (and where each is enforced)

| Claim in the javadoc | Line that makes it true |
|---|---|
| Caller owns the stream; store must not close it | `AttachmentStorageContractTest.storeDoesNotCloseTheCallersStream` |
| A failed store leaves no visible partial object | `AttachmentStorageContractTest.failedStoreLeavesNoPartialVisibility` (see disposition 1) |
| `contentLength` may be `-1` (unknown up front) | `AttachmentStorageContractTest.unknownLengthStreamStores` |
| An indeterminate `exists` throws, never answers false | `exists` signature: `throws AttachmentStorageException` |
| Same-name behavior deliberately unspecified (Connect-API-Code#140 Q2) | No same-name assertion in the contract test, by design |
| `fileName` is remote-controlled; adapters treat it as opaque | SPI javadoc rule; enforcement lands with #52 (hardening) |

## Verification

- `mvn test -pl attachment-receiver`: 7 tests, 0 failures (surefire: `tests="7" errors="0" skipped="0" failures="0"`).
- Probe-the-test: injected two violations into the double (partial visibility before read; closing the caller's stream) — exactly the two guarding tests went red, restored to green.
- Full reactor: `mvn -DskipTests package` BUILD SUCCESS with the new module in the build order.
- JUnit version is parent-managed: `mvn dependency:list` resolves `org.junit.jupiter:junit-jupiter:jar:5.10.3` with no version in this module's pom.

## Advisor (2 calls)

Design call shaped the SPI (store javadoc must not promise `exists()`-based same-name policy — TOCTOU, S3/Azure need conditional writes; `-1` length convention pinned now; stream ownership + no-partial-visibility encoded as tests). Pre-done call raised one blocker, applied: `exists` gained its failure channel so a backend outage can never read as "object absent."

## Pre-PR gate: CONCERNS, dispositions

1. **The no-partial-visibility test passes the in-memory double by construction** — the double never commits before the read finishes, so the rollback path is unexercised, and an incrementally-committing adapter (S3 multipart) could leave a partial object and still pass. Correct observation; not closable generically in a unit contract test. Disposition: carried to #48 and #49 as an explicit acceptance criterion (abort-after-partial-commit exercised against the real backend); comments posted on both issues.
2. **JUnit dependencyManagement unverified by the gate.** Verified here: see the `dependency:list` receipt above.

## Blast radius (generated, `scripts/pre-pr-artifacts.py --base origin/main`)

Detector lines: no positional contracts, no vocabulary changes, no old-term sweeps triggered — this is an additive module; the receive side had no prior form to sweep (receipt: `git grep -n "AttachmentStorage" -- ':!attachment-receiver'` on the branch returns nothing).

PROMPT lines answered:

- **Guard vs representation:** the record compact constructors are constructor invariants on new value types; the records themselves are the representation choice (typed value objects instead of loose strings/longs).
- **Newly reachable failure → receiving handler:** `AttachmentStorageException` today reaches only test code; its designed consumer is the gated ingest endpoint (#51), whose contract is to map every storage failure to a non-2xx so the platform reports `FORWARD_FAILED` instead of a false success. That mapping is #51's acceptance, not this PR's.
- **Abstraction introduced → old-form grep:** the decision "where do received bytes go" had no prior deciders anywhere in the repo (grep receipt above; the receive side did not exist).
- **Schema/contract shape:** the parent reactor grew from 3 to 4 modules; pre-existing state (3 modules) shown working by the full-reactor BUILD SUCCESS above.
- **Prose claims:** every javadoc claim is mapped to its enforcing line in the table above; the one claim a unit test cannot fully enforce is disposition 1.

Third-or-later-fix line: not applicable, new code.

### DIFF
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.tsanet</groupId>
+        <artifactId>tsanet-client-parent</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>attachment-receiver</artifactId>
+    <name>attachment-receiver</name>
+    <description>
+        Receive side of TSANet Connect case attachments: the AttachmentStorage SPI and its
+        storage adapters. The normalized HTTPS ingest endpoint is a separate, gated concern
+        (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/AttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/AttachmentStorage.java
@@ -1,0 +1,70 @@
+package com.tsanet.receiver.storage;
+
+import java.io.InputStream;
+
+/**
+ * Where received attachment bytes go: one implementation per storage backend (AWS S3,
+ * Azure Files, in-memory for tests). Implementations are exercised by the shared
+ * {@code AttachmentStorageContractTest}; an adapter is not done until it passes it.
+ *
+ * <p>This SPI deliberately knows nothing about how bytes arrive. The normalized HTTPS
+ * ingest endpoint is gated on tsanetgit/Connect-API-Code#140 and its wire shape must not
+ * leak in here: no HTTP types, no auth, no path conventions.
+ *
+ * <p>Rules that bind every implementation:
+ * <ul>
+ *   <li><b>Streaming:</b> {@link #store} consumes the stream incrementally. It must not
+ *       buffer the whole file in memory; the platform backend already does that on its
+ *       side and the receiver must not repeat the mistake.</li>
+ *   <li><b>Stream ownership:</b> the caller closes the stream. {@code store} must not.</li>
+ *   <li><b>No partial visibility:</b> when {@code store} throws, a subsequent
+ *       {@link #exists exists(caseNumber, fileName)} must return {@code false}. A failed
+ *       write aborts and cleans up; it never leaves a half-written object visible.</li>
+ *   <li><b>Same-name behavior is implementation-defined for now.</b> Whether a second
+ *       store of the same (caseNumber, fileName) overwrites, versions, or rejects is an
+ *       open contract question (tsanetgit/Connect-API-Code#140, question 2). Callers must
+ *       not build a check-then-store policy on {@link #exists} — that is a race; when the
+ *       answer lands, the policy moves into {@code store} itself (conditional write).</li>
+ *   <li><b>{@code fileName} is remote-controlled input.</b> The platform backend appends
+ *       the sender's filename to the push URL with no encoding, so it may contain path
+ *       separators or traversal sequences. Adapters treat it as an opaque token, never as
+ *       a path to interpret.</li>
+ * </ul>
+ */
+public interface AttachmentStorage {
+
+    /**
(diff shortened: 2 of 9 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
- `pom.xml` — one `<module>` line added registering `attachment-receiver`. Traces to "registered in the parent reactor." ✓
- `attachment-receiver/pom.xml` — new module POM, JUnit test dep with no version (parent-managed). Traces to "New Maven module." ✓
- `AttachmentStorage.java` — the SPI (`store`/`exists`/`verifyAccess`). Traces to "The AttachmentStorage SPI." ✓
- `AttachmentStorageException.java` — checked exception. Traces to "checked AttachmentStorageException." ✓
- `IncomingAttachment.java` / `StoredAttachment.java` — value types with compact-constructor invariants. Traces to "value types." ✓
- `AttachmentStorageContractTest.java` — abstract acceptance bar. Traces to the contract test claim. ✓
- `InMemoryAttachmentStorage.java` / `InMemoryAttachmentStorageTest.java` — reference double + binding. Traces to "InMemoryAttachmentStorage reference double bound to it." ✓

**Correctness:**
- Traced `failedStoreLeavesNoPartialVisibility`: `MidReadFailingStream` serves 4 `'x'` bytes then throws `IOException`. In `InMemoryAttachmentStorage.store`, the read loop catches the `IOException` and throws `AttachmentStorageException` *before* the `objects.put(key, bytes)` line — so nothing is ever inserted, and the subsequent `exists` returns false. Test passes correctly, and by construction the double can't leave a partial (put is the last statement). This is exactly the limitation the PR flags in disposition 1, honestly stated.
- `unknownLengthStreamStores` with `contentLength=-1`: `IncomingAttachment` invariant allows `-1` (`contentLength < -1` rejects only below that); store ignores declared length and returns actual `bytes.length`. Correct.
- `store` catches only `IOException`. If `content.read` threw a `RuntimeException`, it would propagate unwrapped — but for the reference double that path is unreachable, and the SPI javadoc scopes failure to `AttachmentStorageException`, which real adapters must honor. Not a bug in this diff.
- Compact constructor edge: `Objects.requireNonNull(...).isBlank()` — NPE message is "caseNumber"/"fileName"/"storageKey", then `isBlank` for blank-but-nonnull. Both null and blank are rejected. Correct.

**Security:**
- No secrets, no credentials, no external HTTP in the module. `fileName` is documented as remote-controlled and treated as opaque; the `key()` helper concatenates `caseNumber + "/" + fileName` for in-memory map keys only — no filesystem path is derived, so a traversal sequence in `fileName` is inert here. Path hardening is correctly deferred to #52. No concern for this test-only double.

**Regression risk:**
- Additive module. The only edit to existing code is the reactor `<module>` line. No existing function/endpoint/persona touched — sidecar, mirror, invoice paths untouched. `git grep "AttachmentStorage" -- ':!attachment-receiver'` returning nothing (cited in PR) confirms no prior consumer to break.
- Schema-shape edge: reactor grew 3→4 modules; PR cites full-reactor BUILD SUCCESS on pre-existing state. Adequate.

**Recommendation:** Merge — additive skeleton, every hunk traces, the one un-exercisable guarantee (partial-visibility rollback) is honestly disposed to #48/#49; note this is a draft PR, so Shawn merges when he's ready.

---

shawn-tsanet:
Good to merge

Issue #53, opened 2026-08-12
attachment-receiver: config self-registration against the Connect API (gated: Connect-API-Code#140 Q3)


shawn-tsanet:
Part of the P4 attachment-receive epic: #46.

Whatever registration path Andrew lands (opened self-service `PUT .../attachments/config/https`, per-member strategies, or admin provisioning), implement the client side in `connect-library` (upstream-first) and wire it into receiver onboarding: register on enablement, re-register on rotation. Today the endpoint is hardcoded to Cisco (`canHandleUpdateFrom`), so **nothing here is buildable until tsanetgit/Connect-API-Code#140 Q3 answers.**

Acceptance: a non-Cisco test member registers a config and receives a push on DEV/BETA, or the issue is re-scoped to match the landed mechanism.

slava-vasianovych:
@shawn-tsanet looks like we can close this too.

shawn-tsanet:
Closing as superseded; nothing here is buildable and the mechanism it waited on is not the one that will land.

- The gate was answered: on `tsanetgit/Connect-API-Code#140`, Q3 was answered 2026-08-22 with admin provisioning for September.
- The direction then changed (status of 2026-09-04 on `tsanetgit/Connect-API-Code#57`): September is V2 direct delivery (`tsanetgit/Connect-API-Code#147`), with member file-store configuration owned by core under `tsanetgit/Connect-API-Code#146`, provisioned by TSANet admin in September and self-service in November (`tsanetgit/Connect-ReleaseMGT#239`). The HTTPS push config this issue would register survives only as the relay path.
- So the client-side registration flow described here targets an endpoint model that is not being built. The SDK piece that does follow is a client for #146's typed file-store configuration and its test endpoint, wanted when self-service opens in November; that is a new issue filed when #146 has a contract to build against, not a re-scope of this one.

The receiver building blocks this issue assumed (SPI, adapters, config model, go-live verifier) are all merged and unaffected. Epic #46 is updated to say the same.

Issue #46, opened 2026-08-12
Epic: P4 - attachment receive (the attachment-receiver module)


shawn-tsanet:
## Context

The send side of attachments shipped with P1 (`AttachmentsFacade`, CLI, validation; epic #19). The receive side does not exist anywhere in the ecosystem: TSANet's backend pushes forwarded files to whatever HTTPS endpoint a member has registered, and today Cisco is the only member with one (tsanetgit/Connect-API-Code#57). This epic builds the reusable receiver: a normalized HTTPS ingest endpoint behind an `AttachmentStorage` SPI with AWS S3 and Azure Files adapters, consumable by any SDK deployment and hosted by the gateway (tsanetgit/Connect_Gateway) in an attachments-only profile.

Strategy and full contract detail: the attachments strategy doc (v1.0, 2026-08-12; the copy in `tsanetgit/Connect_Gateway` `docs/` is the reference). The observed push contract in brief: `PUT https://{domain}:{port}{path}/{fileName}` with raw bytes, HTTP Basic where username = the case number and password = the registered config password, any 2xx = success, single attempt. Delivery fans out to both the receiver's and the submitter's registered configs, so a receiver also gives sending members their own copies (today every send reports `submitterStatus: CONFIG_MISSING`).

Upstream-first applies (gateway decision D6): all receiver code lands in this repo; the gateway consumes the published artifact verbatim.

## Gate (resolved 2026-09-04; the direction changed)

tsanetgit/Connect-API-Code#140 questions 1 and 3 gated the wire-facing work. Q3 was answered on 2026-08-22 (admin provisioning for September). Then the direction changed: the 2026-09-04 status on tsanetgit/Connect-API-Code#57 sets the September path as **V2 direct delivery** (tsanetgit/Connect-API-Code#147): the sender asks for a grant, uploads straight into the receiver's store, and reports completion; no file bytes pass through TSANet. Member file-store configuration is owned by core (tsanetgit/Connect-API-Code#146), provisioned by TSANet admin in September and self-service in November (tsanetgit/Connect-ReleaseMGT#239). The HTTPS push is kept only as the relay path for receivers with no API-addressable store.

[...]

Issue #26, opened 2026-07-09
Package a TSANet Connect skill for SDK adopters (deploy runbook, adapter cookbook, validation suite)


shawn-tsanet:
## Context

The AWS Q4 integration (`tsanetgit/Member-Integrations#43`) will consume the Connect SDK, and the AWS onboarding proposal (draft under review) estimates the level of effort for a member on two routes: deploying the SDK bridge with a member-built adapter, or building a native connector against the REST API. Both routes lose most of their time to discovery, deployment friction, and test design rather than code volume. A packaged TSANet Connect skill (an expert knowledge base for AI-assisted development, the same mechanism already used for the Zendesk connector) attacks exactly those sinks: estimated reduction is roughly a third off the SDK route and more off the native route.

## Scope

- **Deployment runbook**: container configuration, environment and credential setup, health checks, the single-writer SQLite cache constraint, what a healthy bridge looks like.
- **Adapter cookbook**: facade call sequences with working examples (partner search, form fetch, create; response flows; notes; close), error and ProblemDetail mapping.
- **Business rules not in the OpenAPI spec**: member-domain email requirement, response-flow ordering, webhook signature contract, form handling and known API quirks.
- **Validation suite**: executable end-to-end flows with expected responses and cleanup steps, derived from the existing contract tests already proven against the DEV environment.
- **Format**: markdown knowledge package consumable by AI coding agents and readable as plain documentation.

## Related

- `tsanetgit/Member-Integrations#43` (AWS go-live tracker; the skill should be part of the SDK hand-off package)
- `tsanetgit/Connect_SDK#24` (test and demo plan; source of validated flows)
- `tsanetgit/Connect_SDK#20` (connection hardening; runbook content should reflect the hardened defaults once closed)

shawn-tsanet:
Closing as completed. The four scope items shipped, though not as the single package this issue imagined; the ecosystem split them by audience, which is the better shape.

- **Deployment runbook**: the `connect-sdk-deploy` skill in this repository (`tsanetgit/Connect_SDK#64`, 2026-08-18, kept current through the 2.0.0 doc sweep): access model, build, the three deployment paths, hosting and the mandatory auth gate, and the hardened https defaults from `tsanetgit/Connect_SDK#20`.
[...]

PR #55, opened 2026-08-12
attachment-receiver: tenant config model and go-live verification

### LINKED ISSUE #50: attachment-receiver: tenant config model and go-live verification probes

Per-tenant configuration: routing path segment, push password, storage backend selection + credentials, optional CRM delivery credential. Encrypted at rest; loadable single-tenant (member-deployed) or multi-tenant (gateway-hosted). Go-live verification: storage write+read probe and CRM credential dry-run, run on config save and re-runnable on demand (the Figure 5 flow in the strategy doc). Buildable now.

Acceptance: config round-trips encrypted; probes fail loudly and specifically (wrong-credential vs wrong-target vs no-permission distinguishable).

shawn-tsanet:
Closes #50 (epic #46). **Stacked on #54** — base is `feat/attachment-receiver-skeleton`; GitHub retargets to `main` when #54 merges.

## What this adds

- `TenantConfig`: one member's receive configuration, serving both hosting modes (member-deployed = the single-tenant degenerate case). The routing token is **allowlisted** (`[a-z0-9][a-z0-9-]{0,63}`) in the record *and* re-validated on the store's read path, making it safe as a URL path segment and a filename in one move.
- `EncryptedFileTenantConfigStore`: AES-256-GCM file-per-tenant at rest, random IV per write, the routing token bound as GCM AAD (swapped files fail at decrypt), atomic temp-file + `ATOMIC_MOVE` saves, `keyFromBase64` failing at startup on a wrong-length key.
- `GoLiveVerifier` + `VerificationReport`: the Figure-5 checks from the strategy doc — build the configured storage and prove it via the SPI's `verifyAccess()` probe; dry-run CRM credentials when delivery is configured (`CrmProbe` is a seam; implementations arrive with the delivery adapters). Checks fail independently with the underlying message passed through verbatim.
- `StorageFactory`: adapters register by string id; unknown id surfaces as a STORAGE failure naming it.

## Deliberate decisions (stated, not implied)

- **`all()` is fail-fast: one unreadable tenant file fails whole-directory enumeration.** By design for now — an unreadable config is an operational emergency in hosted mode, and silently skipping it would hide exactly the tenant whose files are at risk. If operating experience wants skip-and-collect, that's a follow-up with its own report shape. `byPathSegment` (the per-push path) isolates failures per tenant.
- **Windows reserved names (`con`, `nul`) pass the allowlist.** The receiver deploys on Linux containers; the rejection test says so explicitly rather than implying protection that doesn't exist.
- **Credentials as `String`, not `char[]`:** zeroing is theater here (values transit `Properties` and cipher buffers regardless); the enforced protections are encryption at rest and full redaction of `toString` (password *and* both property maps' values — only keys print).

## Verification

- 28 tests green (surefire: TenantConfigTest 6, EncryptedFileTenantConfigStoreTest 10, GoLiveVerifierTest 5, plus the 7 SPI tests from #54).
- The swap test asserts the **mechanism**, not just the outcome: the failure message must come from the decrypt ("integrity check"), proving the AAD binding acted — if AAD were dropped, the post-decrypt cross-check's distinct message would appear and the test fails.
- Ciphertext probed for credential bytes (password, CRM token, storage property) — none on disk.
- Bit-flip tamper, wrong-key, traversal-on-read (`../escape`), rotation-replaces-atomically, and no-temp-file-residue all covered.

## Advisor (2 calls)

Design call: switched routing-token validation from blocklist to allowlist (dot-dot, backslash, case-collisions), required `toString` redaction with a test, required AAD binding + file-swap test, atomic saves, 32-byte key check with byte count in the message, per-check `Exception` isolation. Pre-done call: strengthened the swap test to assert the AAD mechanism (applied), temp-file cleanup on failed save (applied), honest comment on Windows reserved names (applied), and the `all()` decision stated above.

## Pre-PR gate

Exit 0 ("Merge"). Its one note is the `all()` fail-fast behavior, dispositioned above as a deliberate decision.

Net-new packages (`config`, `verify`); no existing symbol changed; detector lines all clean. PROMPT lines answered:

- **Guard vs representation:** the allowlist IS the representation change (routing token defined as a grammar, not a sanitized string); the record validators are constructor invariants.
- **Newly reachable failures → handler:** `TenantConfigException` and verification FAILs are consumed today by tests; their designed consumers are the ingest endpoint (#51: unreadable tenant → non-2xx) and the onboarding flow (#53 / gateway #26: a failing report blocks registration). Both are those issues' acceptance, not this PR's.
- **Abstraction introduced:** "which backend stores this tenant's files" had no prior decider (`git grep -n "storageBackend" -- ':!attachment-receiver'` → nothing).
- **Schema/stored shape:** the encrypted file format is introduced here (IV || ciphertext, Properties inside, AAD = token). No pre-existing files can exist; version-marking the format is deferred until a second format exists to distinguish.
- **Prose claims:** every claim in the class javadoc maps to a test named in Verification above; the commit message's "swapped files fail at decrypt" is made true by the mechanism assertion.

### DIFF
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.tsanet</groupId>
+        <artifactId>tsanet-client-parent</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>attachment-receiver</artifactId>
+    <name>attachment-receiver</name>
+    <description>
+        Receive side of TSANet Connect case attachments: the AttachmentStorage SPI and its
+        storage adapters. The normalized HTTPS ingest endpoint is a separate, gated concern
+        (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/config/EncryptedFileTenantConfigStore.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/config/EncryptedFileTenantConfigStore.java
@@ -1,0 +1,206 @@
+package com.tsanet.receiver.config;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.stream.Stream;
+
+/**
+ * File-per-tenant store, AES-256-GCM at rest. {@code <pathSegment>.tenant} under one
+ * directory; the file body is a random 12-byte IV followed by the ciphertext. The
+ * tenant's path segment is bound into the ciphertext as GCM additional authenticated
+ * data, so two tenants' files cannot be swapped on disk and still decrypt. Writes go to
+ * a temp file first and land by atomic move: a crash mid-save can never leave a torn
+ * file that reads as tampering later.
+ */
+public final class EncryptedFileTenantConfigStore implements TenantConfigStore {
+
+    private static final String FILE_SUFFIX = ".tenant";
+    private static final int IV_BYTES = 12;
+    private static final int TAG_BITS = 128;
+    private static final int KEY_BYTES = 32;
+
+    private final Path directory;
+    private final SecretKey key;
+    private final SecureRandom random = new SecureRandom();
+
+    public EncryptedFileTenantConfigStore(Path directory, SecretKey key) throws TenantConfigException {
+        this.directory = directory;
+        this.key = key;
+        try {
+            Files.createDirectories(directory);
+        } catch (IOException e) {
+            throw new TenantConfigException("cannot create config directory " + directory, e);
+        }
+    }
+
+    /**
+     * The deployment-reality key path: a base64 value from an environment variable or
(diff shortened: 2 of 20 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:** All 11 files are net-new, no existing symbol touched (PR's blast-radius section claims this; consistent with the diff — every hunk is `new file mode`). Tracing each to the PR body:

- `TenantConfig.java` + `TenantConfigTest.java` → "TenantConfig: one member's receive configuration" + allowlist. Traces.
- `EncryptedFileTenantConfigStore.java` + test → "AES-256-GCM file-per-tenant at rest". Traces.
- `TenantConfigStore.java`, `TenantConfigException.java` → SPI seam for the store. Traces.
- `GoLiveVerifier.java`, `VerificationReport.java`, `CrmProbe.java`, `StorageFactory.java` + `GoLiveVerifierTest.java` → "GoLiveVerifier + VerificationReport: the Figure-5 checks". Traces.

No incidental tidying; nothing untraced.

**Correctness:**

Traced the swap-defense path (`swappedTenantFilesDoNotDecrypt`): save `acme`, save `globex`, `Files.copy(globex.tenant → acme.tenant)`, then `byPathSegment("acme")`. In `read()`, cipher inits with AAD = `"acme"` but the ciphertext's GCM tag was computed over AAD `"globex"` → `doFinal` throws `AEADBadTagException` → caught as `GeneralSecurityException` → message contains "integrity check". The test asserts the message names "integrity", confirming the failure came from decrypt not the post-decrypt cross-check. Correct, and the mechanism assertion is real (not tautological).

Traced `all()` truncation edge: `read()` guards `body.length <= IV_BYTES` before slicing `doFinal(body, IV_BYTES, body.length - IV_BYTES)`. A 12-byte body (IV only, no ciphertext/tag) is rejected as truncated rather than passing a zero-length ciphertext to GCM. Correct.

One real edge worth noting (non-blocking): `deserialize` calls `props.getProperty("pathSegment")` etc., which return `null` if the decrypted Properties are missing those keys. That `null` flows into the `TenantConfig` constructor, where `Objects.requireNonNull` throws — caught by the `catch (RuntimeException e)` and re-wrapped as `TenantConfigException`. So a corrupt-but-decryptable file (only reachable with the correct key, i.e. self-written) fails loudly with a clear message. Handled.

The `all()` fail-fast decision is stated and defensible for hosted mode. Not a bug.

**Security:**

- No secrets in the diff — test credentials (`push-secret-042`, `crm-token-042`) are fixtures, not real.
- `toString` redaction verified by `toStringRedactsEveryCredential`; only `keySet()` prints for both maps and password is `<redacted>`. Good.
- Path-traversal defense on the read path (`byPathSegment` re-validates against `PATH_SEGMENT` before `directory.resolve`) is present and tested (`../escape`, `null`). The allowlist `[a-z0-9][a-z0-9-]{0,63}` cannot produce `.`, `/`, or `\`, so `resolve` stays within `directory`. Correct.
- AAD binding + random IV per write: sound GCM usage. IV is prepended in cleartext (standard, fine — IV is not secret).
- Ciphertext-has-no-plaintext-credentials is tested against ISO-8859-1 bytes. Good.

No concerns.

**Regression risk:**

Net-new packages (`config`, `verify`) stacked on #54's `storage` SPI. No existing symbol modified, so no consumer of a changed signature to enumerate. Depends on `AttachmentStorage`, `AttachmentStorageException`, `InMemoryAttachmentStorage` from #54 — those must exist on the base branch (`feat/attachment-receiver-skeleton`) for this to compile. The PR is a draft stacked on #54; that's the expected ordering. I have not compiled against the base branch this session — flagging as the one thing to confirm before/at merge, but it's an inherent property of the stack, not a defect in this diff.

**Recommendation:** Merge once #54 has landed on the base branch (compile dependency); the diff itself is clean, well-tested, and every hunk traces.

---

shawn-tsanet:
Good to merge

PR #56, opened 2026-08-12
attachment-receiver: AWS S3 storage adapter

### LINKED ISSUE #48: attachment-receiver: AWS S3 storage adapter

shawn-tsanet:
Closes #48 (epic #46). **Stacked on #54** — base is `feat/attachment-receiver-skeleton`; parallel to #55.

## Merge precondition: SATISFIED (2026-08-13)

The live-bucket acceptance ran green against a private Beta test account (per the 2026-08-13 decision: Beta acceptance runs on private test accounts; corporate accounts arrive with Production). Full branch suite **50/50**, live class **8/8**: the 7 shared SPI contract tests plus abort-after-partial-commit, against a real bucket with the lifecycle rules the adapter's javadoc recommends.

**The run caught one real defect, fixed on this branch:** the Boot parent manages httpclient5 5.3.x while the AWS SDK's apache5 HTTP client requires `TlsSocketStrategy` (httpclient5 5.4+), so every real `S3Client` construction died with `NoClassDefFoundError` — invisible offline because the stub never builds a real client. The module now overrides Boot's httpclient5/httpcore5 versions (pom comment carries the why). **Any Boot-based consumer of this module needs the same override**; noted on the gateway's contract-extension issue.

Operator run note: credentials from the CLI's browser sign-in are not visible to the Java SDK's default chain — bridge with `aws configure export-credentials --format env` before running the suite.

**Deferred with disposition:** the issue's cross-account exercise (assume-role with ExternalId) awaits an IAM role in the Beta account and moves to the wiring work, where the hosted-mode credential path gets its first real consumer. No adapter code is affected: the client is injected and role assumption is wiring by design.

**Account-hygiene rule, binding for this public repo and its published artifacts:** bucket names, account identifiers, and credentials never appear in code, commits, PR text, or artifacts (same rule as #57).

## What this adds

- `S3AttachmentStorage` (new dependency: `software.amazon.awssdk` BOM 2.46.7 + `s3`, version verified on Maven Central today; the issue text authorizes the dependency). Client injected: region and cross-account role assumption are wiring, not adapter code.
- Keys are `[prefix/]caseNumber/fileName`; `keyPrefix` serves hosted-mode tenant namespacing in a shared bucket and per-run test isolation. `fileName` stays opaque per the SPI (S3 keys have no traversal semantics — documented).
- Streaming: 5 MiB buffers (5,242,880 — the real S3 minimum part size, not 5 MB). EOF inside the first buffer → one atomic `PutObject`; otherwise multipart, aborted on any failure. Incomplete multipart uploads are never visible as objects, so **no-partial-visibility holds structurally**, and the SDK only ever sees replayable byte bodies — it never touches (and can never close) the caller's stream.
- `exists()` discriminates on HTTP status: 404 → false, anything else throws (HEAD has no error body, so the typed exception is unreliable across SDK versions; a 403 from missing `s3:ListBucket` throws rather than reading as absent).
- `verifyAccess()`: random sentinel under the prefix, write→read→delete, classifying wrong-credential (`InvalidAccessKeyId`/`SignatureDoesNotMatch`/`ExpiredToken`) vs wrong-target (`NoSuchBucket`/404) vs no-permission (`AccessDenied`) vs connectivity (`SdkClientException`); the sentinel is best-effort deleted even when the probe fails.
- Class javadoc carries the required IAM set (`PutObject`, `GetObject`, `DeleteObject`, `ListBucket`, `AbortMultipartUpload`), the missing-bucket-404-on-HEAD ambiguity (`verifyAccess` is the check that catches it), and the recommended `AbortIncompleteMultipartUpload` lifecycle rule.

## The two hard edges, and their tests

1. **Abort also failing:** whatever killed the upload usually kills the abort; the abort failure rides as suppressed, the original stays primary (`abortFailureIsSuppressedNotPrimary`). The invariant survives — an un-aborted incomplete upload is still never visible.
2. **Ambiguous complete** (failed client-side, committed server-side): detected as abort→`NoSuchUpload` + object visible, reported as the success it was (`ambiguousCompleteThatActuallyCommittedIsReportedAsSuccess`); the check is guarded to the complete phase only, so an upload-phase failure over a pre-existing (overwritten) object can never masquerade as a commit (`ambiguousCompleteWithNoVisibleObjectStillThrows` covers the throw side).

## Verification

- 14 offline unit tests green against a recording stub (SDK v2's default-throwing interface methods make the stub cheap): path selection at the exact 5 MiB boundary, single-part multipart on boundary-EOF, empty stream, dribbling one-byte reads, mid-stream death → abort + nothing pending + nothing visible, both ambiguous-complete directions, 404/403/outage `exists` mapping, prefixed key symmetry between store and exists, three-way `verifyAccess` classification, sentinel cleanup on success *and* failure paths.
- 7 SPI contract tests still green; live suite correctly reports 8 skipped without the env var.
- BOM 2.46.7 resolves and compiles (the green build is the receipt).

## Advisor (2 calls)

Design call set the mechanics (5 MiB exactly, `readNBytes` fill, boundary-EOF single-part multipart, keyPrefix now, status-code-based `exists`, IAM documentation, abort-suppression, the completing-phase guard on ambiguous-complete, sentinel randomization). Pre-done call: sentinel cleanup on failure paths (applied + test), teardown aborts lingering uploads (applied), double-wrap message fix (applied), and the merge-precondition sentence above.

## Pre-PR gate: CONCERNS, dispositions

- **Live acceptance unrun** → the merge precondition at the top; not claimed, tracked.
- **`S3_CONTRACT_TEST_BUCKET` documentation** → test-only env var; documented in the live test's javadoc and this PR. The repo has no test-env manifest to add it to.

Additive package `storage.s3` + one pom dependency block; no existing symbol changed. PROMPT lines: the adapter is the first implementation of an existing abstraction (no old form to sweep — `git grep "S3" -- ':!attachment-receiver'` finds no prior S3 code in the repo); newly reachable failures all surface as `AttachmentStorageException` whose designed handler remains the gated ingest endpoint (#51); no schema/stored-shape change (S3 object layout is introduced here, no pre-existing objects can exist); prose claims are mapped to tests in the sections above.

### DIFF
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -1,0 +1,28 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>com.tsanet</groupId>
+        <artifactId>tsanet-client-parent</artifactId>
+        <version>0.0.1-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>attachment-receiver</artifactId>
+    <name>attachment-receiver</name>
+    <description>
+        Receive side of TSANet Connect case attachments: the AttachmentStorage SPI and its
+        storage adapters. The normalized HTTPS ingest endpoint is a separate, gated concern
+        (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
+    </description>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -18,7 +18,23 @@
         (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
     </description>
 
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>software.amazon.awssdk</groupId>
+                <artifactId>bom</artifactId>
+                <version>2.46.7</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
     <dependencies>
+        <dependency>
+            <groupId>software.amazon.awssdk</groupId>
+            <artifactId>s3</artifactId>
+        </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter</artifactId>
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -18,6 +18,15 @@
         (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
     </description>
 
+    <properties>
+        <!-- The Boot parent pins httpclient5 5.3.x; the AWS SDK's apache5 HTTP client
+             needs TlsSocketStrategy, which exists only from httpclient5 5.4. Without
+             this override every real S3Client construction dies with
+             NoClassDefFoundError - found by the live acceptance, invisible offline. -->
+        <httpclient5.version>5.6.1</httpclient5.version>
+        <httpcore5.version>5.4.2</httpcore5.version>
+    </properties>
+
     <dependencyManagement>
         <dependencies>
             <dependency>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/config/EncryptedFileTenantConfigStore.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/config/EncryptedFileTenantConfigStore.java
@@ -1,0 +1,206 @@
+package com.tsanet.receiver.config;
+
+import javax.crypto.Cipher;
+import javax.crypto.SecretKey;
+import javax.crypto.spec.GCMParameterSpec;
+import javax.crypto.spec.SecretKeySpec;
+import java.io.ByteArrayInputStream;
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.StandardCopyOption;
+import java.security.GeneralSecurityException;
+import java.security.SecureRandom;
+import java.util.ArrayList;
+import java.util.Base64;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Properties;
+import java.util.stream.Stream;
+
+/**
+ * File-per-tenant store, AES-256-GCM at rest. {@code <pathSegment>.tenant} under one
+ * directory; the file body is a random 12-byte IV followed by the ciphertext. The
+ * tenant's path segment is bound into the ciphertext as GCM additional authenticated
+ * data, so two tenants' files cannot be swapped on disk and still decrypt. Writes go to
+ * a temp file first and land by atomic move: a crash mid-save can never leave a torn
+ * file that reads as tampering later.
+ */
+public final class EncryptedFileTenantConfigStore implements TenantConfigStore {
+
+    private static final String FILE_SUFFIX = ".tenant";
+    private static final int IV_BYTES = 12;
+    private static final int TAG_BITS = 128;
+    private static final int KEY_BYTES = 32;
+
+    private final Path directory;
+    private final SecretKey key;
+    private final SecureRandom random = new SecureRandom();
+
+    public EncryptedFileTenantConfigStore(Path directory, SecretKey key) throws TenantConfigException {
+        this.directory = directory;
+        this.key = key;
+        try {
+            Files.createDirectories(directory);
+        } catch (IOException e) {
+            throw new TenantConfigException("cannot create config directory " + directory, e);
+        }
+    }
+
+    /**
+     * The deployment-reality key path: a base64 value from an environment variable or
+     * secret mount. Fails at startup, loudly, on anything but exactly 32 decoded bytes —
+     * a truncated key discovered at first tenant read is a debugging trap.
+     */
+    public static SecretKey keyFromBase64(String base64) {
+        byte[] decoded = Base64.getDecoder().decode(base64);
+        if (decoded.length != KEY_BYTES) {
+            throw new IllegalArgumentException(
+                    "AES-256 key must decode to exactly " + KEY_BYTES + " bytes, got " + decoded.length);
+        }
+        return new SecretKeySpec(decoded, "AES");
+    }
+
+    @Override
+    public Optional<TenantConfig> byPathSegment(String pathSegment) throws TenantConfigException {
+        if (pathSegment == null || !TenantConfig.PATH_SEGMENT.matcher(pathSegment).matches()) {
+            throw new IllegalArgumentException(
+                    "not a valid routing token: '" + pathSegment + "'");
+        }
+        Path file = directory.resolve(pathSegment + FILE_SUFFIX);
+        if (!Files.exists(file)) {
+            return Optional.empty();
+        }
+        return Optional.of(read(file, pathSegment));
+    }
+
+    @Override
+    public List<TenantConfig> all() throws TenantConfigException {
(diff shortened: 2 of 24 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
- `pom.xml` dependencyManagement + s3 dependency (lines +21/+37) → traces to "new dependency: `software.amazon.awssdk` BOM 2.46.7 + s3". OK.
- `S3AttachmentStorage.java` (whole file) → the adapter itself, the core of the PR. OK.
- `S3AttachmentStorageLiveTest.java` → issue-mandated live acceptance + abort-after-partial-commit. OK.
- `S3AttachmentStorageTest.java` → 14 offline tests. OK.
- `StubS3Client.java` → recording stub the offline tests need. OK.

All hunks trace. No incidental tidying.

**Correctness:**

Traced `store()` with a 12 MiB stream (2×PART_SIZE + slack):
- `fill` fills first buffer to PART_SIZE → `firstLength == PART_SIZE` → not `< PART_SIZE` → `multipart`.
- part 1 uploaded from `firstBuffer` (length == PART_SIZE, `last = false`).
- part 2: `partNumber>0` → refill, gets remainder < PART_SIZE → `last = true`, uploaded. `complete` runs. Byte count correct. ✓

Traced boundary EOF (exactly PART_SIZE): part 1 uploaded, `last=false`; loop refills, `fill` returns 0 → `break`. Single-part multipart completes. Matches `exactlyOneBufferGoesMultipartWithASinglePart`. ✓

One real gap worth naming (Correctness / blast-radius edge #5 — newly reachable failure):

- **`resolveAmbiguousComplete` swallows the abort-succeeded case into a throw with no object check.** If `completeMultipartUpload` fails client-side, the upload was *not* committed server-side, and abort *succeeds* (the normal case: genuine complete failure), the code falls through to `throw wrap("complete", ...)`. That's correct. But the ambiguous-success path is only reached when abort throws `NoSuchUploadException`. Consider the race where complete committed server-side **and** the SDK's own internal retry already reaped/aborted differently — not a concern for this diff, the covered directions are the two that matter. No change requested; the two tests (`ambiguousComplete...ReportedAsSuccess`, `...StillThrows`) cover both documented directions. Noting only that the "abort succeeds but object actually committed" corner is unhandled-by-design — acceptable given S3 semantics (successful abort ⇒ not committed).

- **`existsKey` in the ambiguous path can itself throw non-Exception... it can't** — `existsKey` only throws `AttachmentStorageException`, caught by `catch (Exception headFailure)`. Fine.

- **`verifyAccess` read-length check** (line ~ `readAllBytes().length != probe.length`): throws a bare `AttachmentStorageException` with no `stage`-classified message, but `written && !"delete".equals(stage)` still runs cleanup since stage=="read". Sentinel gets deleted. ✓ Consistent with `verifyAccessCleansUpItsSentinelEvenWhenTheProbeFails`.

**Security:**
- No secrets in diff. Credentials come from the default provider chain / injected client — none hardcoded. ✓
- Probe payload is a fixed literal string, not sensitive.
- Key construction concatenates `caseNumber`/`fileName` — the javadoc correctly notes S3 keys have no traversal semantics, so `..`/`/` only create segments. No injection surface (no SQL, no shell). ✓
- No log statements printing anything. ✓

**Regression risk:**
- Purely additive: new package `storage.s3`, new pom dependency. `git grep "S3"` claim in PR body asserts no prior S3 code — I have not run that grep this session, so I'm taking the additive-only claim as stated rather than verified. Given the diff is 5 new files + one additive pom block with no existing symbol touched, blast radius is genuinely contained regardless.
- New dependency `awssdk:s3` via BOM 2.46.7. This pulls a large transitive tree (netty/http client, etc.). Worth confirming it doesn't collide with any other module's HTTP stack — but this module is the attachment-receiver skeleton, so likely isolated. Follow-up, not blocking.
- `AttachmentStorageException` is the declared handler contract (per #51 gated ingest endpoint). All new failure paths surface as that type — consistent with the existing SPI. ✓

**Operational:**
- New dependency added to `pom.xml` ✓ (present in diff).
- `S3_CONTRACT_TEST_BUCKET` is test-only; no `.env.example` needed. Reasonable disposition.
- **The live acceptance suite has not run** — this is the author-stated merge precondition, and it is the correct gate. The diff cannot be considered done until `S3AttachmentStorageLiveTest` runs green against a real bucket. This is the reason for CONCERNS rather than LOOKS GOOD: the PR is draft and self-gates on it, so it's a tracked follow-up, not a request to re-push.
- No LaunchAgent / schema / sidecar impact.

**Recommendation:** Merge is fine once the author-stated live-bucket suite runs green (the draft self-gates on exactly that); nothing in the code itself blocks — track the AWS SDK transitive-dependency check as a post-merge follow-up.

---

shawn-tsanet:
Wait for Credentials and initial authentication

shawn-tsanet:
Dispositions for the auto-QA CONCERNS, item by item:

1. **Live-bucket acceptance unrun (the CONCERNS driver):** agreed and already the PR's stated merge gate — now explicitly on hold until a TSANet-owned AWS account exists (decision 2026-08-12, see the updated precondition section; TSANet test infra does not run on personal cloud accounts). Tracked on the epic (#46 status comment).
2. **AWS SDK transitive tree colliding with other modules' HTTP stack:** verified isolated — `mvn dependency:list -pl connect-library` resolves zero `software.amazon.awssdk` artifacts, and the BOM import lives only in `attachment-receiver/pom.xml` (module-scoped dependencyManagement cannot perturb sibling modules' resolution). Closed with that receipt rather than carried as a follow-up.
3. **`git grep` additive-only claim taken as stated:** the run behind it: `git grep -n '"S3"\|S3Client' -- ':!attachment-receiver'` on the branch returns 0 hits.

No code changes pushed in response; nothing in the review asked for one.

shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
The diff shown is not the S3 adapter the PR describes — it is a wholly different set of files: `TenantConfig`, `TenantConfigStore`, `EncryptedFileTenantConfigStore`, the `verify/*` go-live seam, and their tests. None of these trace to "attachment-receiver: AWS S3 storage adapter" in the PR title/body. Either the PR scope silently expanded to include tenant-config + go-live verification, or the diff was assembled from the wrong branch. This is the finding: 8 new files of encrypted config storage and go-live verification arrived under an S3-adapter PR with no mention in the description. Flagging, not blocking — the code is additive and self-contained, but the scope mismatch needs a one-line answer from the author before merge.

**Correctness:**

Traced `EncryptedFileTenantConfigStore.save`→`read` round-trip for `acme`:
- save: 12-byte IV + GCM ciphertext with `pathSegment` as AAD, temp file, atomic move. ✓
- read: length guard (`<= IV_BYTES`), decrypt with same AAD, then post-decrypt cross-check that decrypted `pathSegment` equals the filename-derived segment. Covered by `swappedTenantFilesDoNotDecrypt`. ✓

One real gap (Correctness):
- **`deserialize` can NPE-free but produce an invalid record silently on a partial file.** `props.getProperty("pathSegment")` returns null if the key is absent; `TenantConfig`'s constructor calls `Objects.requireNonNull(pathSegment)` → NPE, which is a `RuntimeException`, caught by the `catch (RuntimeException e)` wrapper → `TenantConfigException`. So it's handled. ✓ But note the message path: a null `pushPassword` throws NPE *before* the blank check, also caught. Fine — traced, no bug.
- **`all()` propagates one tenant's corruption as a whole-list failure.** `read` throwing inside the loop aborts `all()` entirely — one tampered file makes every tenant unlistable. For a multi-tenant hosted receiver that's a real availability edge, but there are no consumers of `all()` in this diff, so it's design-latent, not a bug here. Follow-up.

**Security:**
- AES-256-GCM, random per-write IV, AAD binding, atomic move — sound at-rest design. ✓
- `toString()` redacts password and both property-map values; `toStringRedactsEveryCredential` asserts it. ✓
- Key required to be exactly 32 bytes, fails loud at startup. ✓
- No secrets in diff (test credentials are literal fixtures). ✓

**Regression risk:**
Purely additive — new `config` and `verify` packages, no existing symbol touched. No consumers wired in yet (`StorageFactory`/`CrmProbe` are seams with no production implementations in this diff). Contained.

**Carried forward:**
- Live-bucket acceptance unrun — FOLLOW-UP (on hold pending TSANet AWS account, per disposition)
- AWS SDK transitive collision — RESOLVED (`mvn dependency:list -pl connect-library` = 0 awssdk artifacts, cited)
- git grep additive-only claim — RESOLVED (0 hits, cited)

**Recommendation:** Merge is fine once the author confirms the tenant-config/go-live scope belongs in this PR (or points to where it was requested); the code itself is clean and additive.

---

PR #57, opened 2026-08-12
attachment-receiver: Azure Files storage adapter

### LINKED ISSUE #49: attachment-receiver: Azure Files storage adapter

shawn-tsanet:
Closes #49 (epic #46). **Stacked on #54** — base is `feat/attachment-receiver-skeleton`; parallel to #55 and #56.

## Merge precondition: SATISFIED (2026-08-13)

The issue-mandated live-share acceptance ran green against a private Beta test account (per the 2026-08-13 decision: Beta acceptance runs on private test accounts; corporate accounts arrive with Production). Full module suite **59/59**, live class **8/8**: the 7 shared SPI contract tests plus abort-after-partial-commit, ~12s against a real share. The run also proves the two SDK facts the offline suite could not: rename destinations are share-root-relative as implemented, and resize-per-chunk behaves correctly against the real service.

**Account-hygiene rule, binding for this public repo and its published artifacts:** storage resource names, connection strings, and account/subscription identifiers never appear in code, commits, PR text, or artifacts. All configuration is injected or env-var-read; acceptance results are reported as counts, never as resource names.

## What this adds

Azure Files gives **neither** of the invariants S3 gave structurally, so this adapter builds both — that's the whole design:

- **No partial visibility, built:** an Azure file is visible from `create`, so bytes stream into `.incoming-<uuid>` in the destination directory and reach the final name only via rename-with-replace. Failed stores best-effort delete their temp (suppressed if that also fails); the final name is never touched.
- **Traversal-inert names, built:** the share has real directory semantics and Windows naming rules, so both path components pass through a canonical percent-encoder (`AzureFileNames`): leading dots always encoded (making temp-namespace impersonation impossible), trailing dot/space encoded (Azure silently *strips* trailing dots by default — unencoded, store/exists symmetry would break), `%` always encoded, uppercase hex, 255-component cap failing fast.
- **Reserved device names neutralized** (`CON`, `NUL`, `PRN`, `AUX`, `COM1-9`, `LPT1-9`, `CLOCK$`, with or without extension, case-insensitive): I verified against the Azure Files naming doc that the service rejects these outright as file names — the advisor's initial recall said they'd pass; the doc says otherwise, so the encoder encodes their first character. This is the check-before-proclaim catch of the PR.
- Unknown-length streaming: `create(0)` then resize-and-range per 4 MiB chunk (Azure Files has no append primitive; ranges cap at 4 MiB).
- Ambiguous rename mirrors the S3 adapter's ambiguous-complete policy: threw, but final present and temp gone → the rename happened, reported as success.
- Auth is wiring: inject a `ShareClient` (SAS URL, connection string, or TokenCredential/managed identity with `ShareTokenIntent` — all three documented in the javadoc). New dependency: `com.azure:azure-sdk-bom` 1.2.35 (latest stable, verified on Maven Central today) + `azure-storage-file-share` (resolves 12.26.0); issue-authorized.
- **Scope parity with #56:** adapter-only, like the S3 PR. Factory registration (`StorageFactory` id → adapter) arrives with the wiring work, not here.

## Testing shape (why there's a seam)

The SDK's clients are final classes, so rather than adding mockito (a new dependency), the adapter is written against a thin package-private seam (`AzureShare`), 1:1 with SDK calls. The offline double is deliberately **strict** — it enforces fixed-size-at-create, no-append, the 4 MiB range cap, visibility-from-create, parent-required, rename semantics, and throws on any raw traversal component. That last one matters: the traversal test passing proves the *encoder* neutralized the hostile name, not that the fake was lenient. `SdkAzureShare` stays logic-free so the live contract run genuinely covers it.

## Verification

- 30 offline tests green: 14 adapter (temp lifecycle, both ambiguous-rename directions, suppressed cleanup, overwrite policy, traversal + temp-impersonation + oversized-name handling, exists-throws-on-outage, verify classification and sentinel cleanup on success and failure) + 9 encoder (round-trips, reserved names, leading/trailing rules, canonical `%`, unicode, 255 cap) + 7 SPI contract via the in-memory reference. Live suite reports 8 skipped without its env vars, as designed.
- One remote-triggerable unchecked-exception escape found by the pre-done advisor call and fixed: the encoder's 255-cap `IllegalArgumentException` now surfaces through the SPI's checked exception in both `store` and `exists`, with a test pinning the contract.

## Advisor (2 calls) + gate

Design call: share-root-relative rename paths, leading-dot encoding, strict-double requirement, case-insensitivity folded into the upstream same-name question (documented in the javadoc). Pre-done call: the IAE escape (fixed + test), `tempCreated` guard so a failed create doesn't attach a meaningless suppressed delete (fixed), rights-list doc correction (fixed), and the reserved-name verification request that turned into the encoder fix above. Gate: CONCERNS — its one code note (length-only verify read) is applied as a content compare; its other item is the live-run precondition already stated up front.

Additive package `storage.azure` + one pom dependency block; no existing symbol changed (`git grep -i "azure" -- ':!attachment-receiver'` on the base branch: no prior Azure code). Newly reachable failures surface as `AttachmentStorageException` toward the gated ingest endpoint (#51), same as the other adapters. The store/exists "can never disagree" claim is asserted by a test, not prose.


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
All hunks are additive under new package `storage.azure`, tracing to "Azure Files storage adapter" (PR #57, issue #49):
- `pom.xml` dependencyManagement + `azure-storage-file-share` → issue-authorized new dependency, traces.
- `AzureFileNames.java` → traversal-inert name encoding, traces.
- `AzureFilesAttachmentStorage.java` → the adapter itself, traces.
- `AzureShare.java` / `SdkAzureShare.java` → the SDK seam described in "Testing shape", traces.
- Four test files → offline + live suites described in Verification, trace.
No untraceable hunks. No changes to existing symbols (consistent with the `git grep -i azure` claim in the PR — I have not independently run it, but the diff shows only new files + one additive pom block).

**Correctness:**
Traced a real input: `store(caseNumber="01234567", fileName="../../etc/passwd")` with 4 bytes.
- `directoryFor` → `AzureFileNames.encode("01234567")` = `"01234567"` (all pass-through), prefixed → `"tenants/acme/01234567"`.
- `encode("../../etc/passwd")`: leading `.` at i=0 not interior → `%2E`, subsequent `.` `/` all non-pass-through → encoded. Result is a single component with no `/`. finalPath = `tenants/acme/01234567/%2E%2E...`. Good — matches the test assertion. `exists` re-encodes identically. Correct.

Empty-stream path: `fill` returns 0 on first read → `break` immediately, `total=0`, temp created at size 0, rename to final. Zero-byte file lands. Correct (test confirms).

One real concern — **`resizeFile` uses `total + length` cumulatively, and the SDK `setProperties(newSize,...)` truncates-or-grows the whole file each chunk.** For multi-chunk this is fine (monotonic growth). But note the offline double's `resizeFile` does `Arrays.copyOf`, preserving prior bytes; the real Azure `setProperties` resize to a *larger* size also preserves existing content, so behavior matches. No bug, but this is exactly the "behavior at large resizes" the PR flags as only provable live — correctly gated.

Reserved-name edge: `encode("CLOCK$")` — `$` is non-pass-through so it encodes anyway; `isReservedDeviceName` finds stem `CLOCK$` in the set and forces first char `%43`. Fine. But note `isReservedDeviceName` splits on the *first* `.`: `"con.tar.gz"` → stem `"con"` → neutralized. Correct per the documented "with or without extension" intent.

Verify-access `classify` ordering: `AUTHORIZATION_FAILURE`/403 checked before `SHARE_NOT_FOUND`/404 — a 403 on a nonexistent share reads as "no permission" rather than "wrong target". Minor diagnostic-only ambiguity, not a correctness bug (message quality only).

Newly-reachable failures (oversized/hostile name → `IllegalArgumentException` → wrapped as `AttachmentStorageException`) are pinned by `oversizedNamesSurfaceThroughTheCheckedExceptionNotARawEscape` for both `store` and `exists`. The receiving handler is the gated ingest endpoint (#51), same as other adapters — stated in PR, consistent.

**Security:**
- No secrets in the diff. Auth is injected `ShareClient`; connection string comes from env only in the live test (`System.getenv`), never logged.
- Traversal defense is the core feature and is tested against a strict double that rejects raw `..`/`\`, so the encoder is proven to neutralize rather than the fake being lenient. Good.
- `AttachmentStorageException` messages include `token`/`path` (e.g. the oversized-name message echoes the full `token`). Remote-controlled filenames land in exception messages → if those propagate to logs, a hostile filename is logged verbatim. Not an injection risk (no shell/SQL), but worth noting as a log-hygiene follow-up. Advisory only.

**Regression risk:**
Purely additive; no existing function/endpoint modified. No consumers exist yet — factory registration is explicitly deferred to the wiring PR (scope parity with #56). Nothing downstream can break because nothing wires this adapter in yet. The dependencyManagement BOM import is scoped to the `attachment-receiver` module pom, so it can't perturb other modules' dependency resolution.

Contract-test inheritance: `AzureFilesAttachmentStorageLiveTest extends AttachmentStorageContractTest` — I have not seen that base class in this diff, so I can't confirm the 7 inherited SPI cases actually exercise this adapter, but the PR claims they run (gated, 8 skipped offline). That's the live-run precondition, already flagged.

**Recommendation:** Merge is fine once the stated live-share acceptance runs green — the offline suite is thorough and the design is sound; the only open items (403-before-404 diagnostic ordering, filename-in-exception-message log hygiene) are post-merge follow-ups, not blockers.

---

shawn-tsanet:
Wait for Credentials and initial authentication

shawn-tsanet:
Dispositions for the auto-QA CONCERNS, item by item:

1. **Live-share acceptance unrun (the CONCERNS driver):** agreed and already the PR's stated merge gate — on hold until a TSANet-owned Azure subscription exists (decision 2026-08-12, see the updated precondition section). Tracked on the epic (#46 status comment).
2. **`classify` checks 403 before 404, so a 403 on a nonexistent share reads as no-permission rather than wrong-target:** acknowledged, message-quality only, and arguably accurate — without authorization the caller cannot distinguish a hidden share from a missing one, and Azure answers 403 before revealing existence. Left as-is; revisit if live-run experience shows the diagnostic misleads.
3. **Remote-controlled filename echoed verbatim in exception messages (log hygiene):** real, advisory, and it belongs with the module's hardening pass — added to #52's scope so it lands where filename-handling policy lands, rather than as a one-off here.
4. **Contract-test base class not visible in this diff:** it lives on the base branch (`attachment-receiver/src/test/java/com/tsanet/receiver/storage/AttachmentStorageContractTest.java`, PR #54) — this is a stacked PR, so the inherited 7 SPI cases are exactly the ones the live run executes.

No code changes pushed in response; nothing in the review asked for one.

shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
The diff pushed since the last review has nothing to do with the PR title. This PR is "Azure Files storage adapter" (issue #49, scope-parity with #56, adapter-only). The new commits add **tenant configuration + go-live verification** infrastructure: `EncryptedFileTenantConfigStore`, `TenantConfig`, `TenantConfigStore`, `TenantConfigException`, `CrmProbe`, `GoLiveVerifier`, `StorageFactory`, `VerificationReport` and their tests. None of these files is under `storage.azure`; none is an Azure adapter.

Hunks that do not trace to the PR title/body:
- `config/EncryptedFileTenantConfigStore.java` — tenant config store, not an Azure adapter.
- `config/TenantConfig.java`, `config/TenantConfigStore.java`, `config/TenantConfigException.java` — same.
- `verify/CrmProbe.java`, `verify/GoLiveVerifier.java`, `verify/StorageFactory.java`, `verify/VerificationReport.java` — go-live verification seam.
- All four new test files.

That is the entire delta. Either this is the wrong branch/PR for these commits, or the PR body needs to be rewritten to cover config+verification work — the current body describes none of it. This is the finding. I'm not blocking on it because the code itself is additive, self-contained, and correct on its own terms, but Shawn should confirm this is intended scope before merge and not a mis-stacked branch.

**Correctness:**
Traced `save(acme())` then `byPathSegment("acme")`:
- `serialize` → Properties with `pathSegment`/`pushPassword`/`storageBackend`/`storage.*`/`crm.*`. Encrypt with random 12-byte IV, AAD = `"acme"`. Body = IV ‖ ciphertext, written to temp, atomic-moved. Read: slice IV, decrypt with AAD `"acme"`, deserialize, cross-check `pathSegment`. Round-trips. ✅
- Swap test: `globex.tenant` copied over `acme.tenant`, read as `"acme"` → AAD mismatch → GCM tag fails → `TenantConfigException("integrity")`. Correct, and the test pins the mechanism (AAD, not the post-decrypt cross-check). ✅
- `StorageFactory` here is `verify.StorageFactory` (a new interface). Note the PR body for #57 says factory registration "arrives with the wiring work, not here" — yet a `StorageFactory` interface lands here. Two different types (`verify.StorageFactory` vs. the future `StorageFactory` id→adapter registry), so no collision, but worth a glance that the wiring PR doesn't expect to own this name.

One real edge: `GoLiveVerifier.storageCheck` catches `Exception`, but `AttachmentStorage.verifyAccess()` / `storageFactory.create()` could throw an unchecked `Error` (e.g. `NoClassDefFoundError` from a missing adapter dep) which escapes the per-check isolation the javadoc promises ("an unexpected RuntimeException ... never as an aborted report"). RuntimeException is caught; the javadoc claim is accurate as written. Not a bug — the doc says RuntimeException, and that's what's caught.

`deserialize`: `props.getProperty("pathSegment")` could be null if the plaintext is valid Properties but missing keys — `TenantConfig`'s compact constructor `Objects.requireNonNull` throws, caught and wrapped as `TenantConfigException`. Handled. ✅

**Security:**
- No secrets in the diff. Test credentials are literals in test files only (`push-secret-042`, etc.), fine.
- AES-256-GCM with per-write random IV, AAD-bound to tenant segment, atomic-move write, loud fail on wrong key / truncation / tamper. Sound design.
- `TenantConfig.toString()` redacts password and prints only map *keys* — good, and pinned by `toStringRedactsEveryCredential`.
- Read-path rejects non-token segments (`../escape`, null) before touching the filesystem, pinned by `readPathRejectsNonTokenSegments`. Traversal-inert by allowlist. ✅

**Regression risk:**
Purely additive new packages `config` and `verify`; no existing symbol modified. No consumers yet (nothing constructs `GoLiveVerifier` or the store). Nothing downstream can break. The only concern is organizational (see Minimal-Diff), not runtime.

**Carried forward:**
- Live-share acceptance unrun — FOLLOW-UP (stated merge gate, on hold for TSANet Azure subscription)
- 403-before-404 classify ordering — FOLLOW-UP (dispositioned: message-quality, left as-is)
- Filename echoed in exception messages — FOLLOW-UP (moved to #52 hardening scope)
- Contract-test base class not visible — RESOLVED (base branch #54, receipt cited in discussion)

**Recommendation:** Merge is fine on the code's own terms, but confirm this config+verification work belongs in the Azure-adapter PR (or update the PR body) before merge — the entire delta is untraceable to the stated scope.

---

shawn-tsanet:
Post-#56 conflict repair pushed: the pom resolves to the union (both storage SDK blocks plus the httpclient5 override #56's live run introduced). Verification on the merged branch: **81/81**, with **both** live acceptance classes green in the same run (S3 8/8, Azure Files 8/8) — the complete module proven against both real backends simultaneously. Mergeable again.

Issue #58, opened 2026-08-12
Adopt the demo-ui web module from connect-sdk-demo


shawn-tsanet:
Proposal: move the \`demo-ui\` module (Spring Boot REST layer + tabbed SPA over the SDK facades) from \`shawn-tsanet/connect-sdk-demo\` into this repo as a reactor module.

**Why now.** The demo repo exists to layer a UI over verbatim copies of this repo's modules. Since the oauth merge (\`tsanetgit/Connect_SDK#44\`) and the test-flag work (\`tsanetgit/Connect_SDK#37\`, \`tsanetgit/Connect_SDK#38\`), main carries everything the UI needs: OAuth client-credentials, per-call \`testSubmission\`, surfaced \`testCase\`. The demo tracks main byte-identically (\`shawn-tsanet/connect-sdk-demo#9\`), so the lift has no drift to reconcile: 16 Java classes plus SPA assets, exercising partner discovery, dynamic process forms, the full case lifecycle, attachments and webhooks against BETA/DEV.

**Scope of the move:**
- Copy \`demo-ui/\` as-is and add it to the root reactor. The publish workflow is unaffected (it deploys \`-pl .,connect-library\` by explicit selection).
- Fix the two \`createRequest\` call sites that use the deprecated no-flag overload (\`shawn-tsanet/connect-sdk-demo#8\`): the canonical repo should not ship a module pinned to an API scheduled for removal.
- Add thin Spring slice tests for the module's core behaviors (it currently has none): credential-gate semantics, session rebuild on credential change, explicit test-flag on create.

**Related:** \`tsanetgit/Connect_SDK#45\` (session-supplier proposal): demo-ui carries the consumer-side session-per-credential-fingerprint rebuild that #45 would make native. The module is a ready-made consumer/testbed for that design, and adopting it here puts that pattern where the discussion is. Also relevant to \`tsanetgit/Connect_SDK#26\` (SDK adopter skill): the UI doubles as the reference consumer.

**Out of scope here:** renaming/de-branding the module and the fate of the \`connect-sdk-demo\` repo after adoption (Shawn's call, tracked separately).

🤖 Generated with [Claude Code](https://claude.com/claude-code)

shawn-tsanet:
Closing with receipts; every item in scope is done.

- **Module copied + reactor entry**: \`tsanetgit/Connect_SDK#59\` (squash-merged 2026-08-12, main \`875b4ee\`), post-merge verify green.
[...]

Issue #45, opened 2026-08-11
Own the runtime-credential session rebuild in connect-library (session supplier)


shawn-tsanet:
## Context

`tsanetgit/Connect_SDK#27` shipped in `tsanetgit/Connect_SDK#44` and is released in `v0.1.0`. One design question from the port report on that issue (2026-08-03 comment) is still open: auth mode and credentials are both fixed at session construction, so any consumer that learns credentials at runtime has to rebuild sessions itself, and the API surface does not make that rule visible (`login()` and `authenticate()` both exist on the facade, but each throws `IllegalStateException` depending on a construction-time choice).

The consumer-side workaround exists and works: the demo UI holds no session until credentials exist, caches one against a SHA-256 fingerprint of them, and rebuilds when the fingerprint changes, about 90 lines. But that pattern will otherwise be reimplemented by every consumer that embeds the library behind a UI or a registry, and the part that is easy to get wrong independently is that the auth *mode* is part of the session's identity.

## Proposal

Add a small session supplier to connect-library, on the order of a hundred lines plus tests, that owns the mapping from (environment, credentials) to session. A consumer hands it an `AccountAuthConfig` whenever credentials arrive or change; it returns the current session while the credential fingerprint matches, and atomically discards and rebuilds when it does not. A credential or mode change is a new session with clean state, never a mutation of a live one.

Sessions themselves stay exactly as designed: immutable, mode fixed at construction, sealed `AccountAuthConfig` untouched. This is deliberately not `session.auth().configure(...)`; I think mutable auth is the wrong direction and want to rule it out on purpose.

## Why this shape

- **Reliability.** Mutable auth would turn the auth mode into shared state across request threads, the reconciliation-style pollers consumers run, and `OAuth401RetryInterceptor`, on top of a refresh path that currently has no locking. Immutability is what keeps the token seam tractable. The supplier preserves that invariant and puts the one genuinely tricky piece, the atomic swap, in one tested place.
- **Maintainability.** The pattern already exists consumer-side in the demo UI, the Gateway engine's per-member session registry is adjacent in shape, and every future adapter will need it again. One library implementation means one test suite and one place to fix, instead of divergent copies with divergent bugs.
[...]

PR #59, opened 2026-08-12
Adopt the demo-ui web module from connect-sdk-demo, with slice tests

shawn-tsanet:
Closes \`tsanetgit/Connect_SDK#58\`. Two commits: the verbatim module adoption, then 15 offline slice tests the module never had.

## What this is

\`demo-ui\` (Spring Boot REST layer + tabbed SPA over the SDK facades: partner discovery, dynamic process forms, full case lifecycle, attachments, webhooks, BETA/DEV switchable) moves here from \`shawn-tsanet/connect-sdk-demo\`, which has tracked this repo's modules byte-identically since its #9. The copy is from \`connect-sdk-demo@243fc3f\`, which already carries the \`shawn-tsanet/connect-sdk-demo#8\` fix: both \`createRequest\` call sites pass \`testSubmission\` explicitly, so the module arrives with zero deprecated-overload usage (receipt: \`grep createRequest demo-ui/src/main\` shows only the flag-carrying overloads; the deprecated forms are \`forRemoval\` and this module no longer blocks that removal).

## Receipts

- **Verbatim claim**: \`diff -r <demo>/demo-ui demo-ui --exclude=target -q\` against the demo clone at \`243fc3f\` returns empty (run pre-commit this session). The only non-copy hunk is the one-line reactor entry in the root pom.
- **Build**: full reactor \`mvn install\`, JDK 21, sibling \`Connect-API-Code\`@\`beta\`: BUILD SUCCESS, 137 tests, 0 failures (connect-library 86, integration-demo 6, integration-app 30, demo-ui 15).
- **Parent linkage is reactor-resolved, not warm-repo luck**: \`demo-ui/pom.xml\` declares parent \`com.tsanet:tsanet-client-parent:0.0.1-SNAPSHOT\` with default relativePath; the GAV matches this repo's root pom exactly and the build log has zero \`relativePath\` warnings.
- **Publish unaffected**: \`.github/workflows/publish.yml:105\` deploys \`-pl .,connect-library\` by explicit selection; adding a reactor module changes nothing about what is published.
- **No new dependencies**: the module pom already declared \`spring-boot-starter-test\`; all versions are Boot-BOM-managed via the parent.

## The tests (commit 2)

Four classes, 15 tests, all offline (sessions build over a temp SQLite dir and never authenticate):

- \`EnvironmentServiceTest\`: session reuse while credentials are unchanged (\`assertSame\`), rebuild on credential change and after \`invalidate()\`, per-environment credential/session isolation, active-environment persistence across service instances, fallback to default when the persisted marker names a removed environment.
- \`SessionGuardTest\`: 428 \`PRECONDITION_REQUIRED\` when no credentials are stored, and when stored credentials cannot form an auth config (OAuth creds, no Entra tenant). The guard itself maps that IllegalArgumentException to 428 (\`SessionGuard.java:54\`), so the assertion targets the property, not the advice's default.
- \`SettingsControllerTest\` (standalone MockMvc, real store over a temp dir): rejected credentials are never persisted, asserted against the store on disk rather than the response body; field validation fires before the store is touched; a valid save is observable in the store; unknown environment 400s.
- \`CollaborationRequestsControllerTest\`: both create modes verified to pass \`testSubmission=true\` to the flag-carrying overloads (the regression pin for \`shawn-tsanet/connect-sdk-demo#8\`); absent \`customFieldValues\` arrives as an empty map; a create with neither \`formTemplate\` nor \`receiverCompanyId\` 400s without touching the facade.

**Probed red before trusting green**: the explicit-flag verify fails when expected as \`eq(false)\` ("Argument(s) are different"), and the guard status assertion fails when expected as 400 ("expected: 400 BAD_REQUEST / but was: 428 PRECONDITION_REQUIRED"). Both restored and re-run green.

## Pre-PR gate dispositions (CONCERNS, all answered)

1. *"Verbatim copy unverifiable from the diff"*: receipt above (empty recursive diff at \`243fc3f\`).
2. *"Publish workflow \`-pl\` claim unverified"*: receipt above (workflow line 105 quoted).
3. *"Hosted deploys must set \`TSANET_DEMO_AUTH_PASSWORD\` or the app is open"*: correct, pre-existing, and documented in the demo repo's Azure deployment doc; carried to Shawn as an ops note.
4. *Lateral notes on copied code* (a malformed \`formTemplate\` surfaces as 502 via \`ApiErrorHandler.handleOther\`; that handler maps every uncaught exception to 502): pre-existing module behavior, deliberately not changed in an adoption PR. Real follow-up candidates once the module is upstream.

## Deliberate exclusions

- The demo repo's docs (USER_GUIDE, RUNBOOK, Azure deployment) stay behind; they document the hosted demo, not this repo.
- The module keeps its \`demo-ui\` name and TSANet demo branding pending Shawn's product call (tracked in \`tsanetgit/Connect_SDK#58\`).
- No CI verify workflow added: this repo has none for Java (only the security audit and release publish), so the new tests run locally and in any future verify workflow. Now that \`CONNECT_API_CODE_TOKEN\` exists (the v0.1.0 publish run used it), a PR-triggered build+test workflow with a sibling-spec checkout is feasible; recommended follow-up, kept out of this PR's scope.

## Related

\`tsanetgit/Connect_SDK#45\`: \`EnvironmentService\` is the living example of the consumer-side session-per-credential-fingerprint rebuild that #45 proposes making native; the new \`EnvironmentServiceTest\` pins that behavior, which becomes the compatibility contract for any #45 implementation.


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
This is a module adoption PR — 26 new files (all under `demo-ui/`) plus one root pom change. Walking the non-trivial hunks:

- `pom.xml:25` — adds `<module>demo-ui</module>` to the reactor. Traces to the PR's central claim (adopt the module). This is the single non-copy hunk the PR flags, and it matches.
- All `demo-ui/src/main/**` files (17 Java + yml + html/js/css/png) — the verbatim module body. Traces to "verbatim module adoption" (commit 1).
- All `demo-ui/src/test/**` (4 classes, 15 tests) — traces to "then 15 offline slice tests the module never had" (commit 2).

Every hunk traces. I cannot independently verify the byte-identical claim from the diff alone (that's what `diff -r ... 243fc3f` in the receipts asserts, and I have not run it this session), but nothing in the copied content looks like incidental tidying grafted onto a copy — it's coherent as a whole module. No scope creep found.

**Correctness:**
Traced the pinned regression path — `CollaborationRequestsController.createRequest` with a form body. `body.formTemplate()` non-null → calls the 6-arg overload with `body.customFieldValues() != null ? ... : Map.of()` and literal `true`. `CollaborationRequestsControllerTest.itCreatesViaTheFormTemplateAsAnExplicitTestSubmission` sends `{"formTemplate":{},...}` with no `customFieldValues` and verifies `eq(Map.of())` and `eq(true)`. The verify is on the real facade mock, the assertion targets the actual argument, and the PR states it was probed red as `eq(false)`. This is a real test, not a trivially-passing one.

Traced `SessionGuard.session()` with no stored credentials: `credentialsFor(key).load()` returns empty → `orElseThrow` → 428 `PRECONDITION_REQUIRED`. `SessionGuardTest.itReturns428WhenNoCredentialsAreStored` confirms both status and reason. Good.

One real edge, pre-existing and correctly disclosed: `ApiErrorHandler.handleOther` (line 34) maps *every* uncaught `Exception` to 502 BAD_GATEWAY. A malformed `formTemplate` that the SDK rejects internally surfaces as 502, not 4xx. The PR names this in disposition #4 and marks it a follow-up, not a fix — appropriate for an adoption PR.

**Security:**
- No secrets in the diff. `application.yml` uses env-var placeholders (`${TSANET_DEMO_ENTRA_TENANT:}` etc.) with empty defaults; the comment explicitly notes real values come from the environment, not the repo. Good.
- `CredentialsStore` writes plaintext credentials to a properties file with `rw-------` posix perms and never logs them. `EnvironmentService.fingerprint` hashes the secret for the cache key rather than holding a second plaintext copy. Reasonable for a local demo.
- The Basic-auth gate (`SecurityConfig`) is disabled when no password is set — the app is open by default. This is disclosed (disposition #3) as pre-existing and carried to Shawn as an ops note. Not a new exposure introduced by this PR.
- XSS surface: `app.js` renders API data via `esc()` for text and `safeBodyNodes()` for HTML note bodies, which rebuilds an allowlist of attribute-free tags and unwraps the rest — the copied code treats partner-originated markup as hostile. This is defensively written, not a new hole.
- `AttachmentsController` writes uploads to `Files.createTempDirectory(...).resolve(original)` where `original` is the client filename. A `../` filename can't escape because `resolve` on a random temp dir with a traversal segment still lands relative to that fresh dir, and cleanup deletes the temp dir. Low risk; pre-existing copied behavior.

**Regression risk:**
- The only shared surface touched is the root `pom.xml` reactor list. The PR asserts `publish.yml:105` uses `-pl .,connect-library` explicit selection, so adding a reactor module does not change what is published. I have not opened `publish.yml` this session to confirm that line, but the claim is specific and checkable, and the mechanism (explicit `-pl`) is sound — a new module is not auto-published under explicit selection.
- No existing file's behavior changes; every other file is new. No consumers of a modified function exist.
- Parent GAV `com.tsanet:tsanet-client-parent:0.0.1-SNAPSHOT` must match the root pom for the reactor to resolve; the PR asserts a clean build with zero relativePath warnings.

**Operational:**
- New module builds under the existing reactor; no new top-level dependency (module pom pulls `spring-boot-starter-web/security/test`, all Boot-BOM-managed).
- New runtime env vars (`TSANET_DEMO_*`) are demo-ui-scoped and documented inline in `application.yml`; this repo has no `.env.example` for the Java side, so nothing to update there.
- No `db/schema.sql` or sidecar impact — this is a separate Spring Boot app, not the ops-sidecar.
- No LaunchAgent restart implication for tsanet-ops infra.
- No CI verify workflow added; the PR notes this repo has none for Java and recommends it as a follow-up. Reasonable to keep out of scope.

**Recommendation:** Merge — every hunk traces, the disclosed 502-mapping and open-by-default gate are pre-existing and correctly deferred, and the slice tests genuinely exercise the pinned behavior.

---

shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
The new commit adds an entire `attachment-receiver` module (9 main + 5 test files, +1 reactor line at `pom.xml:25`). None of this traces to the PR title ("Adopt the demo-ui web module") or body — the PR is about `demo-ui`. This is a large, coherent, well-built module, but it is a *different* deliverable smuggled into a demo-ui adoption PR. That is scope creep by the letter of the rule, even though the code itself is clean. Flagging it as the primary finding: either the PR title/body should be updated to cover the receiver module, or the receiver should land in its own PR.

**Correctness:**
Traced the AAD swap-defense path in `EncryptedFileTenantConfigStore`. `save(acme)` binds `"acme"` as AAD; copying `globex.tenant` over `acme.tenant` then `byPathSegment("acme")` decrypts with AAD `"acme"` against ciphertext bound to `"globex"` → GCM tag mismatch → `GeneralSecurityException` → `TenantConfigException` with "integrity" in the message. `swappedTenantFilesDoNotDecrypt` asserts exactly that. Correct.

Traced `read()` truncation guard: `body.length <= IV_BYTES` (12) throws "truncated" before slicing. A 12-byte file (IV only, no ciphertext/tag) is caught. Good — `<=` is right since a valid GCM body needs IV + 16-byte tag minimum.

Path-segment defense: `byPathSegment("../escape")` fails the `PATH_SEGMENT` regex → `IllegalArgumentException` before any `resolve`. `readPathRejectsNonTokenSegments` confirms. The `all()` path derives segments from on-disk filenames rather than request input, so no traversal there.

One edge: `GoLiveVerifier.crmCheck` catches `Exception` from `crmProbe.dryRun`, but `CrmProbe.dryRun` is declared `throws Exception` and `failureMessage` handles null messages. No gap. Tests cover storage-fail + crm-fail independently.

**Security:**
No secrets in the diff — test credentials are literals like `"push-secret-042"`, clearly fixtures. AES-256-GCM at rest, key required to decode to exactly 32 bytes at startup, credentials redacted in `toString()` and never logged. `ciphertextContainsNoCredentialBytes` asserts the on-disk bytes contain no plaintext credential. This module is security-conscious. No new exposure.

**Regression risk:**
The only shared surface is `pom.xml:25` adding `attachment-receiver` to the reactor. Same mechanism as the demo-ui line already reviewed: publish is `-pl .,connect-library` by explicit selection (per prior receipt), so a new reactor module changes nothing published. The module has no dependents yet (all files new, SPI only), so no consumer breakage. Build impact: adds a module to the full reactor build — the PR body's build receipt (137 tests) predates this commit and does not mention `attachment-receiver`, so the combined build+test result for this module is unverified in the PR.

**Carried forward:**
- 502-mapping on malformed formTemplate (demo-ui) — FOLLOW-UP (pre-existing, deferred).
- open-by-default auth gate (demo-ui) — FOLLOW-UP (pre-existing ops note to Shawn).
- publish.yml:105 -pl claim unverified — FOLLOW-UP (mechanism sound, not re-checked this session).

**Recommendation:** Merge is fine on code quality, but resolve the scope mismatch first — update the PR title/body to declare the attachment-receiver module, or split it into its own PR; and post a build+test receipt covering `attachment-receiver`.

---

PR #60, opened 2026-08-12
Add a full-reactor verify workflow for PRs and main

shawn-tsanet:
Adds the Java build/test CI this repo never had: until now the only PR check was the security audit, so compile and test breakage could reach main unseen and build receipts lived on whichever machine last ran \`mvn install\` (the \`tsanetgit/Connect_SDK#59\` review called this out).

## What it does

- \`mvn -B verify\` on the **whole reactor** (deliberately no \`-pl\` selection: publishing is selective, verification is not — every module, including never-published ones like the app, demo, and the incoming demo-ui, must build and pass tests).
- Triggers on \`pull_request\` and \`push\` to main. Concurrency cancels superseded PR runs; pushes to main get a per-commit group so a newer push can never cancel a pending older one and leave a main commit unverified.
- Reuses the publish workflow's proven pieces verbatim: the fail-loud \`CONNECT_API_CODE_TOKEN\` gate (secret name identical to \`publish.yml:36\`, proven working by the v0.1.0 publish run), the side-by-side \`Connect_SDK\` / \`Connect-API-Code\` checkout layout the root pom's \`connect.openapi.spec\` property expects, the spec pinned to \`beta\` (the README's build notes: the branch matters), and temurin 21 with the Maven cache.

## Known behaviors, by design

- **Fork PRs fail at the token gate** with the remediation in the error text (fork runs receive no repository secrets). Org members push branches to this repo, so this affects no current contributor. No Dependabot on this repo (no config, zero historical PRs), so the Dependabot-secrets variant of this trap does not apply.
- \`pull_request\`, not \`pull_request_target\`: no elevated secret exposure to PR code.

## Validation

This workflow runs on this very PR — a new workflow triggers on the PR that introduces it, which makes this PR the clean-machine end-to-end test. Green checks below are the receipt.

## Follow-up (needs admin, after merge)

Make \`verify\` a **required status check** in branch protection — the workflow makes breakage visible, only branch protection makes it blocking. Known consequence once required: fork PRs cannot merge without a maintainer pushing the branch in-repo, consistent with the gate text.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -1,0 +1,73 @@
+name: verify
+
+# Full-reactor build and test on every PR and on main. Until this workflow
+# existed the repo's only PR check was the security audit, so Java compile and
+# test breakage could reach main unseen; the receipts lived on whichever
+# machine last ran `mvn install`. This is the clean-machine counterpart.
+#
+# Runs the WHOLE reactor deliberately (no -pl selection): the publish workflow
+# selects modules because publishing is selective, but verification is not —
+# every module, including ones never published, must build and pass its tests.
+on:
+  pull_request:
+    branches: [main]
+  push:
+    branches: [main]
+
+# A superseded push to the same PR doesn't need its old run to finish. Pushes
+# to main get a per-commit group instead: a shared group would let a newer
+# push cancel a PENDING older one, leaving that commit with no verify result.
+concurrency:
+  group: verify-${{ github.event_name == 'push' && github.sha || github.ref }}
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
+
+permissions:
+  contents: read
+
+jobs:
+  verify:
+    runs-on: ubuntu-latest
+    timeout-minutes: 30
+    steps:
+      # connect-library's codegen reads the Connect OpenAPI spec from a sibling
+      # checkout of Connect-API-Code (the connect.openapi.spec property), which
+      # is a PRIVATE repository, so this checkout needs a token with read access
+      # to it. Same fail-loudly rule as the publish workflow: a skipped build
+      # that reads as success is worse than a red one. Note this means PRs from
+      # FORKS fail here by design — fork runs do not receive repository
+      # secrets; org members should push branches to this repository instead.
+      - name: Require the spec-repo token
+        env:
+          HAVE_TOKEN: ${{ secrets.CONNECT_API_CODE_TOKEN != '' }}
+        run: |
+          set -euo pipefail
+          if [ "$HAVE_TOKEN" != "true" ]; then
+            echo "::error::The CONNECT_API_CODE_TOKEN secret is not available. The build generates connect-library's client from the Connect-API-Code spec (a private repo), so a read token for it is required. Fork PRs do not receive secrets; push the branch to this repository instead."
+            exit 1
+          fi
+
+      - uses: actions/checkout@v4
+        with:
+          path: Connect_SDK
+          persist-credentials: false
+
+      # The spec BRANCH matters: the generated symbols depend on which
+      # specification the sibling provides (see the README's build notes), and
+      # this repo builds against beta.
+      - uses: actions/checkout@v4
+        with:
+          repository: tsanetgit/Connect-API-Code
+          ref: beta
+          token: ${{ secrets.CONNECT_API_CODE_TOKEN }}
+          path: Connect-API-Code
+          persist-credentials: false
+
+      - uses: actions/setup-java@v4
+        with:
+          distribution: temurin
+          java-version: "21"
+          cache: maven
+
+      - name: Build and test the full reactor
+        working-directory: Connect_SDK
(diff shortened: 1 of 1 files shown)


shawn-tsanet (commented):
I'll review this PR. Let me note upfront: this is a CI workflow file, so my checks focus on the workflow contract and its correctness against the referenced `publish.yml`. I have not independently verified the claims about `publish.yml:36`, the root pom `connect.openapi.spec` property, or the README build notes against the repo this session — I flag those where they matter.

**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

Single new file, 73 lines, one logical unit. Walking it:

- `name` + top comment (L1–14): traces to "Add a full-reactor verify workflow" — the comment explains the *why* the PR body states. Traces.
- `on: pull_request/push` (L15–19): traces to "Triggers on `pull_request` and `push` to main." Traces.
- `concurrency` block (L21–25): traces to "Concurrency cancels superseded PR runs; pushes to main get a per-commit group." Traces.
- `permissions: contents: read` (L27–28): least-privilege; supports the security posture the body claims. Traces (implied by "no elevated secret exposure").
- Token gate step (L36–48): traces to "the fail-loud `CONNECT_API_CODE_TOKEN` gate." Traces.
- Two `checkout` steps (L50–63): traces to "the side-by-side checkout layout... the spec pinned to `beta`." Traces.
- `setup-java` (L65–70): traces to "temurin 21 with the Maven cache." Traces.
- `mvn verify` step (L71–73): traces to "`mvn -B verify` on the whole reactor." Traces.

No untraced hunks. Nothing incidental.

**Correctness:**

Traced the concurrency expression through both event types:
- PR event: `group: verify-refs/pull/60/merge`, `cancel-in-progress: true` — superseded PR pushes cancel. Correct.
- Push to main (say sha `abc123`): `group: verify-abc123`, `cancel-in-progress: false` — per-commit group, no cancellation. Correct, and the per-commit group is the right call: with a shared `refs/heads/main` group and `cancel-in-progress: false` you'd still be fine, but the per-sha group is strictly safer and matches the stated intent. Good.

Token gate: `HAVE_TOKEN` is `secrets.CONNECT_API_CODE_TOKEN != ''` evaluated to the string `"true"`/`"false"`, compared as a string — correct pattern. `set -euo pipefail` present. Fork PRs hit this and exit 1 with remediation text, matching "Known behaviors, by design." Correct.

One real behavior to confirm (not a bug, a question): the token gate runs *before* `actions/checkout`, so it runs in a bare workspace — fine, it uses no repo files. Order is correct.

`persist-credentials: false` on both checkouts is good hygiene. The second checkout uses the token explicitly for the private repo, which is necessary. Correct.

The build step uses `working-directory: Connect_SDK` and the sibling `Connect-API-Code` is checked out as a peer directory — this depends on the root pom's `connect.openapi.spec` resolving to `../Connect-API-Code` (or equivalent). I have not verified the pom property this session; the PR asserts publish.yml uses the identical layout and it's proven by the v0.1.0 run. If that assertion holds, this is correct. Worth a one-line confirmation but not blocking — the workflow itself will fail loudly if the path is wrong, which is the desired outcome.

**Security:**

- No secrets in the diff — only secret *references* (`secrets.CONNECT_API_CODE_TOKEN`). Correct.
- `pull_request` (not `pull_request_target`) means PR code runs without repo secrets in its context except where explicitly passed — the token is only passed to the trusted `actions/checkout@v4` action for the private-repo clone, not exposed to arbitrary build steps as an env var. Good.
- The gate `echo` prints only the *absence* of the token, never its value. No secret leakage in logs.
- `permissions: contents: read` — least privilege, no write scope. Correct.
- Pinned action major versions (`@v4`) rather than full SHAs — acceptable for this repo's posture; SHA-pinning would be stricter but is a follow-up, not a blocker.

No concerns.

**Regression risk:**

New file, no modification to existing workflows or code. Nothing existing calls this. The only interaction with the existing setup is the shared secret `CONNECT_API_CODE_TOKEN` (read-only reference — adding a second consumer of a secret doesn't affect the first) and the `verify` workflow name (I have not checked for a name collision with an existing workflow this session; if `publish.yml` or the security-audit workflow already uses `name: verify`, GitHub tolerates duplicate names but it's confusing — worth a grep, low stakes).

The `beta` spec ref is a moving branch, not a tag — a future breaking change to `Connect-API-Code@beta` would turn this red independent of the PR under test. That's inherent to the design (the PR states the branch is deliberate) and matches publish.yml, so it's not a regression this PR introduces. Noting it as a known property.

**Recommendation:** Merge — this is a clean, well-scoped, fail-loud CI addition; optionally confirm the sibling-checkout path against the root pom and grep for a `verify` name collision before enabling it as a required check, but neither holds the merge.

One operational note for Shawn: this is marked **Draft**, and the PR's own "Validation" section relies on this workflow's green run on this PR as the receipt. Confirm the check actually ran green here before merging, and remember the required-status-check step is a separate post-merge admin action.

---

PR #62, opened 2026-08-14
Remove the CRaSH SSH remote shell from TSANet-integration-app

### LINKED ISSUE #61: CRaSH SSH management shell enabled by default with public crash/crash credentials (RCE) in TSANet-integration-app
Source: Security review of `Connect_SDK @ 82018c2` (main), 2026-08-14. Taxonomy category 3 (authn/authz boundaries) with a category 1 supply-chain angle (unmaintained Groovy-bearing dependency). Pre-existing condition, surfaced during the 2026-08-13 account-exposure audit.

## Problem

`TSANet-integration-app` ships, as its committed default, a CRaSH remote SSH management shell bound to **all network interfaces** with **public, hard-coded credentials**. This is the documented way to drive the module.

The facts, with receipts:

- **Enabled by default.** `TSANet-integration-app/src/main/resources/application.yml` ships `crash.enabled: true` (plus `ssh-port: 2000`, `auth-username: crash`, `auth-password: crash`), and imports `tsanet-demo-crash-ssh.properties`, which also sets `crash.auth-username=crash` / `crash.auth-password=crash`. `CrashConfiguration` is `@ConditionalOnProperty(name = "crash.enabled", havingValue = "true")`, so the committed `application.yml` is what turns it on.
- **Credentials are public.** `crash`/`crash` is in the repo, in a public GitHub repository.
- **Binds all interfaces, not loopback.** CRaSH `crash.connectors.ssh:1.3.2`'s `SSHLifeCycle` calls only `setUpDefaultServer()` + `setPort(...)` — there is no `setHost` path. Apache MINA SSHD with no host set binds the wildcard address (`0.0.0.0`). Loopback-only binding is not achievable in this CRaSH version.
- **Blast radius is remote code execution.** `crash.shell:1.3.2` ships the base command set, including `shell.groovy` (an interactive Groovy REPL) and `java.groovy`, with `groovy-all` on the classpath transitively. Any client that authenticates gets arbitrary code execution inside the JVM.
- **Plus full Connect API takeover.** The custom `tsa` command (`TsaCrashCommands`) additionally exposes `token` (prints the current bearer token), plus create/approve/reject/close collaboration requests, notes, attachments, and webhook CRUD, and reads of the local SQLite store.
- **Documented intended use.** `connect-library/README.md` presents "the console and CRASH SSH application" and gives "Over CRASH SSH (port 2000 by default)" examples throughout, so the vulnerable configuration is the documented path, not an obscure toggle.

**Scope / boundary honesty.** This is **not** present in any built or shipped artifact. The published Maven artifact (`v0.1.0`, GitHub Packages) is `connect-library` only, which carries no CRaSH dependency. The container image (`Dockerfile` → App Runner) packages `demo-ui`, which depends only on `connect-library` and has no `crash` block. CI (`verify.yml`) runs a `@SpringBootTest` in this module, but the test `application.yml` sets `crash.enabled: false` and `web-application-type: none`, so nothing binds there, and runners are ephemeral GitHub-hosted. The exposure is realized only when someone builds and runs `TSANet-integration-app` from source — which the README tells them to do.

**Severity: HIGH.** It is the committed default of a documented-runnable module, and on any cloud VM, shared host, or shared LAN it is unauthenticated-equivalent RCE via public credentials (NAT does not protect against same-network attackers). It is rated HIGH rather than "actively exposed" precisely because it is not live in a shipped artifact today. Note `crash`/`crash` in git history is not the concern — a secret was never leaked; the default *behavior* is the defect.

## Fix

Remove the CRaSH SSH feature. It is CRaSH `1.3.2` (circa 2015, unmaintained), ships a Groovy REPL by design, drags in an old `groovy-all` with its own standing CVE surface, and cannot bind loopback. Every `tsa` capability already exists over the module's plain console CLI (the README shows both forms), so deletion loses only the *remote* convenience, not functionality.

Remove:
- the three `org.crashub:crash.*:1.3.2` dependencies from `TSANet-integration-app/pom.xml`, and the `crash.jvm.opens` plumbing that only exists for them;
- `CrashConfiguration.java`, `CrashProperties.java`, `TsaCrashCommands.java`;
- `src/main/resources/crash/**` (the `tsa` command + `crash.properties`) and `src/main/resources/tsanet-demo-crash-ssh.properties`, plus the `spring.config.import` line that loads it;
- the `crash.*` block from `application.yml`;
- the `scripts/*crash*` helpers (`tsa-crash-ssh.sh`, `crash-ssh-common.sh`, `crash-ssh-run.expect`, `load-crash-password.sh`);
- the "Over CRASH SSH" examples from `connect-library/README.md` (keep the equivalent console-CLI examples).

If maintainers decide the remote shell must be kept, the fallback is fail-closed: default `crash.enabled: false`; delete the `crash`/`crash` defaults everywhere (`application.yml`, the `CrashProperties` record defaults, and the properties file); require operator-supplied credentials from the environment with no default, failing startup if enabled without them; and document that enabling exposes all interfaces (loopback binding is impossible in 1.3.2), so it may only be enabled on a trusted, firewalled host.

## Acceptance

- [ ] `TSANet-integration-app` builds and its tests pass with the CRaSH dependencies removed from `pom.xml`.
- [ ] No `crash`/`crash` (or any `crash.auth-*`) default remains anywhere in the repo; `grep -ri 'crash.auth-password' .` returns nothing.
- [ ] Running the module opens no listener on `:2000` (`ss -ltnp | grep 2000` empty after startup).
- [ ] `connect-library/README.md` no longer instructs use of a CRaSH SSH shell; the documented path is the console CLI.
- [ ] `scripts/security-audit.py` gains a check that FAILs on a shipped `crash.enabled=true` alongside a committed `crash.auth-password` (tracked separately in the security register so the mechanical floor can't miss this class again).

shawn-tsanet:
Closes #61.

## What this does

Removes the CRaSH SSH remote-shell feature from `TSANet-integration-app`. As committed, CRaSH shipped **enabled by default** (`crash.enabled: true`) with **public `crash`/`crash` credentials**, binding an SSH management shell on **all interfaces** (`0.0.0.0:2000`). CRaSH's base command set includes an interactive Groovy REPL, so any client that authenticated with the committed credentials had **arbitrary code execution** in the JVM, plus full Connect API session control via the `tsa` command (bearer-token disclosure, request approve/reject/close, webhook CRUD). See #61 for the full finding.

**Delete rather than gate:** the remote shell duplicated the module's existing console CLI, so it is removed rather than hardened. CRaSH 1.3.2 is unmaintained (2015-era), cannot bind loopback only, and drags in an old `groovy-all`.

## Changes

Removed:
- The three `org.crashub:crash.*:1.3.2` dependencies (`TSANet-integration-app/pom.xml`).
- `CrashConfiguration.java`, `CrashProperties.java`, `TsaCrashCommands.java`.
- `src/main/resources/crash/**` (the `tsa` command + `crash.properties`) and `tsanet-demo-crash-ssh.properties`.
- The `crash.*` block and the `spring.config.import` of the deleted properties file from `application.yml`.
- The `crash.enabled: false` line from the test `application.yml`.
- `scripts/crash-ssh-common.sh`, `scripts/crash-ssh-run.expect`, `scripts/load-crash-password.sh`, `scripts/tsa-crash-ssh.sh`, `scripts/simple-test.sh` (the last is entirely a CRaSH SSH test driver).
- The `crash.jvm.opens` property (root `pom.xml`) and the crash-only surefire / `spring-boot-maven-plugin` config in the app pom.
- The seven "Over CRASH SSH" example blocks from `connect-library/README.md`, and the "console and CRASH SSH application" phrasing in its intro.

`spring-boot-maven-plugin` stays declared (so `spring-boot:run` / repackage behavior is unchanged); only its crash-only `<jvmArguments>` was removed.

## Verification

- `mvn test` on the **full reactor** passes — all 6 modules green (connect-library 86, TSANet-integration-demo 6, TSANet-integration-app 30, attachment-receiver 28, demo-ui 15).
- The app's `@SpringBootTest` (`ConnectFacadeApplicationTest`) boots the full context **without** CRaSH.
- Live smoke: `mvn -pl TSANet-integration-app spring-boot:run` boots and **no listener opens on :2000** (`lsof -nP -iTCP:2000 -sTCP:LISTEN` empty); the app's web server comes up on :8090 as before.
- `git grep -n 'crash.jvm.opens'` → empty. `git grep -in crash` → only the unrelated "a crash mid-save" comment in `attachment-receiver` (see disposition below).
- Removing the `--add-opens` flags is safe: `jvmArguments` only affects `spring-boot:run`, never the packaged jar's launch, and the 30 app tests (incl. the context boot) pass without them.

## Deliberate non-changes (dispositions)

- `scripts/security-audit.py:33` — a **past-tense comment** that cites `scripts/simple-test.sh` as the historical reason `.sh` files are scanned for hardcoded credentials. It is not a functional consumer of the deleted script, the rationale still holds (`.sh` should be scanned), and the audit tooling is out of scope for a feature-deletion PR. Left as-is intentionally.
- The `crash.enabled=false` removal from the test yml and the `spring.config.import` removal from `application.yml` leave nothing dangling — the property and the file they pointed to no longer exist.

## Blast-radius notes

- **Deleted shared symbols / scripts:** `git grep` for each deleted script name (`simple-test`, `tsa-crash-ssh`, `crash-ssh-common`, `crash-ssh-run`, `load-crash-password`) returns no consumer except the one documented comment above. No `.github/workflows/` job invokes any of them.
- **Root-pom property removal:** `crash.jvm.opens` had no references outside the app pom's two now-removed uses; the full reactor builds clean without it.

## Out of scope

This is a code deletion. The `security-audit.py` enhancement from #61's acceptance (a mechanical check that FAILs on a shipped `crash.enabled=true` + committed `crash.auth-password`) is audit-tooling work tracked separately in the security register, not part of this PR.


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
- `pom.xml` (app) — description string, three `org.crashub` deps, surefire plugin block, and `<jvmArguments>` removal → all trace to "Removed the three CRaSH dependencies" and "crash-only surefire / spring-boot-maven-plugin config." The `spring-boot-maven-plugin` element is retained with empty config, matching the PR's explicit "stays declared" note. Traces.
- `CrashConfiguration.java`, `CrashProperties.java`, `TsaCrashCommands.java` deletions → trace directly to the three named file removals.
- `application.yml` (main) — removes `spring.config.import` and the `crash.*` block → traces to "the crash.* block and the spring.config.import."
- `crash/commands/tsa/tsa.java`, `crash/crash.properties`, `tsanet-demo-crash-ssh.properties` deletions → trace to "src/main/resources/crash/** and tsanet-demo-crash-ssh.properties."
- test `application.yml` — removes `crash.enabled: false` → traces to "the crash.enabled: false line from the test application.yml."
- `connect-library/README.md` — intro phrasing + seven "Over CRASH SSH" blocks → traces to the README disposition. I count seven removed `Over CRASH SSH` blocks (attachments, partners, reject, information exchange, notes, close, webhooks). Matches "seven."
- root `pom.xml` — `crash.jvm.opens` property removal → traces.
- `scripts/*` deletions (5 files) → trace to the named script removals.

No untraced hunks. Diff is pure deletion plus two string edits, both justified.

**Correctness:**
- The `@ConditionalOnProperty(name = "crash.enabled", havingValue = "true")` on the deleted `CrashConfiguration` means the whole CRaSH stack only wired when `crash.enabled=true`. Removing the config class + the yml key that set it true removes the bean entirely — no dangling bean reference, since nothing else `@Autowired`s `CrashProperties` or `SpringBootstrap` (both were CRaSH-only). Traced: on `spring-boot:run`, context loads `application.yml` with no `crash:` block → no property source imports the deleted file (the `import` was `optional:`, so even a stale reference wouldn't have failed, but it's gone anyway) → no listener on :2000. The PR's `lsof` receipt confirms the runtime outcome.
- Test yml removed `crash.enabled: false`. Since the config is `havingValue = "true"`, absence and `false` are equivalent — the bean stays off. `ConnectFacadeApplicationTest` context boot is unaffected. Correct.
- The `spring-boot-maven-plugin` retained as a bare `<plugin>` element is valid — Maven falls back to plugin-management defaults. `spring-boot:run`/repackage behavior unchanged, as claimed.
- `TsaCrashCommands` extended `BaseCommand` and dispatched to `CliCommandDispatcher`/executors that live in `com.tsanet.facade.cli`. Those executors are NOT deleted — only the CRaSH wrapper is. The console CLI path is the surviving caller. No executor loses its only consumer here.

None found.

**Security:** None. This PR removes a security exposure (public `crash`/`crash` creds on `0.0.0.0:2000` with a Groovy REPL). No secrets introduced; the deleted `tsanet-demo-crash-ssh.properties` and `simple-test.sh` carried/consumed the hardcoded `crash`/`crash` pair and are gone. No new inputs, no injection surface added.

**Regression risk:**
- Consumers of the deleted scripts: PR claims `git grep` for each script name returns only the past-tense comment in `scripts/security-audit.py:33`. I have not re-run that grep this session, so I'm relaying the author's receipt rather than confirming it. The named residual — a past-tense comment, not a functional `source`/exec of the deleted script — would not break the audit tool at runtime. Low risk; worth a spot-check that no `.github/workflows/` job invokes `simple-test.sh` (author asserts none).
- `crash.jvm.opens` root-pom property: only referenced by the two now-removed app-pom uses per the PR. If any other module referenced it, `mvn` would leave an unresolved `${crash.jvm.opens}` literal in an argLine — but the PR's full-reactor `mvn test` pass is the receipt that nothing else consumes it. Acceptable.
- `spring.config.import` removal: the import was `optional:`, so nothing downstream depended on that property source being present. No regression.
- README is documentation only — no code consumer.

**Recommendation:** Merge — this is a clean, fully-traced deletion of a real RCE-class exposure, with runtime receipts (`lsof` empty on :2000, full-reactor tests green); the only residual is a documented past-tense comment that is not a functional consumer.

---

shawn-tsanet (commented):
## QA review of "Remove the CRaSH SSH remote shell from TSANet-integration-app"

Reviewed at `857a9d3` against `origin/main` at `eca8225`. Every claim below is a check I ran this session, not a restatement of the PR body.

### Minimal-Diff Rule

Pass. All 16 files trace to the deletion:

- The three `org.crashub` deps, the crash-only surefire `<argLine>` and `spring-boot-maven-plugin` `<jvmArguments>`, and the root-pom `crash.jvm.opens` that fed both.
- `CrashConfiguration` / `CrashProperties` / `TsaCrashCommands`, `resources/crash/**`, `tsanet-demo-crash-ssh.properties`, and the `spring.config.import` that loaded it.
- The `crash.*` block (main yml) and `crash.enabled: false` (test yml), both now referring to properties that no longer bind.
- The five `scripts/*crash*` helpers plus `simple-test.sh`.
- README: the seven "Over CRASH SSH" blocks and the two wording changes.

The `<description>` rewording and the README intro rewording are in scope: both named the removed feature. The surefire plugin block is deleted whole rather than emptied, which is right since `argLine` was its only element, and nothing else in the reactor consumes `argLine` (`git grep -n "jacoco\|argLine"` on the branch returns empty, so there is no coverage agent whose `argLine` this was shadowing).

### Correctness

- **Full reactor `mvn test` reproduces green.** I re-ran it on the branch worktree under JDK 21: `BUILD SUCCESS`, all six modules, `86 / 6 / 30 / 28 / 15` tests, 0 failures 0 errors. Matches the PR's numbers exactly.
- **The `--add-opens` removal is proven, not argued.** The 30 app tests, including the `@SpringBootTest` context boot, pass with surefire no longer injecting `--add-opens java.base/java.lang=ALL-UNNAMED`. That is the check that mattered, since `argLine` (unlike `jvmArguments`) really did affect the test JVM.
- **Console-CLI parity holds for the commands the README stopped documenting.** Each deleted `tsa` verb maps to a registered console command in kebab-case (`webhookDeliveries` to `webhooks deliveries`, `storedAttachments` to `stored-attachments`, `notesAddDetailed` to `add-note --description`, and so on). The one that looked like it could be CRaSH-only, `partnersSemantic`, is covered: `PartnersCommand.java:26` advertises `--semantic`, and `PartnerSearchExecutor.java:23` names it in the empty-result hint.
- **`simple-test.sh` really is inseparable from CRaSH,** so deleting it with the feature is correct rather than a coverage decision made by accident. Every step in it goes through `crash_ssh_exec`; there is no transport left to point it at.

### Security

Strongly net positive, and larger than the PR body claims.

- **The CRaSH code is off the classpath entirely, not merely unconfigured.** `mvn -pl TSANet-integration-app dependency:tree | grep -iE 'crash|groovy|sshd|bouncy'` returns empty. That is a stronger receipt than the live "no listener on :2000" smoke, because it rules out the Groovy REPL and the MINA SSHD stack being reachable by any configuration path at all.
- **No residue.** `git grep -in crash` on the branch returns two hits, both unrelated English usage (`EncryptedFileTenantConfigStore.java:30` "a crash mid-save", `security-audit.py:269` "a crashed audit"). Plain `git grep -n 2000` returns one hit, `CaseRejectionValidation.MAX_REASON_LENGTH`. This satisfies `tsanetgit/Connect_SDK#61`'s acceptance item 2.
- **Undisclosed bonus:** the deleted `scripts/crash-ssh-common.sh` also re-enabled retired SSH crypto and disabled host verification for its own client calls (`StrictHostKeyChecking=no`, `UserKnownHostsFile=/dev/null`, `PubkeyAuthentication=no`, `HostKeyAlgorithms=+ssh-rsa`, `KexAlgorithms=+diffie-hellman-group14-sha1`, lines 8 to 14). Worth naming in the PR body, since it is a second security improvement the reader would otherwise miss.

No secrets introduced. `scripts/security-audit.py` passes on the branch: `0 fail, 2 warn, 5 pass`, and both warnings are pre-existing on `main`.

### Regressions

- **No dangling consumers.** `git grep` on the branch for each deleted script name (`simple-test`, `tsa-crash-ssh`, `crash-ssh-common`, `crash-ssh-run`, `load-crash-password`) and for `crash.jvm.opens` returns nothing except the one comment the PR already disposes of at `scripts/security-audit.py:33`. That disposition is right: it is past-tense rationale for why `.sh` is in `SOURCE_EXT`, and the rationale survives the script.
- **No CI breakage.** The three workflows on the branch (`publish.yml`, `security-audit.yml`, `verify.yml`) invoke none of the deleted scripts.
- **`resources/crash/**` is fully pruned, not partially.** `git ls-tree` over `TSANet-integration-app/src/main/resources/` leaves only `application.yml` and the `META-INF/spring` post-processor registration.
- **Docs are swept beyond the one README touched.** Root `README.md` and all five files under `docs/` are free of both `crash` and `2000`.
- **Pre-existing external state degrades safely.** An operator carrying `crash.enabled` or `crash.auth-*` in a local override or environment now hits an unbound property that Spring Boot ignores rather than a startup failure, and the resulting behavior (no shell) is the intended one.

### Operational

- **The `Closes #61` reference is the one thing I would change before merge.** `tsanetgit/Connect_SDK#61`'s acceptance item 5 asks `security-audit.py` to gain a check that FAILs on a shipped `crash.enabled=true` alongside a committed `crash.auth-password`. The PR correctly declares that out of scope, and #61 itself says it is "tracked separately in the security register". I could not find that tracking anywhere: `git grep -in "security register"` on the branch is empty, there is no register under `docs/`, and an org-wide `gh search issues --owner tsanetgit` for `security-audit` and for `crash` surfaces no issue covering it other than #61 itself.

  That the floor genuinely misses this class is verifiable: running `scripts/security-audit.py` against `origin/main`, which still carries `crash.auth-password: crash` in `application.yml` and in `tsanet-demo-crash-ssh.properties`, reports `PASS [credentials] no embedded credentials in source` and `0 fail`. So merging as written closes the only surviving record of an unmet acceptance item, for a check that is demonstrably absent.

  Either file the follow-up issue first and cite it here, or change `Closes #61` to `Refs #61` and close #61 by hand once the follow-up exists. Not a merge blocker on its own, and I would not hold a HIGH-severity RCE deletion for it.

- **`scripts/` now contains only `security-audit.py`.** The login-to-`authorized: true` assertion `simple-test.sh` carried has no console-driven successor. Correct to delete it here, worth a follow-up if that smoke had value.

- No env var, schema, dependency-addition, or restart implications. This module is not in a shipped artifact, so nothing needs redeploying.

### Nits

- The PR body says `git grep -in crash` leaves "only the unrelated 'a crash mid-save' comment". There are two unrelated hits; the second is `scripts/security-audit.py:269`. The conclusion is unaffected.
- The PR is still marked draft. Flip it to ready when the `Closes #61` question above is settled.

shawn-tsanet:
Follow-up to the Operational note in my review: the `security-audit.py` gap is now tracked in `tsanetgit/Connect_SDK#63`, and `tsanetgit/Connect_SDK#61` carries a comment pointing at it.

That resolves the only reason I flagged `Closes #61`. The unmet acceptance item now has a home outside #61's body, so `Closes #61` is fine as written and this PR can merge without stranding it.

One correction to my own note, from investigating #63: the check #61 asked for would not actually have caught this class. `SOURCE_EXT` already includes `.properties`/`.yml`/`.yaml`, so the files were scanned and the detector still passed. The pattern requires a quoted value of 8 or more characters, and unquoted is the convention in those formats. #63 fixes the matcher and pins the crash pair as a regression fixture instead.

No change requested here. Verdict stands at merge with notes.

Issue #65, opened 2026-08-21
connect-library: first-class retention/eviction API for the SQLite cache


shawn-tsanet:
The connect-library store caches case content — `collaboration_request.summary`, `case_note.summary`/`description` (full note text) plus creator identity, `case_response` engineer contact details and next steps, and `webhook_inbound_event.raw_payload` — with no eviction: rows persist until the file is deleted. For any consumer whose deployment faces a security or data-retention review, that means case content outlives its operational usefulness in both the live store and whatever backup path the file rides.

The Fin/Intercom adapter now runs an external age-based sweep against the file (`tsanetgit/Fin-Intercom_App#69`, from the review finding `tsanetgit/Fin-Intercom_App#66`). Its semantics, offered as the starting spec:

- **Terminal cases only** (status `CLOSED`/`REJECTED`) older than a configurable window, judged by `COALESCE(updated_at, fetched_at)`, deleted with their `case_note`/`case_response`/`attachment_config`/`attachment_forward_result` children in one transaction (children first, chunked IN-lists). **OPEN cases never age out** — unresolved is live state, not residue.
- Aged orphan children (parents evicted earlier) and `webhook_inbound_event` rows by `received_at`.
- Reference/config tables (`user_context`, `partner_selection`, `collaboration_request_form`, `webhook_subscription`) untouched.
- Safe because the store is a cache, not the source of truth: consumers re-fetch from the Connect API by token.

**Proposal:** a first-class `retentionSweep(Duration window)` (or equivalent) on the storage layer, same semantics, so the schema knowledge lives with its owner and every SDK consumer gets retention without reimplementing it. The adapter-side sweep is explicitly interim and will be retired in favor of the library call when this lands — its external copy of the schema shape is the maintenance argument for doing this upstream.

Filed from the Fin adapter work; happy to bring the sweep's tests and semantics over as the starting point.

shawn-tsanet:
As offered: the adapter sweep is merged (`tsanetgit/Fin-Intercom_App` PR #69, main @ c54026d), and here are its semantics and tests as the starting spec for the library API. Written self-contained so nothing below requires opening the adapter repo; permalinks at the end for convenience.

## Proposed API shape

```java
/** Returns per-table deleted counts; caller supplies the clock for testability. */
RetentionSweepResult retentionSweep(Duration window, Instant now);
```

[...]

PR #66, opened 2026-08-26
feat(storage): CacheRetention — age-based eviction for the SQLite cache

shawn-tsanet:
Implements the retention API proposed on #65: the SQLite cache holds case content (collaboration summaries, note text, response details, raw webhook payloads) with no eviction; this puts the eviction where the schema lives.

## API (naming is yours to bikeshed, @slava-vasianovych)

```java
CacheRetention.sweep(JdbcTemplate jdbc, Duration window, Instant now) → SweepResult
```

Static on `JdbcTemplate`, mirroring `DatabaseInitializer`; config (window source, scheduling, on/off) stays the consumer's concern. The library is silent: `SweepResult` carries per-table counts for the caller's audit line, and a mid-run failure throws `RetentionSweepException` naming the failed phase **with the committed-so-far counts** — a failed run can still have deleted rows, and the caller's audit trail should say so. Invalid windows (null/zero/negative) throw `IllegalArgumentException`. The open placement question from #65 (storage service vs maintenance type): this is the maintenance-type answer; happy to reshape.

## Semantics

Verbatim from the adapter's interim sweep (`tsanetgit/Fin-Intercom_App#69`), spec'd in full on #65: terminal cases only (`CLOSED`/`REJECTED`, age by `COALESCE(updated_at, fetched_at)` against an `Instant.toString()` cutoff — the format this library's own repositories write); **OPEN never ages out**; doomed set + children in one transaction (chunked IN-lists, children first); aged orphans; `webhook_inbound_event` by `received_at`; reference tables untouched; `busy_timeout` only, `journal_mode` untouched; connection autocommit restored on exit.

## The one defect fixed beyond the port

The pre-PR review gate caught a subtle JDBC hazard the adapter never had (it closes its connection outright): restoring autocommit in `finally` while a failed transaction is open **commits the partial phase** — `setAutoCommit(true)` commits pending work. Fixed with an explicit `rollback()` in the catch (suppressed-exception safe), and pinned by a test that fails partway THROUGH a phase (`attachment_config` dropped, so phase 1 dies after its `case_note` deletes ran) and asserts those deletes rolled back. Red-probed: removing the rollback fails exactly that test — sqlite-jdbc really does commit the partial phase without it.

## Receipts

- Tests build the schema via `DatabaseInitializer.createSchema` itself — SQL and schema move together, which is half the point of #65.
- connect-library suite green (93 tests, 7 new); red probes: (1) inverting the three age comparisons fails the age-dependent cases; (2) removing the rollback fails the mid-phase test.
- Design lineage: these semantics were advisor-reviewed three times and live-verified during the adapter arc (receipts on `tsanetgit/Fin-Intercom_App#66`/#69); this PR is a semantics-verbatim port, declared in lieu of re-reviewing an unchanged design.

## Follow-through once merged

The adapter's interim sweep gets retired for this call (tracked on `tsanetgit/Fin-Intercom_App#66`), and its CI schema-compat test becomes unnecessary by construction. No consumer wiring in this PR.

### DIFF
--- a/connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
+++ b/connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
@@ -12,33 +12,56 @@ import org.springframework.jdbc.core.JdbcTemplate;
 /**
  * Age-based eviction for the SQLite cache this library owns (Connect_SDK#65). The
  * cache holds case content — collaboration summaries, note text, response details,
- * raw webhook payloads — and rows otherwise persist until the file is deleted. Safe
- * to evict because the store is a cache, not the source of truth: consumers re-fetch
- * from the Connect API by token.
+ * raw webhook payloads — and rows otherwise persist until the file is deleted.
+ * Re-fetchability is the criterion for what this class touches, not a blanket
+ * property of everything cached here: eligible rows can be re-fetched from the
+ * Connect API by token, so losing them costs a round trip, not the data.
  *
  * <p>Semantics (ported verbatim from the Fin/Intercom adapter's interim sweep, where
- * they were probed and red/green-tested — tsanetgit/Fin-Intercom_App#69):
+ * they were probed and red/green-tested — tsanetgit/Fin-Intercom_App#69 — then
+ * narrowed by round-3 QA on Connect_SDK#66):
  *
  * <ul>
  *   <li><b>Terminal cases only</b> ({@code CLOSED}/{@code REJECTED}) older than the
  *       window, age judged by {@code COALESCE(updated_at, fetched_at)}; {@code OPEN}
  *       cases never age out — unresolved is live state, not residue.</li>
- *   <li>Phase 1, one transaction: the doomed token set with its {@code case_note} /
- *       {@code case_response} / {@code attachment_config} /
- *       {@code attachment_forward_result} children, children before parents, chunked
- *       IN-lists. Phase 2: aged orphan children whose parent is already gone. Phase 3:
- *       {@code webhook_inbound_event} by {@code received_at}.</li>
+ *   <li>Two phases, each committed in its own transaction: phase 1 deletes the
+ *       doomed token set with its {@code case_note} / {@code case_response} /
+ *       {@code attachment_config} children, children before parents, chunked
+ *       IN-lists, all in one transaction. Phase 2, separately: {@code
+ *       webhook_inbound_event} by {@code received_at} — a delivery log, not case
+ *       content, aged independently of case status.</li>
+ *   <li>{@code attachment_forward_result} is never touched by this class:
+ *       {@code POST /v1/collaboration-requests/{token}/attachments} has no GET
+ *       counterpart, so it is a local record of an outbound action, not a cache of
+ *       remote state — eviction here would be unrecoverable. It carries no foreign
+ *       key to {@code collaboration_request} (see {@link DatabaseInitializer}), so a
+ *       doomed case's forward-result rows deliberately outlive it: phase 1 deletes
+ *       the case and its other children but leaves these behind, permanently
+ *       orphaned by {@code case_token}. Nothing reaps them; that is this class
+ *       choosing not to, not an oversight.</li>
+ *   <li>Children with no cached parent are never swept, at any age: three fetch
+ *       paths ({@code getNotes}, {@code getResponses}, {@code getAttachmentConfig})
+ *       cache {@code case_note}/{@code case_response}/{@code attachment_config}
+ *       without ever writing the parent {@code collaboration_request} row, so for a
+ *       webhook-driven consumer a parentless-but-live child is the steady state, not
+ *       a transient race. An earlier version of this class aged these out by
+ *       {@code fetched_at} regardless of the case's actual status on the server;
+ *       round-3 QA on #66 traced the three fetch paths and found that assumption
+ *       false. The tradeoff accepted instead: rows for cases this library never
+ *       learns the fate of accumulate until the parent is eventually cached and ages
+ *       out normally, or forever if it never is.</li>
  *   <li>Reference/config tables ({@code user_context}, {@code partner_selection},
  *       {@code collaboration_request_form}, {@code webhook_subscription}) untouched.</li>
- *   <li>{@code fetched_at}, {@code forwarded_at}, and {@code received_at} are
- *       {@code Instant.toString()} (see the repositories) — UTC, so a lexicographic
- *       compare against a same-format cutoff is correct to within fraction-length
- *       noise at the boundary. {@code updated_at}, the first {@code COALESCE} term
- *       phase 1 ages by, is written as {@code OffsetDateTime.toString()} straight from
- *       the API response instead and can carry a non-UTC offset, so its lexicographic
- *       compare against the cutoff isn't the same tight guarantee — bounded skew, not
- *       exact, and not a concern at retention-window scale, but a real gap versus the
- *       other three columns (QA round 3 on #66).</li>
+ *   <li>{@code fetched_at} and {@code received_at} are {@code Instant.toString()}
+ *       (see the repositories) — UTC, so a lexicographic compare against a
+ *       same-format cutoff is correct to within fraction-length noise at the
+ *       boundary. {@code updated_at}, the first {@code COALESCE} term phase 1 ages
+ *       by, is written as {@code OffsetDateTime.toString()} straight from the API
+ *       response instead and can carry a non-UTC offset, so its lexicographic
+ *       compare against the cutoff isn't the same tight guarantee — bounded skew,
+ *       not exact, and not a concern at retention-window scale, but a real gap
+ *       versus the other two columns (QA round 3 on #66).</li>
  * </ul>
  *
  * <p>The library stays silent (no logging): the result carries the per-table counts
@@ -59,12 +82,9 @@ public final class CacheRetention {
     }
 
     /** Per-table deleted counts for one sweep run. */
-    public record SweepResult(int cases, int notes, int responses, int attachConfigs,
-            int attachForwards, int orphanNotes, int orphanResponses, int orphanAttachConfigs,
-            int orphanAttachForwards, int events) {
+    public record SweepResult(int cases, int notes, int responses, int attachConfigs, int events) {
         public int total() {
-            return cases + notes + responses + attachConfigs + attachForwards
-                    + orphanNotes + orphanResponses + orphanAttachConfigs + orphanAttachForwards + events;
+            return cases + notes + responses + attachConfigs + events;
         }
     }
 
@@ -113,11 +133,6 @@ public final class CacheRetention {
         int notes = 0;
         int responses = 0;
         int attachCfg = 0;
-        int attachFwd = 0;
-        int orphanNotes = 0;
-        int orphanResponses = 0;
-        int orphanAttachCfg = 0;
-        int orphanAttachFwd = 0;
         boolean originalAutoCommit = c.getAutoCommit();
         CacheRetentionRepository repo = new CacheRetentionRepository(c);
         try {
@@ -129,22 +144,13 @@ public final class CacheRetention {
                 notes += repo.deleteByTokens("case_note", "case_token", chunk);
                 responses += repo.deleteByTokens("case_response", "case_token", chunk);
                 attachCfg += repo.deleteByTokens("attachment_config", "case_token", chunk);
-                attachFwd += repo.deleteByTokens("attachment_forward_result", "case_token", chunk);
                 cases += repo.deleteByTokens("collaboration_request", "token", chunk);
             }
-            c.commit();
-            phase = "orphans";
-            // The age check protects a child cached mid-fetch before its parent lands.
-            orphanNotes = repo.deleteOrphans("case_note", "fetched_at", cutoff);
-            orphanResponses = repo.deleteOrphans("case_response", "fetched_at", cutoff);
-            orphanAttachCfg = repo.deleteOrphans("attachment_config", "fetched_at", cutoff);
-            orphanAttachFwd = repo.deleteOrphans("attachment_forward_result", "forwarded_at", cutoff);
             c.commit();
             phase = "webhook-events";
             int events = repo.deleteExpiredWebhookEvents(cutoff);
             c.commit();
-            return new SweepResult(cases, notes, responses, attachCfg, attachFwd,
-                    orphanNotes, orphanResponses, orphanAttachCfg, orphanAttachFwd, events);
+            return new SweepResult(cases, notes, responses, attachCfg, events);
         } catch (Exception e) {
             // Roll back the failing phase EXPLICITLY before the finally restores
             // autocommit: setAutoCommit(true) on a connection with an open
@@ -160,17 +166,11 @@ public final class CacheRetention {
                 e.addSuppressed(rollbackFailed);
             }
             boolean terminalCommitted = !"terminal-cases".equals(phase);
-            boolean orphansCommitted = "webhook-events".equals(phase);
             throw new RetentionSweepException(phase, new SweepResult(
                     terminalCommitted ? cases : 0,
                     terminalCommitted ? notes : 0,
                     terminalCommitted ? responses : 0,
                     terminalCommitted ? attachCfg : 0,
-                    terminalCommitted ? attachFwd : 0,
-                    orphansCommitted ? orphanNotes : 0,
-                    orphansCommitted ? orphanResponses : 0,
-                    orphansCommitted ? orphanAttachCfg : 0,
-                    orphansCommitted ? orphanAttachFwd : 0,
                     0), e);
         } catch (Error e) {
             // Same partial-commit hazard as above, but an Error isn't wrapped: don't
--- a/connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
+++ b/connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
@@ -1,0 +1,228 @@
+package com.tsanet.api.storage;
+
+import java.sql.Connection;
+import java.sql.PreparedStatement;
+import java.sql.ResultSet;
+import java.sql.SQLException;
+import java.sql.Statement;
+import java.time.Duration;
+import java.time.Instant;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Objects;
+import org.springframework.jdbc.core.ConnectionCallback;
+import org.springframework.jdbc.core.JdbcTemplate;
+
+/**
+ * Age-based eviction for the SQLite cache this library owns (Connect_SDK#65). The
+ * cache holds case content — collaboration summaries, note text, response details,
+ * raw webhook payloads — and rows otherwise persist until the file is deleted. Safe
+ * to evict because the store is a cache, not the source of truth: consumers re-fetch
+ * from the Connect API by token.
+ *
+ * <p>Semantics (ported verbatim from the Fin/Intercom adapter's interim sweep, where
+ * they were probed and red/green-tested — tsanetgit/Fin-Intercom_App#69):
+ *
+ * <ul>
+ *   <li><b>Terminal cases only</b> ({@code CLOSED}/{@code REJECTED}) older than the
+ *       window, age judged by {@code COALESCE(updated_at, fetched_at)}; {@code OPEN}
+ *       cases never age out — unresolved is live state, not residue.</li>
+ *   <li>Phase 1, one transaction: the doomed token set with its {@code case_note} /
+ *       {@code case_response} / {@code attachment_config} /
+ *       {@code attachment_forward_result} children, children before parents, chunked
+ *       IN-lists. Phase 2: aged orphan children whose parent is already gone. Phase 3:
+ *       {@code webhook_inbound_event} by {@code received_at}.</li>
+ *   <li>Reference/config tables ({@code user_context}, {@code partner_selection},
+ *       {@code collaboration_request_form}, {@code webhook_subscription}) untouched.</li>
+ *   <li>Timestamps here are {@code Instant.toString()} (see the repositories), so a
+ *       lexicographic compare against a same-format cutoff is correct to within
+ *       fraction-length noise at the boundary — irrelevant at retention scale.</li>
+ * </ul>
+ *
+ * <p>The library stays silent (no logging): the result carries the per-table counts
+ * for the caller's audit line, and a mid-run failure throws
+ * {@link RetentionSweepException} naming the failed phase with the counts from the
+ * phases that already COMMITTED — a failed run can still have deleted rows, and the
+ * caller's audit trail should say so. The failing phase's own deletes roll back.
+ *
+ * <p>Static on {@link JdbcTemplate}, like {@link DatabaseInitializer}: config (window
+ * source, scheduling, enable/disable) is the consumer's concern.
+ */
+public final class CacheRetention {
+
+    /** Chunk size for IN-list deletes: works on every sqlite build (no DELETE..LIMIT). */
+    private static final int CHUNK = 500;
+
+    private CacheRetention() {
+    }
+
+    /** Per-table deleted counts for one sweep run. */
+    public record SweepResult(int cases, int notes, int responses, int attachConfigs,
+            int attachForwards, int orphanNotes, int orphanResponses, int orphanAttachConfigs,
+            int orphanAttachForwards, int events) {
+        public int total() {
+            return cases + notes + responses + attachConfigs + attachForwards
+                    + orphanNotes + orphanResponses + orphanAttachConfigs + orphanAttachForwards + events;
+        }
+    }
+
+    /** A mid-run failure: {@code phase} failed, {@code committed} is what already landed. */
+    public static final class RetentionSweepException extends RuntimeException {
+        private final String phase;
+        private final SweepResult committed;
+
+        RetentionSweepException(String phase, SweepResult committed, SQLException cause) {
+            super("retention sweep failed in phase " + phase + " (committed so far: "
+                    + committed.total() + " rows)", cause);
+            this.phase = phase;
+            this.committed = committed;
+        }
+
+        public String phase() {
+            return phase;
+        }
+
+        /** Counts from the phases that committed before the failure. */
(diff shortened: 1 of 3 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
Two new files, both trace cleanly to "implement the retention API from #65."

- `CacheRetention.java` — the sweep implementation. Traces to the API/semantics sections of the PR body.
- `CacheRetentionTest.java` — the test matrix. Traces to the "Receipts" section (7 new tests claimed; I count 7 `@Test` methods).

No incidental tidying, no unrelated files, no scope creep. Both files are new (+403 −0), consistent with a green-field addition. Pass.

**Correctness:**

Traced the happy path through `terminalOldEvictsWithChildren_freshAndOpenSurvive`: cutoff = `2026-07-22T12:00:00Z`. `gone-closed` (updated_at OLD `2026-07-07…`) < cutoff → doomed. `gone-rejected` (updated_at NULL → COALESCE falls to fetched_at OLD) → doomed. `stay-fresh` (updated_at FRESH) survives. `stay-open` (OPEN, not in `('CLOSED','REJECTED')`) survives regardless of age. `doomedTokens` returns 2, chunk-deletes children-first then parents. Counts match the assertions. Correct.

One real correctness concern:

- **Children of the fresh survivor are not orphan-swept but ARE OLD.** In `seedChildren`, note the two doomed cases `gone-closed`/`gone-rejected` — their children are deleted in phase 1 by token. But consider a case that is **terminal-and-fresh** (`stay-fresh`, CLOSED, FRESH) whose children were seeded with an **OLD** age. Phase 1 skips them (parent not doomed). Phase 2 (`orphanDelete`) skips them because `case_token IN (SELECT token FROM collaboration_request)` — the parent still exists. So old children of a still-cached fresh terminal parent persist until the parent itself ages out. That is arguably correct-by-design (child lifetime bound to parent), and the test seeds `stay-fresh` children as FRESH so it never exercises the OLD-child-of-fresh-parent case. Not a bug, but the semantics ("age by parent, not by child, while parent lives") aren't asserted. Follow-up, not blocking.

- **`orphanDelete` uses `NOT IN (SELECT token …)`** — the comment correctly notes `token` is NOT NULL UNIQUE so no NULL-swallow hazard. Verified the guard is sound: if the subquery ever returned a NULL, `NOT IN` would return no rows, but the NOT NULL constraint prevents that. Correct, and the reasoning is asserted in a test (`orphanChildrenAgeOut`).

- **Partial-failure contract:** traced `partialFailureKeepsCommittedPhases`. Phase 1 commits (cases=1), phase 2 commits (0 orphans), phase 3 throws on missing table. In catch: `phase="webhook-events"`, `terminalCommitted = !"terminal-cases".equals("webhook-events")` = true, `orphansCommitted = "webhook-events".equals(phase)` = true. So committed result carries cases=1, orphan counts, events=0. Matches assertion. Correct.

  Note the `orphansCommitted` flag is only true when `phase=="webhook-events"` — meaning if phase 2 (`orphans`) itself throws, orphan counts report 0 even though phase 1 committed. That's right: phase 2 hadn't committed. But it means the orphan `int` locals accumulate before `c.commit()`, and a throw mid-phase-2 correctly reports them as 0. Verified logically consistent.

- **`journal_mode` untouched / `busy_timeout` only** — matches the stated semantics. The PRAGMA runs before `setAutoCommit(false)`, which is correct (PRAGMA inside a transaction can be ignored on some sqlite builds).

**Security:**

- Table and column names are interpolated into SQL strings (`deleteByTokens`, `orphanDelete`, `"DELETE FROM " + table`). These are **all compile-time constant literals** passed from `run()` — never external input. Values (tokens, cutoff) go through `setString` parameterization. No injection surface. Confirmed by walking every call site.
- No secrets, no credentials, no logging of payloads.
- `raw_payload` cache content is deleted, not logged. Clean.

**Regression risk:**

- **New class, no existing consumers.** PR body states "No consumer wiring in this PR." The adapter's interim sweep (`Fin-Intercom_App#66`) is retired *after* merge, not in this diff — so nothing currently calls `CacheRetention.sweep`. Zero regression to existing runtime paths.
- **`DatabaseInitializer.createSchema` is invoked by the test only** — the diff doesn't modify it. No blast radius there.
- **Schema-shape dependency (blast-radius edge #8):** the sweep hard-codes column names (`updated_at`, `fetched_at`, `forwarded_at`, `received_at`, `status`) and table names against the *current* schema. If `DatabaseInitializer` schema drifts, this silently breaks. The test builds schema from `createSchema` itself, which is the right coupling — schema drift breaks the test. Good. But this is a library shipping to consumers whose *existing* cache files predate this code; `sweep` runs against pre-existing data. The tests use a persisted file (`target/test-cache-retention.db`) and drop/recreate swept tables, so pre-existing-data behavior is partially exercised. Acceptable.

**Recommendation:** Merge — this is a clean green-field addition with no live consumers and a strong test matrix; track the "old-child-of-fresh-terminal-parent" semantic as a post-merge note if that lifetime coupling is ever questioned.

---

shawn-tsanet:
**Disposition on the auto-QA note (old child of a fresh terminal parent):**

Confirmed by tracing the code directly, not just the review text. A terminal case whose own `updated_at`/`fetched_at` stays fresh keeps its children out of both sweep phases: phase 1 skips them because the parent isn't in the doomed set, and phase 2's orphan check skips them because the parent still exists in `collaboration_request`. So an old child row can persist as long as its parent stays fresh.

Accepted as-is. Child lifetime bound to a live parent is the intended coupling here, not a bug. Not adding a test for it now since it doesn't change behavior; leaving this comment so it's a documented decision instead of a surprise later.

🤖 via Claude Code

connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
```diff
@@ -0,0 +1,228 @@
...
+                // restoring autocommit failed; the connection is being discarded anyway
+            }
+        }
+    }
+
+    /** Terminal set: CLOSED + REJECTED; PENDINGACTION is deliberately not terminal. */
+    private static List<String> doomedTokens(Connection c, String cutoff) throws SQLException {
+        List<String> tokens = new ArrayList<>();
+        try (PreparedStatement ps = c.prepareStatement(
+                "SELECT token FROM collaboration_request WHERE status IN ('CLOSED','REJECTED')"
```
slava-vasianovych: We do use ORM in this codebase, so would be better to implement it rather via ORM, or named query in isolated repository implementation. 
I can take that if you want :) 


It's ok to use raw SQL in tests, but would be better to use mapped queries in library code.

connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
```diff
@@ -0,0 +1,228 @@
...
+                    + orphanNotes + orphanResponses + orphanAttachConfigs + orphanAttachForwards + events;
+        }
+    }
+
+    /** A mid-run failure: {@code phase} failed, {@code committed} is what already landed. */
+    public static final class RetentionSweepException extends RuntimeException {
+        private final String phase;
+        private final SweepResult committed;
+
+        RetentionSweepException(String phase, SweepResult committed, SQLException cause) {
```
slava-vasianovych: Exception handling and elegant exception hierarchy is in my todo list, So, just a side-note: exceptions will be changed soon by my PR.

slava-vasianovych (commented):
Minor comments. 

Up to you - you can rather merge it, and I can cover the findings, or you can try to cover the SQL comment. 

Looks good.

connect-library/src/main/java/com/tsanet/api/storage/CacheRetention.java
shawn-tsanet: Moved to `CacheRetentionRepository` (e3e99b3) — same SQL, isolated out of the sweep logic as you suggested.

Kept it Connection-threaded rather than JdbcTemplate-based: the sweep holds one connection across three phases with its own explicit commit/rollback, and JdbcTemplate would check out its own connection per call (breaking that) and wrap SQLException in DataAccessException (the type the phase-attribution catch block depends on). No ORM lives in this codebase yet, so I took the "named query in isolated repository" half of your suggestion rather than the ORM half. Repository is package-private since it only works correctly inside that caller-managed transaction — not a general-purpose repository like its siblings.

🤖 via Claude Code

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pure refactor: SQL extracted verbatim from `CacheRetention.run()` into new package-private `CacheRetentionRepository`. Traces cleanly to the reviewer's inline comment (CacheRetention.java:191) asking for isolated named queries. Removed imports in `CacheRetention.java` (`PreparedStatement`, `ResultSet`, `Statement`, `ArrayList`) trace to the extraction. Pass.

**Correctness:**
Verified the SQL moved byte-identical: `findDoomedTokens`, `deleteByTokens`, `deleteOrphans`, `deleteExpiredWebhookEvents` carry the same query strings, parameter binding, and comments as the originals. `setBusyTimeout(5000)` now builds the PRAGMA via `"PRAGMA busy_timeout = " + millis` — `millis` is an `int` caller-supplied constant (5000), so no injection and no string-format surprise. The transaction orchestration (autocommit, per-phase commit, rollback-in-catch, phase attribution) stays entirely in `CacheRetention.run()` and is unchanged by this diff. The repo wraps the caller's connection and issues no commit/rollback of its own — correct, since the sweep owns the transaction. Behavior identical to the reviewed version.

**Security:**
Table/column names still compile-time constants from `run()`; values still `setString`-parameterized. `busy_timeout` now interpolates an `int`, not a string — no new surface. No secrets, no logging. Clean.

**Regression risk:**
`CacheRetentionRepository` is new, package-private, and called only from `CacheRetention.run()`. No existing consumers of either class (PR still ships no wiring). The extraction preserves the exact call sequence, so the test matrix from the prior review (including the mid-phase rollback probe) exercises the same paths. Zero blast radius outside this package.

**Carried forward:**
- Old child of fresh terminal parent — RESOLVED (accepted as documented design decision in discussion).
- SQL in library code should be isolated repository — RESOLVED (extracted to `CacheRetentionRepository` this commit).
- Exception hierarchy will change soon (reviewer note) — FOLLOW-UP (reviewer's own future PR).

**Recommendation:** Merge — clean verbatim extraction addressing the reviewer's isolation comment, no behavior change.

---

shawn-tsanet (commented):
## QA review of `feat(storage): CacheRetention` (round 3, head `e3e99b3`)

**Verdict:** ❌ Needs changes

Suite verified green on the PR head, not taken from the body: `mvn test` in `connect-library` against a clean detached worktree at `e3e99b3` reports `Tests run: 93, Failures: 0, Errors: 0` / `BUILD SUCCESS`, with `CacheRetentionTest` at 7/7. Both claimed red probes reproduce exactly as described: removing the explicit `rollback()` fails `midPhaseFailureRollsBackThatPhaseEntirely` and only that one; inverting the three age comparisons fails 6 of 7 (all but `invalidWindowsAreRejected`). The receipts in the body hold.

Two blocking findings below are things the earlier rounds did not reach, and three prose claims in the new javadoc are not true of this library.

---

### Correctness

**1. BLOCKING. A non-`SQLException` escape re-opens the exact hazard this PR says it closed.**

`CacheRetention.java:142` catches only `SQLException`. A `RuntimeException` or `Error` raised inside a phase skips the rollback entirely, falls through to the `finally` at line 168, and `setAutoCommit(true)` commits the partial phase. That is verbatim the failure the comment at line 143 declares as the gate finding, just on the other exception path.

Probed rather than asserted. I injected an `IllegalStateException` between the `case_note` and `case_response` deletes in phase 1, seeded one doomed case with one note, and ran the sweep:

```
QA RTE PROBE surviving notes = []
```

The `case_note` delete that had already executed was committed. The row is gone. Two contract breaks in one: the partial phase lands, and the caller receives a bare `IllegalStateException` instead of a `RetentionSweepException`, so it gets neither the phase name nor the committed-so-far counts that the whole design exists to guarantee.

A success flag consulted by the `finally`, or a `catch (Throwable)` that rolls back and rethrows, closes it. Worth a test on that path too, since the current matrix only fails via SQL.

**2. BLOCKING. `attachment_forward_result` is not re-fetchable, so sweeping it destroys the only record.**

The class doc (lines 15-17) rests the whole design on one premise: "Safe to evict because the store is a cache, not the source of truth: consumers re-fetch from the Connect API by token." That holds for four of the five swept tables. It does not hold for `attachment_forward_result`.

- The only attachments write path in the spec is `POST /v1/collaboration-requests/{token}/attachments` (`operationId: forwardAttachment`). There is no GET counterpart; the three sibling attachment paths are all `config`.
- The table is populated solely from that POST's response, via `AttachmentForwardResultRepository.saveAll` called from `ConnectApiAttachmentsGateway.forwardAttachments`.
- It is read back only from local storage: `AttachmentForwardResultRepository.findAll()` and `findByCaseToken()`. Nothing re-hydrates it.

So it is a local log of an outbound action, not a cache of remote state. The sweep evicts it twice over: phase 1 by token, phase 2 as an aged orphan on `forwarded_at`. After the window, a consumer's record that a forward happened, and what the receiver and submitter said about it, is gone with no way to recover it.

Either exclude it from both phases, or the doc has to stop claiming everything swept is re-fetchable and say plainly that the library deletes unrecoverable forward records at the window. My read is the former, but it is a product call.

**3. "OPEN never ages out" is not true on this SDK's token-only fetch paths.**

The PR body states it in bold and the class doc repeats it at lines 23-25. It holds for `collaboration_request` rows. It does not hold for a case's children, because three fetch paths cache children keyed by `case_token` and never write the parent row:

- `ConnectApiNotesGateway.getNotes(token)` stores notes at line 39, parent never stored.
- `ConnectApiResponsesGateway.getResponses(token)` at line 54. This one fetches the full parent DTO on line 49 and still stores only the responses.
- `ConnectApiAttachmentsGateway.getAttachmentConfig(token)` at line 51, and `forwardAttachments` at line 110.

For a consumer driven by inbound webhooks, which is the shape this library ships webhook support for, `webhook_inbound_event.request_token` to `getNotes(token)` is the main path, and parentless children are the **steady state**, not a transient. Phase 2 then ages them out on `fetched_at` with no knowledge of the case's status.

Probed: seeded aged `case_note` / `case_response` / `attachment_config` for a single token with no parent row, swept, and got `orphanNotes=1 orphanResponses=1 orphanAttachConfigs=1`. Three children of a case that is live and OPEN on the server, evicted.

The comment at line 131, "The age check protects a child cached mid-fetch before its parent lands", describes a race that assumes the parent eventually lands. On these paths it never does. Whichever way you resolve it (accept and rewrite the claim, skip orphan deletion for these tables, or upsert a stub parent, which is out of scope here), the current wording asserts an invariant the code does not hold.

This is the one place where "semantics-verbatim port" carries a cost: the adapter closed its own connection outright and controlled its own cache shape, so the orphan case really was residue there. In this host it is a normal state.

**4. `PENDINGACTION` does not exist in this API.**

`CacheRetentionRepository.java:39` reads: "Terminal set: CLOSED + REJECTED; PENDINGACTION is deliberately not terminal."

The `CaseStatus` enum is `OPEN, INFORMATION, REJECTED, ACCEPTED, CLOSED`. Identical on `origin/master` and `origin/beta` of `tsanetgit/Connect-API-Code`, and identical in the generated `CaseStatus.java` this module compiles against. `PENDINGACTION` appears exactly once in the whole repository: that comment. Another adapter-vocabulary artifact.

The partition itself is right, and I confirmed the casing matches so the `IN` clause actually fires: the generated enum is `REJECTED("REJECTED")` / `CLOSED("CLOSED")`, and `ConnectApiCollaborationGateway.toDto` stores `getStatus().getValue()`. But the comment documents a deliberate decision about a status that cannot occur, while saying nothing about `INFORMATION` and `ACCEPTED`, which are the two non-terminal statuses that actually exist and that will therefore accumulate in the cache forever.

**5. The `Instant.toString()` claim is wrong for the one column phase 1 ages by first.**

Class doc lines 33-35: "Timestamps here are `Instant.toString()` (see the repositories), so a lexicographic compare against a same-format cutoff is correct to within fraction-length noise at the boundary."

True for `fetched_at`, `forwarded_at` and `received_at`. I checked all five writers: `CaseNoteRepository:25`, `CaseResponseRepository:25`, `AttachmentConfigRepository:24`, `AttachmentForwardResultRepository:27`, `WebhookInboundEventRepository:33`, all `Instant.now().toString()`.

Not true for `updated_at`, which is the first term of the `COALESCE` phase 1 ages by. It is written straight from the API through `ConnectApiCollaborationGateway.formatDateTime` (line 88), which is `OffsetDateTime.toString()`, so it renders whatever offset the API sends. The residual error is therefore offset skew of up to roughly 14 hours in either direction, not fraction-length noise. Still bounded and still harmless against a 30-day window, so this is a doc fix rather than a code fix, but the stated bound is the wrong bound, and no test carries an offset-format `updated_at` so nothing exercises it.

---

### Security

No concerns. Table and column names reaching the SQL builders are compile-time constants supplied by `run()`, never caller data; all values go through `setString`; the `busy_timeout` PRAGMA concatenates an `int`. No secrets, no logging, nothing that reads external input.

### Regression risk

Low, and unchanged from round 2. Both new classes are additive, `CacheRetentionRepository` is package-private, and nothing calls `CacheRetention.sweep` yet, so nothing existing changes behavior. The blast radius arrives with the consumer wiring, which is why findings 2 and 3 are worth settling before that lands rather than after.

One coverage gap worth naming rather than demanding: chunking above `CHUNK = 500` has no test. I ran it, 1200 doomed cases with 1200 notes, and it swept all of them correctly across three chunks, so the path works. It is unpinned, not broken.

### Operational

Nothing to do on merge. No wiring, no new dependency, no schema change, no env var, no service restart.

One thing outside the PR: the local clone at `~/projects/tsanetgit-Connect_SDK` is carrying an uncommitted javadoc addition to `CacheRetentionRepository.java` (the "two reasons this can't be JdbcTemplate-shaped" paragraph) that is not on the branch. It reads like the prose half of the reply to the SQL-isolation comment. Worth pushing or discarding so it does not go stale.

### Notes

- `fetched_at` is refreshed on every re-fetch (`ON CONFLICT ... fetched_at = excluded.fetched_at`). A case whose `updated_at` is NULL therefore never ages out at all while any consumer keeps calling `getCollaborationRequests()`, since the COALESCE fallback is restamped on each poll. Fail-safe direction, so a note rather than a finding, but the fallback is not an age for an actively-refreshed cache.
- The test class points at a persistent file DB and drops/recreates the six swept tables in `@BeforeEach`. Correct serially; it would collide under parallel execution.
- Slava's exception-hierarchy comment is a standing side-note about a future PR, not something this one has to absorb.

🤖 via Claude Code

shawn-tsanet:
**Disposition on round-3 QA** (reviewed head `e3e99b3`; fixes now at [72d1bd8](https://github.com/tsanetgit/Connect_SDK/commit/72d1bd8)):

1. **BLOCKING — non-SQLException escape re-opens the partial-commit hazard.** Fixed. Widened `catch (SQLException e)` to `catch (Exception e)` (still wraps in `RetentionSweepException`, whose cause type is now `Exception`), plus a new `catch (Error e)` that rolls back and rethrows raw rather than wrapping. New regression test reproduces the reviewer's exact probe (a `java.lang.reflect.Proxy`-wrapped `Connection`/`DataSource`, no new test dependency, throwing between the `case_note` and `case_response` deletes). Red-probed twice: reverting the catch to `SQLException`-only fails only the new test; separately pulling just the `c.rollback()` call (keeping the wider catch) fails the new test's *and* the existing mid-phase test's rollback assertion identically — confirms both assertions exercise the same live rollback path, not a vacuous DB.

2. **BLOCKING — `attachment_forward_result` is not re-fetchable.** Not fixed. Agreed this is a product call, not mine to make silently. Holding for a decision: exclude the table from both sweep phases, or is losing the forward record after the window acceptable?

3. **Children can be evicted while the parent case is still OPEN, via the orphan-sweep path.** Not fixed, same reason — this and #2 are both "is silent unrecoverable eviction OK here" calls. Scoped the class doc's "never age out" claim to phase 1 only so it stops contradicting this finding while it's open.

4. **`PENDINGACTION` doesn't exist in this API.** Fixed. Verified against the generated `CaseStatus.java` (`OPEN, INFORMATION, REJECTED, ACCEPTED, CLOSED` — no `PENDINGACTION` anywhere in the repo). Comment now names the real non-terminal statuses.

5. **`Instant.toString()` claim wrong for `updated_at`.** Fixed. Verified `updated_at` is written via `OffsetDateTime.toString()` in `ConnectApiCollaborationGateway.formatDateTime` (read the source directly). Class doc now distinguishes it from the three columns that are genuinely `Instant.toString()`.

Also: the uncommitted javadoc paragraph flagged under Operational (the "two reasons this can't be JdbcTemplate-shaped" addition) was a slip on my end — edited after staging, committed before re-staging. It's in 72d1bd8 now.

One thing my own local gate flagged that round-3 also touched on (regression risk / widened exception contract): grepped the whole repo for `CacheRetention.sweep` and `CacheRetention.RetentionSweepException` — zero callers outside the test file, so the widened catch changes nothing for any existing consumer.

Not addressed: the untested `catch (Error)` rollback path, and chunking above `CHUNK = 500` still has no dedicated test (round-3 ran it manually and confirmed it works). Noting both as follow-ups, not blocking.

94/94 tests green, CI green on 72d1bd8.

🤖 via Claude Code

shawn-tsanet:
**Findings 2 and 3, resolved** — Shawn's call, at [0a96257](https://github.com/tsanetgit/Connect_SDK/commit/0a96257):

- **`attachment_forward_result`**: excluded from the sweep entirely, both phases. No API read-back path, so eviction was unrecoverable.
- **Parentless children**: the orphan sweep is gone in its entirety (not just for the forward-result table) — three fetch paths cache children without ever writing the parent row, so it was silently evicting live case data with no status check.

The sweep is now 2 phases (terminal-cases, webhook-events) instead of 3. `SweepResult` drops from 10 fields to 5. Verified `CacheRetention.java` doesn't exist at the `v0.1.0` tag and has zero non-test consumers anywhere in the repo, so the field removal predates any published or consumed surface — not a breaking change to a real caller.

"OPEN never ages out" is unconditionally true again now that the orphan sweep is gone, rather than just documented around — removing the exception instead of caveating it.

Also checked something neither of us had: `attachment_forward_result` has no foreign key to `collaboration_request` and nothing sets `PRAGMA foreign_keys`, so a doomed case's forward-result rows deliberately outlive it as inert dangling data, not a delete-order hazard. Documented as intentional in the class javadoc, along with the accepted tradeoff that parentless children now persist until/unless their parent is eventually cached, or forever if it never is.

Tests: repurposed the orphan test to assert an old AND a fresh parentless row both now survive (the deleted test it replaces is still in git history as proof those same rows would have failed under the old code); added a survival assertion for `attachment_forward_result` across every seeded case in the main happy-path test, including the doomed one.

94/94 tests green, compiles clean, CI green on 0a96257. That's all 5 round-3 findings addressed or resolved by decision.

🤖 via Claude Code

shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**
Pass. Every hunk traces to a round-3 finding disposition: findings 1/4/5 fixed, findings 2/3 handled by removing the orphan-sweep entirely (`deleteOrphans` deleted, phase renumbering, `SweepResult` fields dropped) and rewriting the doc.

**Correctness:**
The round-3 disposition said findings 2 and 3 were "not fixed, holding for a product call." This diff goes further than the disposition text describes: it **removes the entire orphan-sweep phase** (`deleteOrphans`, phase 2) and drops `attachment_forward_result` from phase 1. That resolves both blocking findings by construction — orphans are never swept (finding 3), and forward-results are never touched (finding 2). The new doc says so explicitly and a test pins each:
- `parentlessChildrenAreNeverSwept_regardlessOfAge` seeds old+fresh parentless notes, asserts both survive and `r.notes()==0`. Traces correctly through the new code: `findDoomedTokens` returns no doomed set for a parentless child, and there's no orphan phase, so nothing deletes them.
- `terminalOldEvictsWithChildren...` now asserts `attachment_forward_result` retains all three tokens including `gone-closed`. Correct: phase 1 no longer deletes that table.

Finding 1 (Error/RuntimeException escape): `catch (Exception e)` now wraps in `RetentionSweepException`; new `catch (Error e)` rolls back and rethrows raw. Traced `nonSqlFailureMidPhaseStillRollsBackAndCarriesPhaseAndCommittedCounts`: injects `IllegalStateException` on the 3rd `prepareStatement` (case_response DELETE), asserts phase=`terminal-cases`, cause is `IllegalStateException`, committed total 0, and `case_note` still has `note-1` (rolled back). This is the reviewer's exact probe. Correct.

One live gap, carried by the author's own disposition: the `catch (Error e)` rollback path has no test. The new test exercises only the `Exception` path. Not blocking — Error mid-sweep is not a runtime path any consumer hits, and the rollback logic is identical to the tested `Exception` branch.

**Security:**
None. No new SQL surface; `deleteOrphans` (which concatenated a table name, though from constants) is gone. Values still `setString`-parameterized.

**Regression risk:**
`SweepResult` lost 6 fields (`attachForwards`, `orphan*`). Any consumer reading those would break — but the author grepped for `CacheRetention.sweep`/`RetentionSweepException` and found zero callers outside tests, and this PR ships no wiring. Blast radius is the test file, which is updated in this same diff. `RetentionSweepException` cause type widened `SQLException`→`Exception` — same zero-consumer argument.

**Carried forward:**
- Non-SQLException escape (partial commit) — RESOLVED (`catch (Exception)`/`catch (Error)`, tested).
- attachment_forward_result not re-fetchable — RESOLVED (dropped from phase 1 entirely).
- OPEN children evicted via orphan sweep — RESOLVED (orphan phase removed entirely).
- PENDINGACTION comment — RESOLVED (comment names real statuses).
- Instant.toString claim wrong for updated_at — RESOLVED (doc distinguishes updated_at as OffsetDateTime).
- catch(Error) rollback path untested — FOLLOW-UP (author-acknowledged; identical logic to tested branch).
- Chunking above CHUNK=500 untested — FOLLOW-UP (author-acknowledged).
- Exception hierarchy will change (Slava's PR) — FOLLOW-UP (reviewer's own future PR).

**Recommendation:** Merge — both round-3 blocking findings are resolved by removing the orphan/forward-result sweep entirely; the two remaining items are author-acknowledged follow-ups that don't hold the merge.

---

PR #67, opened 2026-08-28
Migrate to Spring Boot 4.1.1 and Jackson 3

shawn-tsanet:
Gets this repo off a Spring Boot line with no patch path. 3.3 left OSS support 2025-06-30 and commercial extended support 2026-06-30. Tracking issue: `tsanetgit/Fin-Intercom_App#76`. First of three hops (SDK -> Gateway -> Fin); this is the leaf both others consume.

**Draft on purpose.** The reactor is green, but this changes a published artifact's contract and no consumer has been compiled against it yet. See "Before this leaves draft".

## What it does

| Area | Change |
|---|---|
| Root pom | Boot parent 3.3.4 -> 4.1.1; openapi-generator 7.8.0 -> 7.21.0 |
| connect-library pom | `openApiNullable=false`, `useJackson3=true`, `useSpringBoot4=true`; jackson-databind -> `tools.jackson.core` groupId; drops jackson-datatype-jsr310 and jackson-databind-nullable |
| 4 hand-written files | `com.fasterxml.jackson.databind/core` -> `tools.jackson` |
| 3 Boot 4 breakages | `DataSourceAutoConfiguration` relocated; `@MockBean` -> `@MockitoBean`; `@WebMvcTest` -> new `spring-boot-webmvc-test` artifact |
| Wire format | `fail-on-null-for-primitives: false` in 3 apps + the library's own mapper |

`jackson-annotations` keeps its package in Jackson 3, which is why the 371 annotation imports here do not move.

## Receipts

**Full reactor green, test counts identical to unmodified main.** Built the baseline in a separate worktree rather than assuming:

```
                          this branch      origin/main (worktree)
connect-library                94                94
TSANet-integration-demo         6                 6
TSANet-integration-app         30                30
attachment-receiver     81 (16 skip)      81 (16 skip)
TSANet-demo-ui                 15                15
```

**openApiNullable=false is safe** — swept all three repos, zero consumers:
```
grep -rn "_JsonNullable" <each repo> --include="*.java" | grep -v /target/   # 0 hits in all three
```
Plain accessors are unchanged: `getCaseNumber()` still returns `String`. Only the `_JsonNullable` twins disappear.

**The generator refuses the alternative**, which independently confirms the dependency is a dead end:
```
IllegalArgumentException: openApiNullable=true is not compatible with useJackson3=true
(jackson-databind-nullable has no Jackson 3 release). Please set openApiNullable=false explicitly.
```
Confirmed on Maven Central: jackson-databind-nullable stops at 0.2.6, no Jackson 3 successor.

**The primitive-null fix is red-probed.** Removing `.disable(FAIL_ON_NULL_FOR_PRIMITIVES)` -> 1 failure; restoring -> green. Not decoration.

## PROMPT lines from `pre-pr-artifacts.py`

**Guard added, why not a representation change?** The representation fix is `long documentId` -> `Long` on `CollaborationRequestFormTemplateDto`. Deliberately not taken here: it changes a published DTO's shape in the same PR that moves the framework, so a rollback could not be reasoned about. Shawn's call was to preserve the accepted wire format and treat tightening as its own change. Recorded as a follow-up below.

**Newly reachable failure, receiving handler and user-visible outcome.** A payload omitting a primitive creator property would reach `ApiErrorHandler.handleOther(Exception)` and surface as **HTTP 502** with `{"error":"JSON parse error: Cannot map null into type long"}`. That is the observed pre-fix behavior, captured with `MockMvcResultHandlers.print()`. The fix makes it 200 again.

**Schema/stored shape changed, what does pre-existing state look like?** `CollaborationRequestFormRepository` reads `List<FormFieldDto>` out of the SQLite cache, and `FormFieldDto.required` is a primitive `boolean`. Rows written by this code always carry it (the same mapper writes all record components), but rows written by any earlier shape may not, and those would have failed to read back. The library mapper is configured to match, so old rows still load.

**Every behavioural claim cites its line.** The three `application.yml` comments, the `connect-library/pom.xml` `openApiNullable` comment, the `CollaborationRequestFormRepository` mapper comment, and the `CliJsonFiles` catch comment each describe the hunk they sit on.

**Abstraction centralising a decision / third+ fix to one class:** neither applies.

## Contract changes for consumers

Four, all affecting `Connect_Gateway` and `Fin-Intercom_App`:

1. Jackson types in the dependency surface move to `tools.jackson` — consumers must move to Jackson 3 in the same step.
2. `JsonNullable` accessors are gone from two generated DTOs (verified unused in all three repos).
3. Consumers inherit Jackson 3's strict `FAIL_ON_NULL_FOR_PRIMITIVES` unless they set it themselves. The library's own mapper is aligned; **an application.yml cannot travel inside a jar.**
4. **Jackson 3 parse exceptions are unchecked.** `JacksonException extends RuntimeException`; Jackson 2's `JsonProcessingException extended IOException`. Any consumer `catch (IOException)` around a mapper call still compiles and silently stops catching malformed JSON. This bit this repo — `CliJsonFiles` had exactly that shape and is fixed here. Consumers must be swept for it.

This warrants a major version bump on `connect-library`.

## Gate

`qa-review.py --local`: **CONCERNS**, one item — the yaml setting does not reach the library's own mappers. Fixed in `4bc77c4` rather than dispositioned, since `FormFieldDto.required` is a real primitive on a real deserialization path.

## Known gaps, stated not hidden

- **The `application.yml` setting has no test.** The test covers the converter, not the yaml. Delete those yaml lines and the suite stays green while runtime starts rejecting payloads. A `@JsonTest` would cover it; note that Boot 4 likely relocated `@JsonTest` the way it moved `@WebMvcTest`, so its module needs looking up rather than assuming.
- **The properties migrator produced no output and I could not prove it engaged.** It resolves onto the test classpath and the context boots on 4.1.1, but a `level=error` deprecated property taken from the 4.1.1 jars' own `spring-configuration-metadata.json` (`spring.http.encoding.charset`) produced no warning. So this is **unproven, not clean** — I am not claiming it found nothing.
- **`EnvironmentPostProcessor` is deprecated-for-removal in Boot 4**, implemented by `WebhookWebApplicationTypeEnvironmentPostProcessor`. Still functions; left alone under minimal-diff.

## Before this leaves draft

- [ ] Compile `Connect_Gateway` and `Fin-Intercom_App` against this as a SNAPSHOT
- [ ] Sweep both consumers for `catch (IOException)` around mapper calls (contract change 4)
- [ ] Decide the `connect-library` version bump
- [ ] Decide record-vs-flag as a separate change

## Blast radius (generated)

`scripts/pre-pr-artifacts.py --base origin/main` at `4bc77c40f2` vs `2eed253944` on `feat/spring-boot-4-jackson-3`. AUTO means evidence gathered below — you still judge every hit. "not triggered" means not triggered by these detectors; the pre-flight checklist still binds.

| pre-flight line | status |
|---|---|
| Positional contract renumbered/reordered -> consumer list produced by a command, in the PR body | not triggered by these detectors |
| Guard added -> PR says why a representation change was not the better move | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Check added or changed -> PR names the wrong-but-well-formed input it rejects, with a probe | not triggered by these detectors |
| Third+ fix to one class -> PR justifies not changing the representation | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Newly reachable failure -> PR names the receiving handler and the user-visible outcome | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Vocabulary gained a member -> old term grepped, every hit decided | not triggered by these detectors |
| Abstraction INTRODUCED to centralise a decision -> PR greps the OLD FORM of that decision (literal, inline condition, duplicated branch — not the new symbol) and decides every hit | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Schema/contract/stored shape changed -> PR states what pre-existing state looks like and shows the code working against it | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Every behavioural claim in the diff's prose (javadoc, decision record, commit message) cites the line that makes it true | PROMPT — judgment line, no detector; answer it in the PR body if it applies |
| Deliberate duplication asserted by a check, not a comment | not triggered by these detectors |

### DIFF
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CliJsonFiles.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CliJsonFiles.java
@@ -1,7 +1,8 @@
 package com.tsanet.facade.cli;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
+import tools.jackson.core.JacksonException;
+import tools.jackson.core.type.TypeReference;
+import tools.jackson.databind.ObjectMapper;
 import com.tsanet.api.connectapi.dto.NormalizedHttpsAttachmentConfigDto;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -17,7 +18,7 @@ final class CliJsonFiles {
     static Map<String, Object> readObjectMap(Path path) {
         try {
             return OBJECT_MAPPER.readValue(Files.readString(path), new TypeReference<>() {});
-        } catch (IOException ex) {
+        } catch (IOException | JacksonException ex) {
             throw new IllegalArgumentException("Failed to read JSON file " + path + ": " + ex.getMessage());
         }
     }
@@ -25,7 +26,7 @@ final class CliJsonFiles {
     static NormalizedHttpsAttachmentConfigDto readHttpsConfig(Path path) {
         try {
             return OBJECT_MAPPER.readValue(Files.readString(path), NormalizedHttpsAttachmentConfigDto.class);
-        } catch (IOException ex) {
+        } catch (IOException | JacksonException ex) {
             throw new IllegalArgumentException("Failed to read HTTPS config file " + path + ": " + ex.getMessage());
         }
     }
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CliJsonFiles.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CliJsonFiles.java
@@ -2,7 +2,9 @@ package com.tsanet.facade.cli;
 
 import tools.jackson.core.JacksonException;
 import tools.jackson.core.type.TypeReference;
+import tools.jackson.databind.DeserializationFeature;
 import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.json.JsonMapper;
 import com.tsanet.api.connectapi.dto.NormalizedHttpsAttachmentConfigDto;
 import java.io.IOException;
 import java.nio.file.Files;
@@ -10,7 +12,13 @@ import java.nio.file.Path;
 import java.util.Map;
 
 final class CliJsonFiles {
-    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    // Jackson 3 flipped FAIL_ON_UNKNOWN_PROPERTIES to lenient by default; Jackson 2
+    // rejected unknown keys. Re-enabled so a typo'd key in an operator-authored
+    // config file fails loudly (as on main) instead of silently leaving the field
+    // unset. The Map reads are unaffected: the feature only applies to typed binding.
+    private static final ObjectMapper OBJECT_MAPPER = JsonMapper.builder()
+        .enable(DeserializationFeature.FAIL_ON_UNKNOWN_PROPERTIES)
+        .build();
 
     private CliJsonFiles() {
     }
--- a/TSANet-integration-app/src/main/resources/application.yml
+++ b/TSANet-integration-app/src/main/resources/application.yml
@@ -1,3 +1,13 @@
+spring:
+  jackson:
+    deserialization:
+      # Jackson 3 flipped FAIL_ON_NULL_FOR_PRIMITIVES to true by default, so a
+      # payload that omits a primitive creator property is now rejected where
+      # Jackson 2 defaulted it (0/false). Restored to the Jackson 2 default so
+      # the Boot 4 upgrade does not change the accepted wire format. Tightening
+      # this is a deliberate contract change, not an upgrade side effect.
+      fail-on-null-for-primitives: false
+
 tsanet:
   api:
     base-url: "http://localhost:8080"
--- a/TSANet-integration-app/src/test/java/com/tsanet/facade/cli/CliJsonFilesTest.java
+++ b/TSANet-integration-app/src/test/java/com/tsanet/facade/cli/CliJsonFilesTest.java
@@ -1,0 +1,47 @@
+package com.tsanet.facade.cli;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.nio.file.Files;
+import java.nio.file.Path;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class CliJsonFilesTest {
+    @TempDir
+    Path tempDir;
+
+    @Test
+    void itRejectsUnknownKeysInHttpsConfig() throws Exception {
+        // Wrong-but-well-formed: valid JSON whose key is a typo of httpsPort. A
+        // default Jackson 3 mapper parses this silently with httpsPort unset;
+        // this locks in the FAIL_ON_UNKNOWN_PROPERTIES restoration.
+        Path config = tempDir.resolve("https-config.json");
+        Files.writeString(config, "{\"domain\":\"example.com\",\"httpsPortt\":443}");
+
+        assertThatThrownBy(() -> CliJsonFiles.readHttpsConfig(config))
+            .isInstanceOf(IllegalArgumentException.class)
+            .hasMessageContaining("httpsPortt");
+    }
+
+    @Test
+    void itParsesValidHttpsConfig() throws Exception {
+        Path config = tempDir.resolve("https-config.json");
+        Files.writeString(config, "{\"domain\":\"example.com\",\"password\":\"p\",\"httpsPort\":443}");
+
+        var parsed = CliJsonFiles.readHttpsConfig(config);
+
+        assertThat(parsed.domain()).isEqualTo("example.com");
+        assertThat(parsed.password()).isEqualTo("p");
+        assertThat(parsed.httpsPort()).isEqualTo(443);
+    }
+
+    @Test
+    void itAcceptsArbitraryKeysInObjectMapReads() throws Exception {
+        Path fields = tempDir.resolve("fields.json");
+        Files.writeString(fields, "{\"anything\":1,\"goes\":\"here\"}");
+
+        assertThat(CliJsonFiles.readObjectMap(fields)).containsEntry("goes", "here");
+    }
+}
--- a/TSANet-integration-demo/pom.xml
+++ b/TSANet-integration-demo/pom.xml
@@ -38,6 +38,14 @@
             <artifactId>spring-boot-starter-test</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <!-- Boot 4 moved @WebMvcTest out of spring-boot-test-autoconfigure into
+                 its own module (org.springframework.boot.webmvc.test.autoconfigure);
+                 spring-boot-starter-test no longer brings it transitively. -->
+            <groupId>org.springframework.boot</groupId>
+            <artifactId>spring-boot-webmvc-test</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
 
     <build>
--- a/TSANet-integration-demo/src/main/java/com/tsanet/application/TsaNetApiApplication.java
+++ b/TSANet-integration-demo/src/main/java/com/tsanet/application/TsaNetApiApplication.java
@@ -2,7 +2,7 @@ package com.tsanet.application;
 
 import org.springframework.boot.SpringApplication;
 import org.springframework.boot.autoconfigure.SpringBootApplication;
-import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.jdbc.autoconfigure.DataSourceAutoConfiguration;
 
 // JDBC is only used for optional Connect DB setup (tsanet.demo.setup.enabled); no default spring.datasource.
 @SpringBootApplication(exclude = DataSourceAutoConfiguration.class)
--- a/TSANet-integration-demo/src/main/resources/application.yml
+++ b/TSANet-integration-demo/src/main/resources/application.yml
@@ -1,3 +1,13 @@
+spring:
+  jackson:
+    deserialization:
+      # Jackson 3 flipped FAIL_ON_NULL_FOR_PRIMITIVES to true by default, so a
+      # payload that omits a primitive creator property is now rejected where
+      # Jackson 2 defaulted it (0/false). Restored to the Jackson 2 default so
+      # the Boot 4 upgrade does not change the accepted wire format. Tightening
+      # this is a deliberate contract change, not an upgrade side effect.
+      fail-on-null-for-primitives: false
+
 server:
   port: 9090
 
--- a/TSANet-integration-demo/src/test/java/com/tsanet/application/web/SyncStatusControllerTest.java
+++ b/TSANet-integration-demo/src/test/java/com/tsanet/application/web/SyncStatusControllerTest.java
@@ -13,8 +13,8 @@ import java.time.Instant;
 import java.util.Map;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
-import org.springframework.boot.test.autoconfigure.web.servlet.WebMvcTest;
-import org.springframework.boot.test.mock.mockito.MockBean;
+import org.springframework.boot.webmvc.test.autoconfigure.WebMvcTest;
+import org.springframework.test.context.bean.override.mockito.MockitoBean;
 import org.springframework.test.web.servlet.MockMvc;
 
 @WebMvcTest(SyncStatusController.class)
@@ -22,7 +22,7 @@ class SyncStatusControllerTest {
     @Autowired
     private MockMvc mockMvc;
 
-    @MockBean
+    @MockitoBean
     private ConnectApiPollingCoordinator pollingCoordinator;
 
     @Test
--- a/TSANet-integration-demo/src/test/java/com/tsanet/application/web/TestControllerTest.java
+++ b/TSANet-integration-demo/src/test/java/com/tsanet/application/web/TestControllerTest.java
@@ -14,8 +14,8 @@ import java.util.List;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 import org.springframework.beans.factory.annotation.Autowired;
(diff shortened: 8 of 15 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**

- `CliJsonFiles.java` imports + `catch (IOException | JacksonException)` — traces to "3 Boot 4 breakages / Jackson 3 parse exceptions are unchecked" (contract change 4). Traces.
- `TSANet-integration-app/application.yml` `fail-on-null-for-primitives` block — traces to "Wire format" row. Traces.
- `TSANet-integration-demo/pom.xml` `spring-boot-webmvc-test` dep — traces to `@WebMvcTest` relocation. Traces.
- `TsaNetApiApplication.java` `DataSourceAutoConfiguration` import — traces to "DataSourceAutoConfiguration relocated". Traces.
- `TSANet-integration-demo/application.yml` jackson block — traces to wire format. Traces.
- `SyncStatusControllerTest.java` + `TestControllerTest.java` `@MockBean`→`@MockitoBean`, `@WebMvcTest` import — traces to Boot 4 breakages. Traces.
- `connect-library/pom.xml` groupId move, drop jsr310 + databind-nullable, generator configOptions — traces to "connect-library pom" row. Traces.
- `OAuthTokenGateway.java`, `WebhookPayloadParser.java` import moves — traces to "4 hand-written files". Traces.
- `CollaborationRequestFormRepository.java` custom mapper — traces to the QA gate item fixed in `4bc77c4`. Traces.
- `demo-ui/application.yml` jackson block — traces to wire format. Traces.
- `CollaborationRequestsControllerTest.java` converter setup — traces to the standaloneSetup gap for wire format. Traces.
- `pom.xml` parent bump + generator version + drop nullable property — traces to "Root pom". Traces.

Every hunk traces. One formatting nit: blank line between the two `import org.springframework.http...` lines in `CollaborationRequestsControllerTest.java` (diff lines around the new `JacksonJsonHttpMessageConverter` import) — cosmetic, does not move verdict.

**Correctness:**

- `CliJsonFiles.readObjectMap`: trace a malformed JSON file. Jackson 3 throws `tools.jackson.core.JacksonException` (unchecked). The widened `catch (IOException | JacksonException)` now catches it and rethrows `IllegalArgumentException`. Correct. Note `IOException` is still reachable via `Files.readString` (file not found / read error), so keeping both arms is right, not redundant. Good.
- `CollaborationRequestFormRepository`: the custom `JsonMapper` with `FAIL_ON_NULL_FOR_PRIMITIVES` disabled is the right fix for the out-of-Spring mapper. Trace a cached row omitting `required`: with Jackson 3 default it would throw; disabled, it defaults `boolean` to `false` — matching Jackson 2. The PR's stated-not-hidden note that pre-existing rows load correctly holds.
- The self-declared gap is real and correctly disclosed: **the `application.yml` `fail-on-null-for-primitives: false` setting has no test.** Deleting those lines leaves the suite green while runtime behavior regresses. This is a genuine correctness-coverage gap, but the PR states it and it does not describe wrong behavior in the current diff — it's a follow-up, not a blocker. Draft status means it isn't merging yet regardless.
- The properties-migrator "unproven not clean" note is honest; no correctness claim rests on it.

**Security:** None. No secrets, no injection surface, no auth changes. Mapper reconfiguration only relaxes a null-defaulting feature, not any security-relevant setting.

**Regression risk:**

- Contract change 4 (`JacksonException` now unchecked) is the sharp edge for consumers. This repo's one instance (`CliJsonFiles`) is fixed, but `Connect_Gateway` and `Fin-Intercom_App` are named consumers of `connect-library` and any `catch (IOException)` around a mapper call there will silently stop catching malformed JSON. The PR correctly flags this and lists the sweep as a pre-draft gate. Since this is a published-artifact contract change with no consumer compiled against it yet, the risk is real but the draft gate holds it — this is the reason to stay draft, not to merge.
- The three app `application.yml` mappers are the deserialization path for the running apps; the library repo mapper is separately aligned. Both places covered. The one place NOT covered by a runtime path is the untested yaml, noted above.
- No positional-contract renumbering, no vocabulary change, no new abstraction — those pre-flight lines correctly not triggered.

**Recommendation:** Do not merge — this is correctly a draft; keep it draft until the four "Before this leaves draft" items (consumer SNAPSHOT compile, `catch (IOException)` sweep, version bump, record-vs-flag) are done, and add the missing yaml-setting test before this ships since deleting those lines regresses silently.

---

shawn-tsanet (commented):
## QA review of Migrate to Spring Boot 4.1.1 and Jackson 3

Independently verified, not just read: I rebuilt the full reactor from a fresh checkout of this branch (JDK 21, `mvn test`), regenerated the `origin/main` (`2eed253`) baseline sources in a separate worktree, and ran behavior probes through the built classes with jshell. Receipts inline below. The two findings are follow-up tier; neither would hold the merge. The PR's own "Before this leaves draft" checklist (consumer SNAPSHOT compiles, `catch (IOException)` sweep of both consumers, version bump) remains the actual gate, and nothing here adds to it.

### Minimal-Diff Rule
- Pass. Every hunk traces to one of the five declared change groups (parent/generator bump, generator configOptions + dependency swap, 4 hand-written import moves, 3 Boot 4 breakage fixes, primitive-null wire-format preservation). The `CollaborationRequestsControllerTest` converter hunk traces to keeping the red-probe honest under `standaloneSetup`. No incidental tidying found.

### Correctness
- **Reactor green, counts reproduced:** my build gives connect-library 94, TSANet-integration-demo 6, TSANet-integration-app 30, attachment-receiver 81 (16 skipped), TSANet-demo-ui 15, BUILD SUCCESS. Identical to the PR table.
- **The primitive-null fix re-proven independently:** jshell against the built classes: bare Jackson 3 `ObjectMapper` rejects `{"label":"x"}` into `FormFieldDto` (`MismatchedInputException` on primitive `required`); the repository-style mapper with `FAIL_ON_NULL_FOR_PRIMITIVES` disabled accepts it with `required=false`.
- **The three unaligned hand-built mappers are safe as left:** `WebhookPayloadParser` and `OAuthTokenGateway` use `readTree` plus manual field extraction into all-String/boxed types (no primitive binding exists on those paths), and both catch broad `Exception`, so Jackson 3's unchecked exceptions cannot escape them. `CliJsonFiles` binds only a `Map` and the fully-boxed `NormalizedHttpsAttachmentConfigDto` (`Integer httpsPort`).
- **Generated surface verified:** zero `JsonNullable` references in regenerated sources; generated models are fully boxed (grep for primitive fields: none), so the strict default cannot fire on the generated deserialization path. `useJackson3`/`useSpringBoot4` are real `JavaClientCodegen` options in the 7.21.0 jar, whose own docstring states "Incompatible with 'openApiNullable'" and that resttemplate requires `useSpringBoot4=true`, matching the PR body's claims.
- **The untested yaml key is at least real:** `spring.jackson.deserialization` exists in `spring-boot-jackson-4.1.1.jar` configuration metadata, and the probes confirm the Jackson 3 enum constant. The end-to-end yaml-to-mapper binding test stays a known gap as the PR states; a typo'd feature name would fail loudly at startup (enum binding error), so the silent-failure mode is narrower than the PR body implies.

### Findings (both non-blocking)
1. **`CliJsonFiles` lost its config-typo detection, silently.** Jackson 3 also flipped `FAIL_ON_UNKNOWN_PROPERTIES` to lenient. Probe receipt: bare Jackson 3 mapper tolerates `{"label":"x","required":true,"bogusKey":1}` into `FormFieldDto`. Under Jackson 2, `new ObjectMapper()` rejected unknown keys, so a misspelled key in a CLI JSON config file used to error loudly and now silently yields a config with that field unset. Suggested follow-up: `.enable(FAIL_ON_UNKNOWN_PROPERTIES)` on the `CliJsonFiles` mapper (or accept and note it). Not blocking: it only affects operator-authored local config files.
2. **Latent contract change 5 for the PR body's consumer list:** the six formerly `JsonNullable` fields on `CaseApprovalUpdateDTO` and `CollaborationRequestSubmitterUpdateDTO` now serialize as explicit `null` when unset. Probe receipt: `new CaseApprovalUpdateDTO().caseNumber("C-1")` through a bare Jackson 3 mapper produces `{"caseNumber":"C-1","engineerName":null,"engineerPhone":null,...}`; the old shape initialized these to `JsonNullable.undefined()` (omitted). Unreachable today: grep shows zero call sites constructing either DTO in any module. But both classes ship in the published jar, so a consumer wiring up the update endpoints later would send explicit nulls where a Jackson 2 consumer omitted the keys, and an update endpoint could read explicit null as "clear this field". Worth adding to the Contract changes section since `tsanetgit/Connect_Gateway` and `tsanetgit/Fin-Intercom_App` can import generated models directly.

### Security
- No concerns. No secrets in the diff, no new external-input paths, no logging changes. The widened `catch (IOException | JacksonException)` in `CliJsonFiles` preserves the existing message-wrapping behavior.

### Regressions
- Swept the whole reactor at this head: zero non-annotation `com.fasterxml.jackson` imports, zero `@MockBean`/`@SpyBean`, zero old-package `WebMvcTest` (new-package usage is only the two migrated demo tests, and the new `spring-boot-webmvc-test` artifact is scoped there), zero `findAndRegisterModules`, zero post-construction `.configure(` on mappers, zero string-form `spring.autoconfigure.exclude` references in yml/properties.
- attachment-receiver legitimately needs no yaml: grep shows zero Jackson usage in that module.
- Remaining `catch (IOException)` sites (`EnvironmentService`, `CredentialsStore`, `AttachmentsController`, receiver tests) all wrap file/stream IO, not mapper calls, so the unchecked-exception change cannot silently bypass any of them.

### Operational
- No new env vars, schema, or runtime dependency beyond the pom changes themselves. The consumer-side actions (SNAPSHOT compile, IOException sweep, major version bump) are already tracked in the PR's own draft checklist.

### Nits
- None worth noting.

shawn-tsanet (commented):
## Disposition of QA finding 1 (unknown-properties regression in `CliJsonFiles`)

**Fixed in e035284.** `FAIL_ON_UNKNOWN_PROPERTIES` re-enabled on the `CliJsonFiles` mapper via `JsonMapper.builder()`, the same builder pattern as `CollaborationRequestFormRepository`.

**Probe, red then green on the same class of input:**
- Pre-fix (QA probe, this session): a bare Jackson 3 mapper silently tolerated a bogus key during typed binding.
- Post-fix, through the rebuilt shipped class via reflection: `readHttpsConfig` on `{"domain":"example.com","httpsPortt":443}` now throws `IllegalArgumentException: Failed to read HTTPS config file ...: Unrecognized property "httpsPortt" ... (5 known properties: "domain", "password", "expiration", "httpsPath", "httpsPort")`. A valid config still parses (`domain=example.com, httpsPort=443` bound), and `readObjectMap` is unaffected: the feature only applies to typed binding, not `Map` reads.

**Receiving handler and user-visible outcome:** `UnrecognizedPropertyException` extends `JacksonException`, so it lands in the `catch (IOException | JacksonException)` this PR already widened, and every CLI executor catches the resulting `IllegalArgumentException` and prints it via `EntityPrinter.error(...)`. The operator sees the error line above instead of a silently unset field.

**Bug-class sweep, not just the caught instance:** five hand-built mappers in the reactor. `CliJsonFiles` is fixed here. `WebhookPayloadParser` and `OAuthTokenGateway` use `readTree` plus manual extraction, so the feature does not apply to them. `CollaborationRequestFormRepository` stays lenient deliberately: it reads back cache rows written by earlier shapes, where strict unknown-key handling would break reads. The three app yamls and the demo-ui test converter are web-path binding, where Boot's auto-configured mapper was already lenient on Boot 3, so no behavior changed there in the first place.

**Gate:** `qa-review.py --local --base 4bc77c4` on this commit: LOOKS GOOD, exit 0. Module build green, 30/30 tests in TSANet-integration-app.

**Deliberately not included:** a committed regression test (it would be a new test file, so it needs an explicit scope go-ahead). The suite was green both before and after this fix, so it is blind to the behavior; a ~10-line `CliJsonFilesTest` asserting the rejection and the valid-config parse would lock it in. Flagged to Shawn as an open scope decision.

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass. The single hunk (imports for `DeserializationFeature`/`JsonMapper` + builder-based mapper with `FAIL_ON_UNKNOWN_PROPERTIES` enabled) traces directly to prior finding 1, dispositioned in `e035284`.

**Correctness:**
Trace a config file with a typo'd key through `readHttpsConfig`: the new `JsonMapper.builder().enable(FAIL_ON_UNKNOWN_PROPERTIES).build()` mapper rejects the unknown key with `UnrecognizedPropertyException` (extends `JacksonException`), lands in the existing widened `catch (IOException | JacksonException)`, rethrown as `IllegalArgumentException`, surfaced via `EntityPrinter.error`. This restores the Boot 3 loud-failure behavior. The comment correctly notes `Map` reads are unaffected (feature applies only to typed binding). The disposition's red/green probe on `httpsPortt` confirms it. Correct.

**Security:** None. Making the mapper stricter on unknown keys narrows silent-config-drift, no new surface.

**Regression risk:** None new. Enabling `FAIL_ON_UNKNOWN_PROPERTIES` only affects typed binding in this one CLI mapper — `readObjectMap`'s `Map` path is untouched, and the deliberately-lenient `CollaborationRequestFormRepository` (cache-row reads) is left as-is per the disposition's bug-class sweep.

**Carried forward:**
- CliJsonFiles unknown-properties regression — RESOLVED (e035284)
- JsonNullable fields serialize explicit null (contract change 5) — FOLLOW-UP (add to Contract changes section; unreachable today, grep shows zero call sites)
- application.yml primitive-null setting has no test — FOLLOW-UP (self-disclosed; draft gate holds)
- committed CliJsonFiles regression test — FOLLOW-UP (scope decision flagged to Shawn)

**Recommendation:** Merge-ready on this delta, but keep it draft until the four "Before this leaves draft" items are done.

---

shawn-tsanet (commented):
Closing the open item from the finding-1 disposition: the regression test is in as 61ab153 (`CliJsonFilesTest`, 3 cases: typo'd key rejected naming the key, valid config binds all fields, `readObjectMap` still accepts arbitrary keys).

Red-probe receipt: with `CliJsonFiles` reverted to its pre-fix `4bc77c4` version, exactly `itRejectsUnknownKeysInHttpsConfig` fails (Tests run 3, Failures 1, Errors 0); with the fix in place the module suite is 33/33 green. Gate on the commit: LOOKS GOOD, exit 0.

PR #68, opened 2026-09-02
Add GCS storage adapter and a concrete StorageFactory

shawn-tsanet:
## What

Adds the Google Cloud Storage delivery adapter and the first production `StorageFactory`, for Option B member file transfer (action item from the 2026-09-02 design call). S3 and Azure Files adapters already existed against the `AttachmentStorage` SPI; GCS did not, and no production factory existed, so no adapter was reachable from a `TenantConfig`.

## Scope (confirmed with Shawn)

Multi-file, one PR for both the adapter and the factory. New files only, plus one pom change. No existing source edited.

- `storage/gcs/GcsAttachmentStorage` + `GcsBucket` seam + `SdkGcsBucket`
- `verify/RegisteringStorageFactory` — maps `storageBackend` (`s3`/`azure`/`gcs`) + `storageProperties` to a constructed adapter
- tests: contract-bound offline test, GCS-edge test, env-gated live test, factory test
- `attachment-receiver/pom.xml`: `com.google.cloud:libraries-bom` 26.83.0 (resolves google-cloud-storage 2.68.0), alongside the existing AWS and Azure BOMs

## Design notes

**GCS mirrors the SPI contract, not the S3 structure.** First-buffer-decides single-shot vs resumable, but with GCS semantics:
- A resumable upload is invisible until `finish()`. On any mid-stream failure the upload is **abandoned, never finished** — finishing a partial upload would finalize a truncated visible object. There is deliberately no try-with-resources on the channel (`WriteChannel.close()` finalizes), and no abort round-trip (GCS has none; unfinalized sessions expire).
- Ambiguous-finish (finish fails client-side after committing server-side) mirrors the S3 ambiguous-complete policy: object visible at the exact byte count written is reported as the success it was.
- `exists` is cleaner than S3: a missing GCS object is a null get, not a 404 to disambiguate from a permission error.

**Factory credential schema** is documented in the `RegisteringStorageFactory` javadoc (the visible decision). Credentials are read from `storageProperties` and never logged; `TenantConfig.toString()` redacts both property maps and the password (verified). The `credentialsJson` parse-failure path deliberately does not echo the parser message, so a malformed key cannot leak into a logged config error.

## Blast radius

- **Single-place decision:** `grep` for `storageBackend()` / adapter construction in `src/main` shows `RegisteringStorageFactory` is the only production place that decides backend → adapter; `GoLiveVerifier` and `EncryptedFileTenantConfigStore` only read/persist the id string. Every other adapter construction is in each adapter's own tests.
- **No shared-symbol or positional-contract change.** The SPI, `StorageFactory`, and `TenantConfig` are untouched. New public API (`GcsAttachmentStorage.forBucket`, `RegisteringStorageFactory`) has no callers outside the new files (grep confirmed).
- **No schema/stored-shape change.** The factory reads new keys from the existing free-form `storageProperties` map; pre-existing configs simply lack these keys and get a clear `require()` error.

## Tests

`mvn -pl attachment-receiver -am test`: **113 run, 0 failures, 23 skipped** (the skipped are the env-gated S3/Azure/GCS live tests). 25 tests are new: 17 in `GcsAttachmentStorageTest` (the 7-test shared contract bound to an in-memory fake + 10 GCS edges), 8 in `RegisteringStorageFactoryTest`. The GCS live contract test is env-gated on `GCS_CONTRACT_TEST_BUCKET` and skips without credentials, mirroring S3/Azure.

## Known parity caveat (documented, matches the S3 twin)

The ambiguous-finish probe checks visible size equals bytes written; on a same-name overwrite of a same-size object where finish never actually committed, it could report success while stale content persists. Same class as the acknowledged concurrent-writer caveat and identical to the S3 adapter's policy; CRC32C disambiguation is a possible future hardening, not done here.

- `google-auth-library-oauth2-http` (used via `GoogleCredentials`, present transitively) could be declared explicitly if `dependency:analyze` is ever enforced.

## Review

This is a `Connect_SDK` change; its own CI plus a QA-persona pass are the review gate (per the tsanet-ops Coder/QA split). Advisor was consulted at start and pre-done; the abandon-vs-finish invariant and no-partial-visibility guarantee were the adversarial focus.

### DIFF
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -43,6 +43,13 @@
                 <type>pom</type>
                 <scope>import</scope>
             </dependency>
+            <dependency>
+                <groupId>com.google.cloud</groupId>
+                <artifactId>libraries-bom</artifactId>
+                <version>26.83.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
 
@@ -54,6 +61,10 @@
         <dependency>
             <groupId>com.azure</groupId>
             <artifactId>azure-storage-file-share</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>com.google.cloud</groupId>
+            <artifactId>google-cloud-storage</artifactId>
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsAttachmentStorage.java
@@ -1,0 +1,255 @@
+package com.tsanet.receiver.storage.gcs;
+
+import com.tsanet.receiver.storage.AttachmentStorage;
+import com.tsanet.receiver.storage.AttachmentStorageException;
+import com.tsanet.receiver.storage.IncomingAttachment;
+import com.tsanet.receiver.storage.StoredAttachment;
+import com.google.cloud.storage.Storage;
+import com.google.cloud.storage.StorageException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * {@link AttachmentStorage} on Google Cloud Storage. The client is injected: project,
+ * region and credentials are wiring concerns, not this class's.
+ *
+ * <p>Object keys are {@code [prefix/]caseNumber/fileName}. Per the SPI, {@code fileName} is
+ * an opaque token: GCS object names have no traversal semantics, so a name containing
+ * {@code /} or {@code ..} merely creates deeper name segments. {@code exists} builds keys
+ * through the same method, so store and check can never disagree.
+ *
+ * <p>Streaming: bytes are read in {@value #BUFFER_SIZE}-byte buffers. A file that ends
+ * inside the first buffer is written with a single atomic create; anything larger becomes a
+ * resumable upload. A GCS object is not visible until its resumable upload is finished, so
+ * the SPI's no-partial-visibility rule holds structurally, and the streamed body is never
+ * the caller's stream (each buffer is a fresh copy), so the stream is never closed by this
+ * class.
+ *
+ * <p>Two GCS facts shape the failure paths, and neither has an S3 or Azure twin:
+ * <ul>
+ *   <li><b>No abort call.</b> A resumable upload that is never finished produces no visible
+ *       object and its session expires server-side. On any mid-stream failure the upload is
+ *       {@code abandon()}ed, not finished — finishing a partial upload would finalize a
+ *       truncated object. There is deliberately no cleanup round-trip that could itself
+ *       fail, unlike S3's {@code AbortMultipartUpload}.</li>
+ *   <li><b>Ambiguous finish.</b> A {@code finish} that fails client-side may have committed
+ *       server-side. When the object is then visible at exactly the byte count written, the
+ *       store is treated as committed and reported as success, mirroring the S3 adapter's
+ *       ambiguous-complete policy. (Like both twins, this cannot distinguish a concurrent
+ *       same-name writer's object.)</li>
+ * </ul>
+ *
+ * <p>Required IAM on the bucket/prefix: {@code storage.objects.create},
+ * {@code storage.objects.get} (read metadata for {@code exists} and the ambiguous-finish
+ * probe, and read bytes for the verify sentinel), and {@code storage.objects.delete} (the
+ * verify sentinel). Unlike S3's HEAD, a missing GCS object reads as a null get rather than
+ * a 404 that must be told apart from a permission error, so {@code exists} has no
+ * status-ambiguity to guard.
+ */
+public final class GcsAttachmentStorage implements AttachmentStorage {
+
+    /** Stream read granularity; a 256 KiB multiple, so it is also a valid resumable chunk size. */
+    static final int BUFFER_SIZE = 5 * 1024 * 1024;
+
+    private final GcsBucket bucket;
+    private final String bucketName;
+    private final String prefix;
+
+    /** Production entry point: wraps the injected client in the logic-free SDK seam. */
+    public static GcsAttachmentStorage forBucket(Storage storage, String bucketName, String keyPrefix) {
+        return new GcsAttachmentStorage(new SdkGcsBucket(storage, bucketName), bucketName, keyPrefix);
+    }
+
+    GcsAttachmentStorage(GcsBucket bucket, String bucketName, String keyPrefix) {
+        this.bucket = Objects.requireNonNull(bucket, "bucket");
+        this.bucketName = Objects.requireNonNull(bucketName, "bucketName");
+        if (bucketName.isBlank()) {
+            throw new IllegalArgumentException("bucketName must not be blank");
+        }
+        this.prefix = normalize(keyPrefix);
+    }
+
+    @Override
+    public StoredAttachment store(IncomingAttachment attachment, InputStream content)
+            throws AttachmentStorageException {
+        String key = key(attachment.caseNumber(), attachment.fileName());
+        byte[] first = new byte[BUFFER_SIZE];
+        int firstLength = fill(content, first, key);
+        if (firstLength < BUFFER_SIZE) {
+            // EOF inside the first buffer: one atomic create, no partial-visibility window.
+            try {
+                bucket.create(key, attachment.contentType(), Arrays.copyOf(first, firstLength));
+            } catch (RuntimeException e) {
+                throw wrap("create", key, e);
+            }
+            return new StoredAttachment(key, firstLength);
+        }
+        return resumable(attachment, content, key, first);
+    }
+
+    private StoredAttachment resumable(IncomingAttachment attachment, InputStream content,
+                                       String key, byte[] firstBuffer)
+            throws AttachmentStorageException {
+        GcsBucket.Upload upload;
+        try {
+            upload = bucket.startResumable(key, attachment.contentType());
+        } catch (RuntimeException e) {
+            throw wrap("resumable start", key, e);
+        }
+
+        long total = 0;
+        try {
+            byte[] buffer = firstBuffer;
+            int length = firstBuffer.length;
+            int chunk = 0;
+            boolean last = false;
+            while (!last) {
+                if (chunk > 0) {
+                    buffer = new byte[BUFFER_SIZE];
+                    length = fill(content, buffer, key);
+                    if (length == 0) {
+                        break; // EOF landed exactly on the previous chunk's boundary.
+                    }
+                }
+                chunk++;
+                upload.write(buffer, length);
+                total += length;
+                last = length < BUFFER_SIZE;
+            }
+        } catch (AttachmentStorageException streamFailure) {
+            // Already carries the key and phase; re-wrapping would just nest the message.
+            upload.abandon();
+            throw streamFailure;
+        } catch (RuntimeException uploadFailure) {
+            upload.abandon();
+            throw wrap("store", key, uploadFailure);
+        }
+
+        try {
+            upload.finish();
+        } catch (RuntimeException finishFailure) {
+            return resolveAmbiguousFinish(key, total, finishFailure);
+        }
+        return new StoredAttachment(key, total);
+    }
+
+    /**
+     * A failed finish is ambiguous: GCS may have committed server-side. The object visible
+     * at exactly the byte count written means exactly that — report the commit as the
+     * success it was. Every other shape throws.
+     */
+    private StoredAttachment resolveAmbiguousFinish(String key, long total,
+                                                    RuntimeException finishFailure)
+            throws AttachmentStorageException {
+        try {
+            if (bucket.sizeOrAbsent(key) == total) {
+                return new StoredAttachment(key, total);
+            }
+        } catch (RuntimeException probeFailure) {
+            finishFailure.addSuppressed(probeFailure);
+        }
+        throw wrap("finish", key, finishFailure);
+    }
+
+    @Override
+    public boolean exists(String caseNumber, String fileName) throws AttachmentStorageException {
+        String key = key(caseNumber, fileName);
+        try {
+            return bucket.sizeOrAbsent(key) >= 0;
+        } catch (RuntimeException e) {
+            throw wrap("exists check", key, e);
+        }
+    }
+
+    @Override
+    public void verifyAccess() throws AttachmentStorageException {
+        String sentinel = (prefix.isEmpty() ? "" : prefix + "/") + ".verify-" + UUID.randomUUID();
+        byte[] probe = "attachment-receiver go-live probe".getBytes();
+        String stage = "write";
+        boolean written = false;
+        try {
+            bucket.create(sentinel, "application/octet-stream", probe);
+            written = true;
+            stage = "read";
+            if (!Arrays.equals(probe, bucket.download(sentinel))) {
+                throw new AttachmentStorageException(
+                        "verify read returned different content than was written to bucket '"
+                                + bucketName + "'");
+            }
+            stage = "delete";
+            bucket.delete(sentinel);
+        } catch (Exception e) {
+            AttachmentStorageException failure = e instanceof AttachmentStorageException ase
+                    ? ase
+                    : new AttachmentStorageException(classify(stage, e), e);
+            // A sentinel from a failed probe must not linger; best-effort, the probe's own
+            // failure stays primary. Pointless only when delete itself just failed.
+            if (written && !"delete".equals(stage)) {
+                try {
+                    bucket.delete(sentinel);
+                } catch (Exception cleanupFailure) {
+                    failure.addSuppressed(cleanupFailure);
+                }
+            }
+            throw failure;
+        }
+    }
+
+    /** Wrong-credential, wrong-target, and no-permission must read differently. */
+    private String classify(String stage, Exception e) {
+        if (e instanceof StorageException se) {
+            int code = se.getCode();
+            if (code == 401) {
+                return "wrong credential: GCS rejected the identity (verify " + stage + ")";
+            }
+            if (code == 403) {
+                return "no permission: " + stage + " denied on bucket '" + bucketName + "'";
+            }
+            if (code == 404) {
+                return "wrong target: bucket '" + bucketName + "' does not exist (verify " + stage + ")";
+            }
+            return "verify " + stage + " failed on bucket '" + bucketName + "': "
+                    + se.getMessage() + " (HTTP " + code + ")";
+        }
+        return "connectivity: cannot reach GCS (verify " + stage + "): " + e.getMessage();
+    }
+
+    private int fill(InputStream content, byte[] buffer, String key)
+            throws AttachmentStorageException {
+        try {
+            // readNBytes loops over short reads; < buffer.length only ever means EOF.
+            return content.readNBytes(buffer, 0, buffer.length);
+        } catch (IOException e) {
+            throw new AttachmentStorageException(
+                    "stream failed mid-read for " + key + "; upload abandoned", e);
+        }
+    }
+
+    private String key(String caseNumber, String fileName) {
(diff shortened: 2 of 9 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
- `pom.xml` +libraries-bom import and +google-cloud-storage dep — traces to "one pom change" / GCS SDK need. ✓
- `GcsAttachmentStorage.java` (new) — the adapter, core of the PR. ✓
- `GcsBucket.java` (new) — the seam the description names. ✓
- `SdkGcsBucket.java` (new) — logic-free SDK impl, named in scope. ✓
- `RegisteringStorageFactory.java` (new) — the factory, second half of the PR. ✓
- `GcsAttachmentStorageTest.java` / `InMemoryGcsBucket.java` / `GcsAttachmentStorageLiveTest.java` / `RegisteringStorageFactoryTest.java` (new) — the four test artifacts named in "Tests". ✓

Every hunk traces. No existing source edited, consistent with the "new files only" claim. No incidental tidying.

**Correctness:**
Traced a `BUFFER_SIZE + 4096` input through `store` → resumable path:
- `fill` reads first buffer to `BUFFER_SIZE` (firstLength == BUFFER_SIZE), so `store` takes `resumable()`. ✓
- Loop: chunk 0 writes `firstBuffer` (length BUFFER_SIZE), `last = false`. chunk 1 does `fill` → 4096 bytes, `last = true`. `total = BUFFER_SIZE + 4096`. ✓ Matches `fileLargerThanOneBufferGoesResumableAndCountsEveryByte`.

Edge — EOF exactly on boundary (`BUFFER_SIZE` bytes): first buffer fills to BUFFER_SIZE → resumable. chunk 0 writes BUFFER_SIZE, `last=false`. chunk 1 `fill` returns 0 → `break`. `total = BUFFER_SIZE`. finish commits. ✓ Matches `eofExactlyOnBufferBoundaryStillStoresEveryByte`.

One real observation, non-blocking: `resolveAmbiguousFinish` reports success when `sizeOrAbsent(key) == total`. The PR itself documents the same-name-overwrite-same-size false-positive as a known caveat identical to the S3 twin. This is disclosed and matches an accepted policy, so not a finding to hold merge.

`fill` correctly treats a short `readNBytes` as EOF only (readNBytes loops over short reads) — the comment is true of the JDK contract.

`verifyAccess` cleanup path: only best-effort deletes when `written && stage != "delete"`, and suppresses cleanup failure under the primary. Correct — no double-delete after a delete-stage failure.

**Security:**
- No secrets in the diff. Test keys are obvious fakes (`AKIAEXAMPLE`, `secret`, base64 `Zm9vYmFyYmF6`).
- `credentialsJson` parse-failure path (RegisteringStorageFactory line ~118) deliberately does not echo the parser message into the thrown message, keeping key material out of logged config errors — cause chain retains library detail. Matches the PR claim.
- No log statements in the adapter or factory at all, so no runtime credential leakage path.
- `wrap()` includes `gs://bucket/key` in messages — key is `caseNumber/fileName`, not credential material. Fine.
- Keys built from `caseNumber + "/" + fileName`; GCS object names have no traversal semantics, so a `../` filename is just a name segment. Correctly reasoned; no path-injection risk.

**Regression risk:**
- No existing source modified — SPI, `StorageFactory`, `TenantConfig`, `S3AttachmentStorage`, `AzureFilesAttachmentStorage` all untouched. New public API (`GcsAttachmentStorage.forBucket`, `RegisteringStorageFactory`) has no pre-existing callers per the PR's grep claim; I cannot independently re-run the grep from a diff, but the "new files only, +976/−0" stat is consistent with it.
- No positional/tuple/regex-group renumbering. The `switch` on `storageBackend` adds `gcs` as a new vocabulary member; the old dispatch site did not exist (no prior production factory), so there is no stale term to sweep.
- New dep is BOM-managed, isolated to `attachment-receiver`.

**Recommendation:** Merge — the diff is fully traced, the failure invariants are exercised by the offline tests, and the one false-positive edge is a disclosed, S3-parity caveat rather than a new defect.

Note: this is a `Connect_SDK`/Connect-API repo change, outside `tsanet-ops`; I reviewed the diff on its merits but have not run the build or grep myself, so the "113 run, 0 failures" and grep-of-callers claims rest on the PR body, not a check I ran this session.

---

shawn-tsanet (commented):
## QA review of Add GCS storage adapter and a concrete StorageFactory

**Verdict:** ❌ Needs changes

Second-session QA pass (separate from the auto-QA comment above, which reviewed the diff text only). This one ran the build, the dependency-tree diff, and three probes against the branch at `ab601ba`. Every claim below carries the check it rests on.

### Minimal-Diff Rule
- Pass. Nine files, all new except the one `pom.xml` hunk (BOM import + one managed dependency). Every hunk traces to the PR's stated scope. `git status` in the clone after all probes: only pre-existing untracked files, so nothing here touched the branch.

### Correctness
- **Build receipt:** `mvn -pl attachment-receiver -am test` on JDK 21.0.11 at `ab601ba`: 113 run, 0 failures, 23 skipped (8 S3 live + 8 Azure live + 7 GCS live). Matches the PR body exactly.
- **Vacuous assertion in `GcsAttachmentStorageTest.streamFailureMidResumableAbandonsAndLeavesNothingVisible`** (`attachment-receiver/src/test/java/com/tsanet/receiver/storage/gcs/GcsAttachmentStorageTest.java:96`). The test stores under case `01234567`, so the key is `01234567/truncated.bin`, but the assertion reads `gcs.contentOf("truncated.bin")`, which is `null` in every possible state. Probe: I temporarily made the fake's `abandon()` commit the staged bytes (the exact bug this assertion claims to guard against). The original assertion still passed. Re-pointed at `01234567/truncated.bin`, it failed as it should, and with the real fake restored it passed. The `abandonCount == 1` assertion on the previous line is what actually holds the abandon-not-finish invariant today. The S3 twin at `S3AttachmentStorageTest.java:126` uses the full key, so this is a transcription slip, not a copied pattern. One-string fix.
- **Ambiguous-finish stale-overwrite false positive** (`GcsAttachmentStorage.resolveAmbiguousFinish`): a same-name, same-size pre-existing object plus a finish that never committed reports success over stale bytes. The PR body discloses this under "Known parity caveat", and it is real. Note the S3 twin has one more disambiguator GCS cannot have (abort answering `NoSuchUpload`), so the GCS version is strictly weaker. Disclosed and accepted, so not held against the merge; the CRC32C or generation-check hardening belongs in a follow-up issue so it does not evaporate.
- Traced `BUFFER_SIZE + 4096`, exactly `BUFFER_SIZE`, and sub-buffer inputs through `store`/`resumable`; the single-shot vs resumable split and the `fill() == 0` boundary break are right, and no path reaches `finish()` after a failure.
- Nit-level: `fill()`'s message says "upload abandoned" even on the single-shot path, where nothing was started.

### Security
- **Blocking: a misconfigured `sasUrl` leaks its value, including a SAS signature or an account key, through the exception cause chain** (`RegisteringStorageFactory.azure`, `attachment-receiver/src/main/java/com/tsanet/receiver/verify/RegisteringStorageFactory.java:89`). `builder.endpoint(sasUrl)` is not guarded. Probe (six inputs through `factory.create` with backend `azure`, marker string in the value, full `getCause()` chain inspected):
  - `not a url at all sig=MARKER` → `IllegalArgumentException: ...endpoint url is malformed` caused by `MalformedURLException: no protocol: not a url at all sig=MARKER`. **Leaked.**
  - `ht!tp://bad host/share?sig=MARKER` → same shape. **Leaked.**
  - A full Azure connection string pasted into the `sasUrl` key (the realistic operator mistake) → `MalformedURLException: no protocol: DefaultEndpointsProtocol=...;AccountKey=MARKER;...`. **Leaked, and that is the account key.**
  - Well-formed SAS URL with share → builds. URL with no share path → `NullPointerException: 'shareName' cannot be null` (clean, but see next bullet). URL to a file/directory path → clean message.
  
  Two things are wrong here. The value reaches logs the moment anyone prints the exception with its causes (standard logger behaviour), which contradicts the PR's "credentials ... are never logged here" and the factory javadoc's "a required key that is missing or blank is a config error". And the exceptions escaping are `IllegalArgumentException` / `NullPointerException`, not the `AttachmentStorageException` the `StorageFactory` contract declares (`StorageFactory.java:18`), so the azure branch also breaks the factory's own contract on malformed input. `GoLiveVerifier.failureMessage` happens to render only the top-level message, so the verification report is clean today; that is luck, not a control. Fix shape: wrap `endpoint(...)` and `buildClient()` in the same pattern the `gcs` branch already uses for `credentialsJson`, a value-free `AttachmentStorageException` message and no value-carrying cause. Then add the negative test naming the wrong-but-well-formed input it rejects.
- **`credentialsJson` cause chain verified clean.** The PR claims the cause keeps "value-free" library detail without evidence. Probe: four malformed keys (valid JSON missing fields, truncated JSON, non-JSON, near-complete key), marker in the `private_key` value, full cause chain inspected. No marker in any message; the library says only which fields it expected, and gson's `MalformedJsonException` reports a position, not content. So the GCS branch does what the comment says. The azure branch does not.
- `TenantConfig.toString()` redaction claim: verified against `TenantConfig.java` on the branch, prints key sets and `<redacted>` only. The diff does not touch it.
- No secrets in the diff; test credentials are obvious fakes. No log statements in any new main-source file.

### Regressions
- **Dependency-tree diff, main vs branch** (`mvn dependency:tree` on both, parsed by group:artifact): 86 pre-existing artifacts, **0 version changes**; 81 new artifacts, every one under `com.google.cloud:google-cloud-storage:2.68.0`. So the second BOM import shifts nothing the S3 or Azure adapters already resolve. One thing to know: it brings in `com.fasterxml.jackson.dataformat:jackson-dataformat-xml:2.21.5` and woodstox, i.e. Jackson 2 on the classpath next to the Jackson 3 migration from #67. Different package namespaces, they coexist; noting it because the security-audit workflow will start seeing Jackson 2 CVEs against this module.
- `RegisteringStorageFactory` is the only production site that turns `storageBackend()` into an adapter (grep of `src/main` on the branch: `GoLiveVerifier` reads the id for a message, `EncryptedFileTenantConfigStore` persists it). No stale vocabulary to sweep, no positional contracts changed.
- No Spring or `main` entry point exists in this module yet (grep for `@Configuration`, `@Bean`, `SpringApplication`, `main(`: none), so there is no wiring site that should have been updated to reference the new factory.

### Operational
- `RegisteringStorageFactory` builds a fresh `S3Client` / GCS `Storage` per `create()` and nothing ever closes them; `AttachmentStorage` has no `close()`. `GoLiveVerifier.storageCheck` calls `create()` once per verification, so this is bounded by onboarding frequency today. That is a property of the `StorageFactory` shape this PR did not change, so a question for the eventual hosting layer rather than this diff: whoever calls the factory per push, not per onboarding, will leak connection pools.
- CI: both `audit` and `verify` checks green on the head commit (`gh pr checks 68`).
- No env, schema, or LaunchAgent implications for tsanet-ops.

### What unblocks this
1. Guard the azure branch's `endpoint`/`buildClient` so malformed `sasUrl` (and a missing share name) become a value-free `AttachmentStorageException`, with a negative test.
2. Fix the key in the abandon test's `contentOf` assertion.
3. Optional in this PR, otherwise a follow-up issue: the ambiguous-finish hardening.

The advisor was consulted on the diff and independently raised the `sasUrl` path as a possibility; the probe above is what turned it into a finding.

shawn-tsanet:
Follow-up filed for the ambiguous-finish hardening noted in the QA review: `tsanetgit/Connect_SDK#69`. It covers both the GCS probe from this PR and the S3 twin, and is listed under epic #46. Not a merge condition for this PR.

shawn-tsanet:
## Response to QA (second-session review)

Fixed in 790273c. Per-finding disposition, each with the check it rests on.

### Blocking 1, azure `sasUrl` / `connectionString` value leak: FIXED
`RegisteringStorageFactory.azure()`. Both branches now wrap client construction and throw a value-free `AttachmentStorageException` with **no cause**: here the cause was the leak. (The gcs branch keeps its cause because that one was probed value-free.)

Receipt: four new negative tests in `RegisteringStorageFactoryTest`, each asserting that a `MARKER` standing in for a secret appears nowhere in the full `getCause()` chain, and that the escaping type is the contract's `AttachmentStorageException`, so the former `IllegalArgumentException` / `NullPointerException` escapes now fail `assertThrows`:
- `azureMalformedSasUrlNeverEchoesItsValue`
- `azureConnectionStringPastedIntoSasUrlNeverEchoesTheAccountKey`, the realistic operator mistake you demonstrated
- `azureSasUrlWithoutShareNameIsAContractErrorNotAnNpe`
- `azureMalformedConnectionStringNeverEchoesItsValue`, same class on the branch you did not probe; it also threw with the value before the guard

All four pass. `RegisteringStorageFactoryTest`: 8 to 12.

### Blocking 2, vacuous `contentOf("truncated.bin")` assertion: FIXED
`GcsAttachmentStorageTest` now reads `contentOf("01234567/truncated.bin")`. Probed your way: mutated `InMemoryGcsBucket.abandon()` to commit the staged bytes, and the corrected assertion **fails** (`expected: <null> but was: <[120, 120, ...]>`); restored the committed fake (`git diff` on it empty) and it passes. It now genuinely holds the abandon-not-finish invariant alongside `abandonCount == 1`.

### Nit, `fill()` message: FIXED
"upload abandoned" is now "nothing stored", accurate on the single-shot path too.

### Ambiguous-finish stale-overwrite false positive: ACCEPTED, follow-up proposed
Disclosed parity caveat; agreed the GCS version is strictly weaker than S3's, which has the `NoSuchUpload` disambiguator. Not changed here. Proposed as a follow-up issue (CRC32C or generation check); filing is Shawn's call and has been flagged to him.

### Operational notes: ACKNOWLEDGED, outside this PR
- Per-`create()` clients are never closed: a property of the `StorageFactory` / `AttachmentStorage` shape this PR did not change, bounded by onboarding frequency today. A question for whichever hosting layer calls the factory per push rather than per onboarding.
- Jackson 2 (`jackson-dataformat-xml` 2.21.5 plus woodstox) is now on the classpath via `google-cloud-storage`, alongside the Jackson 3 migration from #67. They coexist in different namespaces; noting that `audit` will start seeing Jackson 2 CVEs against this module.

**Full suite at the new head:** `mvn -pl attachment-receiver -am test` gives 117 run, 0 failures, 23 skipped (8 S3, 8 Azure, 7 GCS live tests, all env-gated). Ready for re-review.

shawn-tsanet (commented):
## QA re-review of Add GCS storage adapter and a concrete StorageFactory

Re-reviewed at `790273c`. The two blocking items from the first pass are fixed and both fixes are proven by the same probes that found them, re-run against the new head. Nothing remaining would change the merge decision. The notes are follow-ups, not a request for another push.

### Minimal-Diff Rule
- Pass. The fix commit touches four files, 76 insertions and 5 deletions, and every hunk traces to a finding in the prior review: the azure guard, the assertion key, the four negative tests, and the `fill()` message nit. No incidental changes.

### Correctness
- **Build receipt:** clean `mvn -pl attachment-receiver -am clean test` on JDK 21.0.11 at `790273c`: 117 run, 0 failures, 23 skipped. Matches the Coder's number. (`RegisteringStorageFactoryTest` 8 to 12; `GcsAttachmentStorageTest` stays 17.)
- **Abandon assertion now holds the invariant.** Re-ran the finalize-on-abandon mutation against the corrected `contentOf("01234567/truncated.bin")`: the test fails with `expected: <null> but was: <[120, ...]>`, and passes with the fake restored. Before the fix the same mutation passed. Confirmed.
- The `fill()` message change to "nothing stored" is accurate on both paths: single-shot never started an upload, resumable abandons without finalizing.

### Security
- **Azure value leak closed on both branches.** Re-ran the leak probe with ten inputs through `factory.create` with backend `azure`, marker in the value, full `getCause()` chain inspected. Every malformed input now surfaces as `AttachmentStorageException` with a value-free message and no cause: non-URL `sasUrl`, bad-host `sasUrl`, `sasUrl` with no share, `sasUrl` to a file path, a full connection string pasted into `sasUrl` (the account-key case), a garbage `connectionString`, and a blank `shareName`. Zero markers leaked; zero SDK exception types escaped. The two well-formed inputs and a connection string with a non-base64 account key still build a client, which is correct because the SDK validates the key lazily.
- The four new negative tests in `RegisteringStorageFactoryTest` assert the same property over the full cause chain, so this is now a regression check, not a one-time probe.
- `credentialsJson` on the gcs branch keeps its cause; that chain was probed value-free in the first pass across four malformed shapes and the branch is unchanged here.

### Regressions
- The guard catches `RuntimeException` from `buildClient()` and drops the cause. A genuine non-value SDK failure on construction (a classpath or SDK-config problem) will now report only "not a well-formed share SAS URL", which could misdirect an operator. Given the SDK builder validates the endpoint and connection string eagerly and does no I/O, this is the right trade, but it is worth knowing when reading a verifier failure. Follow-up only.
- No other files changed; the S3 and GCS adapters, SPI, and `TenantConfig` are as reviewed before. CI `audit` and `verify` both green at `790273c` (`gh pr checks 68`).

### Operational
- Unchanged from the first pass: per-`create()` clients are never closed (a property of the `StorageFactory` shape, bounded by onboarding frequency today), and Jackson 2 XML enters the classpath via `google-cloud-storage`. Both acknowledged in the Coder's disposition. Neither blocks.
- Ambiguous-finish hardening is tracked in `tsanetgit/Connect_SDK#69` under epic #46.

### Notes to track after merge
1. `tsanetgit/Connect_SDK#69` (ambiguous-finish identity check, S3 and GCS).
2. Client lifecycle for whichever hosting layer calls the factory per push rather than per onboarding.
3. Jackson 2 CVE noise in `security-audit` against this module.

Shawn merges; this is a `shawn-tsanet`-authored PR so sign-off is by comment.

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass.

**Correctness:**
The azure guard is correct. Traced the connection-string-in-sasUrl case: `!blank(sasUrl)` true → `builder.endpoint(pasted).buildClient()` throws `RuntimeException` (previously the leaking `IllegalArgumentException`/`MalformedURLException`) → caught → value-free `AttachmentStorageException` with no cause. The account key never enters any message. The `share` variable is now assigned in both branches before `forShare`, so no uninitialized-use path. The abandon-test key fix (`01234567/truncated.bin`) is proven non-vacuous by the finalize-on-abandon mutation the prior review re-ran. `fill()` message change is cosmetic and accurate.

One observation, not a finding: dropping the cause means a genuine construction failure (classpath/SDK-config) now reports only "not a well-formed share SAS URL" — a misdiagnosis risk. Acceptable given the builder validates eagerly and does no I/O; the prior review already logged this as a follow-up.

**Security:**
The blocking leak is closed on both azure branches. Both `endpoint(...).buildClient()` and `connectionString(...).shareName(...).buildClient()` are wrapped, throw value-free messages, attach no cause — correct, since here the cause was the leak vector. The four new negative tests assert `MARKER` appears nowhere in the full `getCause()` chain via `fullChain()`, which walks every cause, and assert the escaping type is `AttachmentStorageException`. This converts the one-time probe into a standing regression check. The gcs branch keeps its cause (probed value-free previously); unchanged here.

**Regression risk:**
Only `RegisteringStorageFactory.azure()` and two test files changed. The azure branch is exercised by `GoLiveVerifier.storageCheck` via `create()`; the change only narrows escaping exception types to the contract's `AttachmentStorageException`, which is what the `StorageFactory` SPI declares — strictly better contract adherence, no new escape path. No positional/vocabulary/schema change.

**Carried forward:**
- azure sasUrl/connectionString value leak — RESOLVED
- vacuous contentOf assertion — RESOLVED
- fill() message on single-shot path — RESOLVED
- ambiguous-finish stale-overwrite false positive — FOLLOW-UP (Connect_SDK#69)
- per-create() client lifecycle never closed — FOLLOW-UP
- Jackson 2 XML on classpath / CVE noise — FOLLOW-UP
- guard drops non-value SDK construction cause (misdiagnosis) — FOLLOW-UP

**Recommendation:** Merge — both blocking items are fixed and proven by re-run probes, and every remaining item is a tracked follow-up.

---

shawn-tsanet:
Follow-up for the ambiguous-finish hardening is now tracked: `tsanetgit/Connect_SDK#69` (ambiguous-finish probe must verify object identity, not just size, for both S3 and GCS). That closes the one open item from the disposition above; the "filing is Shawn's call" line there is superseded by `#69`.

Issue #71, opened 2026-09-04
attachment-receiver: Azure Blob Storage adapter (container-scoped), alongside Azure Files


shawn-tsanet:
**For triage:** label `enhancement`; Project: `11-2026 Release` (project number 40); use `09-2026 Release` (project number 39) instead if the core piece is pulled forward. Parent: `tsanetgit/Connect_SDK#46`. Serves: the core issue (`tsanetgit/Connect-API-Code#146`).

**Why**

The existing Azure adapter targets Azure Files shares (`tsanetgit/Connect_SDK#57`). Member self-service on the Integrations page expects Blob containers: it is the shape the bring-your-own-storage prior art uses, and Azure Files over REST with Entra needs a privileged data role and the backup-intent request header that Blob does not. Blob should be the page's default Azure option, with Files kept for members who need it.

**Change**

- `AzureBlobAttachmentStorage` implementing `AttachmentStorage` for one container (optional directory prefix), constructed by `RegisteringStorageFactory` for backend id `azure_blob` from either a container SAS URL, or a connection string plus container name (SAS preferred; connection string accepted).
- `store`: block-blob upload; large files via staged blocks and a single commit, so no partial object is visible (the same no-partial-visibility rule the other adapters keep); on failure, abandon without committing.
- `verifyAccess`: `Put Block` of an uncommitted block on a probe blob name (needs only write permission, leaves no committed object), replacing the write-read-delete sentinel for this adapter; failure classification as for the other adapters (401 wrong credential, 403 no permission, 404 wrong target).
- `GoLiveVerifier` and the contract test run unchanged; the offline test uses a strict in-memory double as the GCS adapter does; a live test gated on an environment variable runs against a disposable container.
- Naming by mechanism, not by member.

**Acceptance**

- `AttachmentStorageContractTest` passes offline against the double and live against a real container (counts only in the PR, no account or container names).
- A probe against a SAS with `cw` permissions succeeds and leaves no committed blob; a read-only SAS is classified as no permission.
- Factory rejects a malformed SAS URL or connection string with a value-free message (the `tsanetgit/Connect_SDK#68` rule).

shawn-tsanet:
## Implementation plan

Branch: `feat/azure-blob-adapter`. Draft PR only; QA reviews before merge.

**Verified before planning**

- Baseline: `mvn -pl attachment-receiver -am test` on a clean `main` is green (117 run, 0 failures, 23 env-gated live skips).
[...]

Issue #69, opened 2026-09-02
attachment-receiver: ambiguous-finish probe must verify object identity, not just size (S3 and GCS)


shawn-tsanet:
Part of the P4 attachment-receive epic: #46. Follow-up from the QA review of `tsanetgit/Connect_SDK#68` (GCS adapter), where this was disclosed as a known parity caveat and deferred.

## The gap

Both the S3 and GCS adapters resolve an ambiguous finalize (the client-side `CompleteMultipartUpload` / resumable `finish()` fails after the server may have committed) by probing the destination key and treating "object visible at exactly the byte count written" as a committed store.

- `S3AttachmentStorage.resolveAmbiguousComplete` (merged in `tsanetgit/Connect_SDK#56`) additionally requires the abort to answer `NoSuchUpload`, which proves the upload session itself is gone.
- `GcsAttachmentStorage.resolveAmbiguousFinish` (`tsanetgit/Connect_SDK#68`) has no such second signal, because GCS has no abort call. It relies on the size probe alone.

Because `store` deliberately overwrites a same-name object (`GcsBucket.create` javadoc, and the `sameNameStoreOverwrites` test), the size probe has a false-positive shape: a pre-existing object of the same name and the same size, plus a finish that never committed, is reported as success while the stale bytes remain. The caller believes the new content is durable. On S3 the `NoSuchUpload` check narrows the window but does not close it, since a session can also be gone because it expired.

Concurrent same-name writers are a second, already-acknowledged instance of the same blind spot in both adapters.

## Hardening options

1. **Identity, not size.** Stamp each resumable attempt with a per-attempt marker (custom metadata UUID on the `BlobInfo` / `CreateMultipartUpload` request) and have the ambiguous-finish probe read the object's metadata and compare the marker. Requires widening the seam (`GcsBucket.sizeOrAbsent`, the S3 equivalent) to return metadata.
2. **Generation / ETag captured before finish.** Record the destination's generation (GCS) or ETag (S3) before finalizing and require the post-failure probe to show a different value.
3. **Content checksum.** Compute CRC32C (GCS) or the multipart ETag (S3) over the streamed bytes and compare with the visible object's checksum. Strongest, but costs a hash over every resumable upload.

Option 1 or 2 is enough to close the stale-overwrite shape; option 3 also closes the concurrent-writer shape.

## Acceptance

[...]

PR #72, opened 2026-09-07
attachment-receiver: Azure Blob Storage adapter (container-scoped), alongside Azure Files

shawn-tsanet:
Closes `tsanetgit/Connect_SDK#71` (epic `tsanetgit/Connect_SDK#46`). Serves `tsanetgit/Connect-API-Code#146`.

**Account-hygiene rule, binding for this public repo:** no storage account, container, connection string, or subscription identifier appears in code, commits, or this PR. Live results are counts and outcomes only.

## What this adds

`AzureBlobAttachmentStorage`, the container-scoped sibling of the Azure Files adapter, registered in `RegisteringStorageFactory` as backend id `azure_blob` (container SAS URL, or connection string plus `containerName`; optional `prefix`). `azure` keeps meaning Azure Files. Structurally the GCS twin (`tsanetgit/Connect_SDK#68`): a package-private `AzureBlobContainer` seam over the final SDK clients, a logic-free `SdkAzureBlobContainer`, a strict in-memory double for the offline contract run.

- **Small files** are one `Put Blob`; **larger files** are staged blocks (8 MiB) and one `Put Block List`, overwrite on. A blob that only received staged blocks is invisible to `Get Blob Properties` and the service reclaims it after seven days, so a failed store abandons with no commit and no cleanup round-trip. Proven live: after a stream dies past the first block, `exists` is false while the blob is enumerable only with `include=uncommittedblobs`.
- **Ambiguous commit resolves by identity, not size.** Block ids embed a UUID minted per store call, so a `Put Block List` that failed client-side is confirmed only if the committed block list equals this call's ids. The S3/GCS size heuristic (`tsanetgit/Connect_SDK#69`) is deterministically fooled by a same-size overwrite; the offline test `ambiguousCommitOverAPreExistingSameSizeBlobThrows` is that exact case, and it throws here. #69 does not apply to this adapter.
- **`verifyAccess` stages one uncommitted block** on a `.verify-<uuid>` name: write permission only, nothing committed, nothing to delete, as the issue specifies. Stated consequence, in the javadoc: **the probe validates `w` only, while `exists` and the ambiguous-commit probe need `r`.** A `cw`-only SAS passes go-live and fails `exists` at runtime. Required rights are `rw` (or `rcw`) on the container, or Storage Blob Data Contributor.
- **Classification reads the error code before the HTTP status** because Azure answers `AuthenticationFailed` with 403. Proven live: a tampered signature classifies as wrong credential, a read-only SAS as no permission.
- **Factory** rejects a malformed SAS URL or connection string with a value-free message and **no cause** (the #68 rule, MARKER-probed across the full cause chain on every branch). It also rejects an account-level SAS URL by its parsed container name: verified with the SDK that `BlobContainerClientBuilder.endpoint()` on a URL with no container path builds a client for `$root` without throwing.

## Correction to the plan on the issue: names are encoded, not opaque

The plan said blob names stay opaque like S3/GCS. The first live run said otherwise. With opaque names, the service normalized the request path: `../../etc/passwd` was listed under a different name outside the case prefix, a trailing dot was stripped, and a backslash became a separator. Store and `exists` still agreed, so the SPI held, but a remote-controlled name escaping the tenant prefix breaks multi-tenant isolation in the hosted mode. Both name components now go through the existing `AzureFileNames` encoder (same package, reused unchanged): `/`, `\`, `%`, leading and trailing dots are percent-encoded, so nothing can traverse and nothing can start with a dot (which keeps the `.verify-` probe namespace collision-free). This is a representation change rather than a guard: there is no name that needs rejecting. The double now throws on any raw traversal component reaching the seam, so the offline traversal test proves the encoder neutralized the name rather than the fake being lenient.

Live hostile-name outcomes (second run, every name literal under its encoded form):

| pushed name | stored as |
|---|---|
| `../../etc/passwd` | `%2E.%2F..%2Fetc%2Fpasswd` |
| `a/b.txt` | `a%2Fb.txt` |
| `trailing.` | `trailing%2E` |
| `trailing/` | `trailing%2F` |
| `back\slash.txt` | `back%5Cslash.txt` |
| `report#1.txt` | `report%231.txt` |
| `50%.log` | `50%25.log` |
| `spaces and üñïçode.txt` | `spaces%20and%20%C3%BC%C3%B1%C3%AF%C3%A7ode.txt` |
| `q?mark.txt` | `q%3Fmark.txt` |
| `.verify-fake` | `%2Everify-fake` |

## Verification

- Offline, `mvn -pl attachment-receiver -am test`: **166 run, 0 failures, 37 skipped** (baseline on `main` was 117 / 0 / 23; the 37 skipped are the env-gated S3, Azure Files, GCS, and Azure Blob live classes). New offline tests: 20 in `AzureBlobAttachmentStorageTest` (7 shared-contract via the strict double plus the single-shot/staged split, exact-boundary, empty file, abandon on stream death and on stage failure, both ambiguous-commit shapes plus the same-size trap, overwrite, traversal and probe-impersonation and oversized names, all three classification groups driven by real `BlobStorageException`s with the `x-ms-error-code` header, exists-throws-on-outage, block-id shape) and 8 in `RegisteringStorageFactoryTest`.
- Live, `AzureBlobAttachmentStorageLiveTest` against a disposable container in a private test account, gated on `AZURE_BLOB_CONNECTION_STRING` + `AZURE_BLOB_CONTRACT_TEST_CONTAINER`: **14 / 14** in about 37 s. The 7 shared-contract tests, abort after staged blocks, a real staged commit with byte-for-byte readback, overwrite on both paths (small over nothing, staged over small, small over staged, by size), the `cw`-only SAS probe leaving no committed blob, read-only SAS as no permission, tampered signature as wrong credential, and the ten hostile names above.
- SDK facts checked against the jar with `javap` before use: `stageBlock`, `commitBlockListWithResponse`, the options types that carry content type, `listBlocks(COMMITTED)`, and that `BlobStorageException.getErrorCode()` reads the `x-ms-error-code` header (so a HEAD carries it). Azure REST docs for Put Block, Put Block List, Get Blob Properties, and blob naming read today.

| claim | line |
|---|---|
| Uncommitted blocks are invisible; abandon leaves nothing | `staged()` commits only after the loop; live `abortAfterStagedBlocksLeavesNothingVisible` |
| Ambiguous commit is confirmed by identity | `resolveAmbiguousCommit` compares `blockIds.equals(committedBlockIdsOrAbsent(name))` |
| Block ids are fixed-length under the 64-byte cap | `blockId()` is `%s-%05d` of a UUID, 42 bytes; test `blockIdsAreFixedLengthBase64UnderTheServiceCap` |
| Options overloads overwrite (no `If-None-Match`) | live `sameNameStoreOverwritesOnBothPaths` |
| Store and exists cannot disagree | both call `nameOrThrow` |
| Only `BlobNotFound` reads as absent | `SdkAzureBlobContainer.sizeOrAbsent` / `committedBlockIdsOrAbsent` catch that code and rethrow the rest |
| Account-level SAS URL builds a `$root` client | jshell probe against the SDK; factory test `azureBlobSasUrlWithoutContainerNameIsAConfigError` |
| Probe needs `w` only | live `cwOnlySasPassesTheProbeAndLeavesNoCommittedBlob` |

No mechanical detector fired. The judgment lines:

- **Single-place decision.** `grep -rn 'forContainer(\|forShare(\|forBucket(\|new S3AttachmentStorage(' src/main`: every construction is inside `RegisteringStorageFactory`; the only other hits are the three factory-method declarations. Old form of the decision (`"s3"` / `"azure"` / `"gcs"` literals outside the factory): none in `src/main` beyond a javadoc example in `TenantConfig`.
- **Vocabulary gained a member.** The unknown-backend message now lists `azure_blob`, asserted by `unknownBackendMessageListsEveryRegisteredBackend`. Repo-wide doc sweep: one stale surface, `skills/connect-sdk-deploy/SKILL.md` line 36 enumerates "AWS S3, Azure Files, and Google Cloud Storage adapters". Left untouched here because that skill carries its own revision convention; follow-up below.
- **Guard vs representation.** The one guard is the factory's `$root`/blank container-name check, needed because the builder does not throw. Name safety is a representation change (encoding), see above.
- **Schema / stored shape.** The factory reads new keys from the existing free-form `storageProperties` map; pre-existing configs lack them and get a clear `require()` error. No table, contract, or record changed.
- **Newly reachable failures.** Every new failure surfaces as `AttachmentStorageException` toward the gated ingest endpoint (`tsanetgit/Connect_SDK#51`) or as the STORAGE check's FAIL line in `GoLiveVerifier`, the same paths the other adapters use.
- **Dependency.** `com.azure:azure-storage-blob`, version 12.30.0 resolved by the `azure-sdk-bom` 1.2.35 already in the module pom (still the latest stable BOM on Maven Central today).

## Untouched on purpose

- `GoLiveVerifier` and `AttachmentStorageContractTest`, per the issue. The verifier's PASS line still says "write-read-delete probe", which is inexact for this adapter.
- `AzureFileNames` is reused, not modified. Its 255-character component cap is stricter than Blob's 1,024-character name limit; an over-long name fails fast through the SPI's checked exception (offline test).

- `skills/connect-sdk-deploy/SKILL.md` should list the Azure Blob adapter and the `azure_blob` backend id on its next revision.
- `GoLiveVerifier`'s PASS wording, if the probe description should be per-adapter.

## Advisor and gate

Advisor consulted twice. The design call agreed with the GCS-shaped structure and asked for `getProperties`-based absence with only `BlobNotFound` mapped, the `$root` builder probe, all three classification groups in the double, and content type via the options overloads. The pre-done call found the size-based ambiguous-commit false positive on a same-size overwrite (fixed: identity resolution via the block list), and that the first live class never executed a real commit or overwrite (fixed: two live tests). Pre-PR gate: **LOOKS GOOD**, first run. Review is this repo's CI plus a QA-persona pass from a separate session.

### DIFF
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -63,6 +63,10 @@
             <artifactId>azure-storage-file-share</artifactId>
         </dependency>
         <dependency>
+            <groupId>com.azure</groupId>
+            <artifactId>azure-storage-blob</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.google.cloud</groupId>
             <artifactId>google-cloud-storage</artifactId>
         </dependency>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
@@ -1,0 +1,289 @@
+package com.tsanet.receiver.storage.azure;
+
+import com.tsanet.receiver.storage.AttachmentStorage;
+import com.tsanet.receiver.storage.AttachmentStorageException;
+import com.tsanet.receiver.storage.IncomingAttachment;
+import com.tsanet.receiver.storage.StoredAttachment;
+import com.azure.storage.blob.BlobContainerClient;
+import com.azure.storage.blob.models.BlobErrorCode;
+import com.azure.storage.blob.models.BlobStorageException;
+
+import java.io.IOException;
+import java.io.InputStream;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Base64;
+import java.util.List;
+import java.util.Objects;
+import java.util.UUID;
+
+/**
+ * {@link AttachmentStorage} on one Azure Blob Storage container (block blobs): the
+ * container-scoped sibling of {@link AzureFilesAttachmentStorage}. Blob is the default
+ * Azure option for member self-service; Files stays for members who need a share
+ * (tsanetgit/Connect_SDK#71). The client is injected: account, container and credential
+ * are wiring concerns, not this class's.
+ *
+ * <p>Blob names are {@code [prefix/]encode(caseNumber)/encode(fileName)}, both components
+ * passed through the same {@link AzureFileNames} encoder the Files adapter uses. Blob
+ * Storage is documented as a flat namespace, but the live acceptance run proved that the
+ * request path is normalized on the way in: {@code ../../etc/passwd} landed outside the
+ * case prefix, a trailing dot was stripped, and a backslash became a separator. Store
+ * and exists agreed on every one of them, so the SPI contract held, but a remote-controlled
+ * name escaping the tenant prefix would break multi-tenant isolation in the hosted mode.
+ * Encoding removes the possibility rather than guarding against it: {@code /}, {@code \},
+ * {@code %}, leading and trailing dots are all percent-encoded, so no name can traverse
+ * and no name can start with a dot (which keeps the {@code .verify-} probe namespace
+ * collision-free). {@link #exists} builds names through the same method, so store and
+ * check can never disagree. The encoder's 255-character component cap is stricter than
+ * Blob's 1,024-character name limit; an over-long name fails fast through the SPI's
+ * checked exception.
+ *
+ * <p>Streaming: bytes are read in {@value #BLOCK_SIZE}-byte buffers. A file that ends
+ * inside the first buffer is written with a single {@code Put Blob}, one atomic request.
+ * Anything larger is staged as blocks ({@code Put Block}) and committed with one
+ * {@code Put Block List}. Two Blob Storage facts make the SPI's no-partial-visibility rule
+ * hold structurally:
+ * <ul>
+ *   <li><b>Uncommitted blocks are invisible.</b> A blob that has only received staged
+ *       blocks reads as {@code BlobNotFound}, and the service garbage-collects it seven days
+ *       after the last {@code Put Block}. On any mid-stream failure the upload is
+ *       abandoned, never committed — committing a partial block list would publish a
+ *       truncated object — and there is deliberately no cleanup round-trip that could
+ *       itself fail (the GCS adapter's shape, unlike S3's abort call).</li>
+ *   <li><b>Ambiguous commit, resolved exactly.</b> A {@code Put Block List} that fails
+ *       client-side may have committed server-side. The S3 and GCS adapters resolve that
+ *       by size, which a same-size overwrite or a concurrent writer can fool
+ *       (tsanetgit/Connect_SDK#69). Here the block ids carry a UUID minted for this store
+ *       call, so the resolution is identity, not size: the blob's committed block list
+ *       ({@code Get Block List}) equal to the ids this call staged means this commit
+ *       landed, and nothing else can produce that list. Any other shape throws.</li>
+ * </ul>
+ * Block ids are fixed-length, as the service requires: one UUID per store call plus a
+ * zero-padded index. With {@value #BLOCK_SIZE}-byte blocks the service's 50,000-block
+ * ceiling is about 390 GiB, far above any platform cap. Same-name behavior is overwrite
+ * (proven live for both the single-request and the staged path), consistent with the
+ * other adapters while the policy stays implementation-defined upstream
+ * (tsanetgit/Connect-API-Code#140, question 2).
+ *
+ * <p><b>Go-live probe.</b> {@link #verifyAccess} stages one uncommitted block on a
+ * {@code .verify-<uuid>} name: it needs only write permission and leaves no committed blob,
+ * replacing the other adapters' write-read-delete sentinel (this is the issue's design).
+ * The consequence, stated rather than hidden: <b>the probe validates write only, while
+ * operation needs read as well</b> — {@link #exists} and the ambiguous-commit probe are
+ * {@code Get Blob Properties}. A SAS carrying only {@code cw} passes go-live and fails
+ * {@code exists} at runtime. Required rights are therefore SAS {@code rw} (or {@code rcw})
+ * on the container, or the Storage Blob Data Contributor role for an Entra identity.
+ *
+ * <p>Failure classification reads the service error code before the HTTP status, because
+ * Azure Storage answers {@code AuthenticationFailed} with 403, not 401.
+ */
+public final class AzureBlobAttachmentStorage implements AttachmentStorage {
+
+    /** Stream read granularity and staged block size: 8 MiB (the service allows up to 4,000 MiB). */
+    static final int BLOCK_SIZE = 8 * 1024 * 1024;
+
+    static final String VERIFY_PREFIX = ".verify-";
+
+    private final AzureBlobContainer container;
+    private final String containerName;
+    private final String prefix;
+
+    /** Production entry point: wraps the injected client in the logic-free SDK seam. */
+    public static AzureBlobAttachmentStorage forContainer(BlobContainerClient client, String prefix) {
+        return new AzureBlobAttachmentStorage(new SdkAzureBlobContainer(client),
+                client.getBlobContainerName(), prefix);
+    }
+
+    AzureBlobAttachmentStorage(AzureBlobContainer container, String containerName, String prefix) {
+        this.container = Objects.requireNonNull(container, "container");
+        this.containerName = Objects.requireNonNull(containerName, "containerName");
+        if (containerName.isBlank()) {
+            throw new IllegalArgumentException("containerName must not be blank");
+        }
+        this.prefix = normalize(prefix);
+    }
+
+    @Override
+    public StoredAttachment store(IncomingAttachment attachment, InputStream content)
+            throws AttachmentStorageException {
+        String name = nameOrThrow(attachment.caseNumber(), attachment.fileName());
+        byte[] buffer = new byte[BLOCK_SIZE];
+        int firstLength = fill(content, buffer, name);
+        if (firstLength < BLOCK_SIZE) {
+            // EOF inside the first buffer: one atomic Put Blob, no partial-visibility window.
+            try {
+                container.putBlob(name, attachment.contentType(), Arrays.copyOf(buffer, firstLength));
+            } catch (RuntimeException e) {
+                throw wrap("put blob", name, e);
+            }
+            return new StoredAttachment(name, firstLength);
+        }
+        return staged(attachment, content, name, buffer);
+    }
+
+    private StoredAttachment staged(IncomingAttachment attachment, InputStream content,
+                                    String name, byte[] buffer)
+            throws AttachmentStorageException {
+        String uploadId = UUID.randomUUID().toString();
+        List<String> blockIds = new ArrayList<>();
+        long total = 0;
+        // A stream failure inside fill() propagates as the AttachmentStorageException it
+        // already is (name and phase included); abandoning is the same no-op either way.
+        try {
+            int length = BLOCK_SIZE; // the first buffer arrived full, or we would not be here
+            boolean last = false;
+            while (!last) {
+                if (!blockIds.isEmpty()) {
+                    // The one buffer is reused: stageBlock is synchronous, so the SDK has
+                    // finished with it before the next fill overwrites it.
+                    length = fill(content, buffer, name);
+                    if (length == 0) {
+                        break; // EOF landed exactly on the previous block's boundary.
+                    }
+                }
+                String blockId = blockId(uploadId, blockIds.size());
+                container.stageBlock(name, blockId, buffer, length);
+                blockIds.add(blockId);
+                total += length;
+                last = length < BLOCK_SIZE;
+            }
+        } catch (RuntimeException stageFailure) {
+            // Abandon: nothing was committed, so nothing is visible; the blocks expire.
+            throw wrap("stage block", name, stageFailure);
+        }
+
+        try {
+            container.commitBlockList(name, blockIds, attachment.contentType());
+        } catch (RuntimeException commitFailure) {
+            return resolveAmbiguousCommit(name, blockIds, total, commitFailure);
+        }
+        return new StoredAttachment(name, total);
+    }
+
+    /**
+     * A failed commit is ambiguous: the service may have committed. The committed block
+     * list equal to the ids this call staged means exactly that — no other writer holds
+     * this call's UUID — so report the commit as the success it was. A different list
+     * (a previous blob of the same name, another writer's, or nothing at all) throws.
+     */
+    private StoredAttachment resolveAmbiguousCommit(String name, List<String> blockIds, long total,
+                                                    RuntimeException commitFailure)
+            throws AttachmentStorageException {
+        try {
+            if (blockIds.equals(container.committedBlockIdsOrAbsent(name))) {
+                return new StoredAttachment(name, total);
+            }
+        } catch (RuntimeException probeFailure) {
+            commitFailure.addSuppressed(probeFailure);
+        }
+        throw wrap("commit", name, commitFailure);
+    }
+
+    @Override
+    public boolean exists(String caseNumber, String fileName) throws AttachmentStorageException {
+        String name = nameOrThrow(caseNumber, fileName);
+        try {
+            return container.sizeOrAbsent(name) >= 0;
+        } catch (RuntimeException e) {
+            throw wrap("exists check", name, e);
+        }
+    }
+
+    @Override
+    public void verifyAccess() throws AttachmentStorageException {
+        String probeName = (prefix.isEmpty() ? "" : prefix + "/") + VERIFY_PREFIX + UUID.randomUUID();
+        byte[] probe = "attachment-receiver go-live probe".getBytes(StandardCharsets.UTF_8);
+        try {
+            // One uncommitted block: proves write permission, commits nothing, and the
+            // service reclaims it in seven days. No read, no delete, nothing to clean up.
+            container.stageBlock(probeName, blockId(UUID.randomUUID().toString(), 0), probe, probe.length);
+        } catch (RuntimeException e) {
+            throw new AttachmentStorageException(classify(e), e);
+        }
+    }
+
+    /** Wrong-credential, no-permission, and wrong-target must read differently. */
+    private String classify(RuntimeException e) {
+        if (e instanceof BlobStorageException bse) {
+            BlobErrorCode code = bse.getErrorCode();
+            int status = bse.getStatusCode();
+            if (BlobErrorCode.AUTHENTICATION_FAILED.equals(code)
+                    || BlobErrorCode.INVALID_AUTHENTICATION_INFO.equals(code)
+                    || BlobErrorCode.NO_AUTHENTICATION_INFORMATION.equals(code)) {
+                return "wrong credential: the container rejected the identity (verify write, " + code + ")";
+            }
+            if (BlobErrorCode.AUTHORIZATION_FAILURE.equals(code)
+                    || BlobErrorCode.AUTHORIZATION_PERMISSION_MISMATCH.equals(code)
+                    || BlobErrorCode.INSUFFICIENT_ACCOUNT_PERMISSIONS.equals(code)
+                    || status == 403) {
+                return "no permission: write denied on container '" + containerName + "'"
+                        + (code == null ? "" : " (" + code + ")");
+            }
+            if (BlobErrorCode.CONTAINER_NOT_FOUND.equals(code) || status == 404) {
+                return "wrong target: container '" + containerName
+                        + "' (or its account) does not exist (verify write)";
+            }
+            return "verify write failed on container '" + containerName + "': " + code
+                    + " (HTTP " + status + ")";
+        }
+        return "connectivity: cannot reach the container (verify write): " + e.getMessage();
+    }
+
+    private int fill(InputStream content, byte[] buffer, String name)
+            throws AttachmentStorageException {
+        try {
+            // readNBytes loops over short reads; < buffer.length only ever means EOF.
+            return content.readNBytes(buffer, 0, buffer.length);
+        } catch (IOException e) {
+            throw new AttachmentStorageException(
+                    "stream failed mid-read for " + name + "; nothing committed", e);
+        }
+    }
+
+    /**
+     * Fixed length by construction for every index below 100,000 (36-char UUID, dash,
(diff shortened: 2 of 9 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `pom.xml` — adds `azure-storage-blob` dependency. Traces to "New dependency" / the adapter needs the SDK. ✓
- `AzureBlobAttachmentStorage.java` (new) — the adapter itself. Core of the PR. ✓
- `AzureBlobContainer.java` (new) — the package-private seam described in the PR body. ✓
- `SdkAzureBlobContainer.java` (new) — logic-free SDK impl, PR body. ✓
- `RegisteringStorageFactory.java` — registers `azure_blob`, updates the javadoc table and the unknown-backend message. All traces to "registered in RegisteringStorageFactory as backend id azure_blob". ✓
- Three new test files + factory test additions — trace to the "Verification" section. ✓

No untraced hunks. No incidental tidying. `AzureFileNames` was reused unmodified (confirmed: it does not appear in the diff), consistent with the PR's claim.

**Correctness:**

Traced a large-file store through `staged()` (AzureBlobAttachmentStorage.java ~L94–L128):

- First `fill()` returns `BLOCK_SIZE` (buffer full), so `store()` calls `staged()`.
- Loop iteration 0: `blockIds` empty, so it skips the re-fill, uses the length carried from `store()` (`BLOCK_SIZE`), stages block 0, `last = (BLOCK_SIZE < BLOCK_SIZE) = false`.
- Iteration 1: `blockIds` non-empty, re-fills. Say EOF lands at 4096 bytes → `length=4096`, stages block 1, `last=true`. Commit fires. ✓

Edge case — EOF exactly on block boundary (`eofExactlyOnBlockBoundaryCommitsOneBlob` test): iteration 0 stages the full block (`last=false`), iteration 1 re-fills, `readNBytes` returns 0, `break` before staging an empty block. Correct — one block committed, no empty trailing block. ✓

Ambiguous-commit resolution by identity (`resolveAmbiguousCommit`, L133–L142): `blockIds.equals(committedBlockIdsOrAbsent(name))`. When nothing committed, `committedBlockIdsOrAbsent` returns `null`; `blockIds.equals(null)` is `false` → throws. When a pre-existing same-size blob has *different* ids (its own UUID), `.equals` is false → throws. The same-size-trap test exercises this. Sound — this is a genuine improvement over the S3/GCS size heuristic.

One thing worth noting (not a bug): in `resolveAmbiguousCommit`, if the identity probe itself throws, the probe failure is added as suppressed and the original commit failure is rethrown — the store fails closed. Correct for the SPI's no-false-success rule.

`store()` small-file path: `firstLength < BLOCK_SIZE` includes the empty-file case (`firstLength=0`), one `putBlob` with a zero-length array. `emptyFileStoresAZeroByteBlob` covers it. ✓

`verifyAccess` mints its own separate UUID for the block id vs. the probe name's UUID — harmless, single block, no correctness impact.

Classification order (L157–L180): error code checked before status, with `AuthenticationFailed`→wrong-credential ahead of the 403→no-permission fallback. The `403` fallback in the no-permission branch catches a bare-403-no-code as no-permission, and the test `verifyAccessClassifiesNoPermission` asserts the bare-403 case. Correct and matches the PR's stated Azure quirk. ✓

**Security:**

- No secrets, keys, or connection strings in the diff. Test connection strings use `AccountKey=Zm9vYmFyYmF6` (base64 "foobarbaz") and `MARKER` placeholders. ✓
- The factory error messages are value-free and thrown with **no cause** on the malformed-SAS/connection-string branches (L131–L160), matching the #68 rule. The `MARKER` tests assert the value never reaches the full cause chain. ✓
- Traversal/name-normalization is handled by encoding both name components through `AzureFileNames` (`nameOrThrow`, L233–L242), and `exists` uses the same method — store/exists cannot disagree. The offline double's `validate()` throws on raw `.`/`..`/`\`/empty components reaching the seam, so a passing traversal test proves the encoder neutralized the name rather than a lenient fake. This is the correct defense (representation change, not a guard). ✓
- No log statements printing secrets. The one `System.out.println` (live test, hostile-name diagnostic) prints only the *encoded* storage key suffix, and only in the env-gated live class.

**Regression risk:**

- `RegisteringStorageFactory.create` is the only modified production consumer. The `switch` gains one arm; existing `s3`/`azure`/`gcs` arms are byte-for-byte unchanged. The `default` message string changed (now lists `azure_blob`) — the only consumer of that string is `unknownBackendMessageListsEveryRegisteredBackend`, which is updated in the same PR. No other production caller inspects that message.
- `azure` still means Azure Files — the existing `azure(props)` method and `AzureFilesAttachmentStorage` are untouched. No vocabulary collision.
- `AzureFileNames` reused unmodified — no risk to the Azure Files adapter that already depends on it.
- New dependency `azure-storage-blob` resolved via the `azure-sdk-bom` already in the pom (no new version pin). Low risk, but it does pull a new transitive tree into the module — worth a build-clean confirmation, which the PR's `mvn test` (166/0/37) provides.
- Downstream `GoLiveVerifier` and `AttachmentStorageContractTest` untouched by design; the new adapter satisfies the same contract via `AttachmentStorageContractTest` inheritance (both offline double and live). No regression to those harnesses.

Blast-radius checklist: vocabulary sweep done (PR notes the one stale doc surface in `skills/connect-sdk-deploy/SKILL.md` L36, deferred as a follow-up — acceptable, it's a skill doc with its own revision convention, not a runtime contract). Single-place-decision sweep done (all construction inside the factory). No schema/stored-shape change (free-form `storageProperties` map, pre-existing configs get a clear `require()` error). Newly-reachable failures all surface as `AttachmentStorageException` toward the existing ingest path — same handler as the other adapters.

**Operational:** New Maven dependency (`azure-storage-blob`, via existing BOM) — `pyproject.toml`/`.env.example` are not relevant here (Java module). No ops-sidecar, mirror, or invoice code touched; no LaunchAgent restart. The two documented follow-ups (skill doc, GoLiveVerifier PASS wording) are correctly scoped out.

**Recommendation:** Merge — the diff is fully traced, the ambiguous-commit-by-identity logic is a real correctness improvement over the size heuristic, security/secret handling matches the established #68 rule, and the two open items are legitimately post-merge follow-ups.

---

shawn-tsanet (commented):
## QA review of attachment-receiver: Azure Blob Storage adapter (container-scoped), alongside Azure Files

Separate-session QA pass at head `a9cdeab`, independent of the auto-QA comment above: I read the diff, ran the suite, and probed the SDK myself rather than grading from that comment.

Nothing here changes the merge decision. The notes are real, each is small, and the first one is the only one I would call security-adjacent.

Receipts: `mvn -pl attachment-receiver -am test` in the local clone at `a9cdeab`: **166 run, 0 failures, 0 errors, 37 skipped**, matching the PR body; per class, `AzureBlobAttachmentStorageTest` 27 / 0, `RegisteringStorageFactoryTest` 20 / 0, the live class 14 skipped (env-gated). The run used the machine's default JDK 26 rather than the 21 the harness pins; nothing in the diff is version-sensitive, but that is the receipt's caveat. `mvn dependency:tree` resolves `com.azure:azure-storage-blob` to **12.30.0** through the BOM, and Maven Central search shows `azure-sdk-bom` 1.2.35 is still the latest release. The live claims (hostile-name table, `cw`-only probe, tampered signature) I could not re-run: no Blob container credentials in this session. They are stated as counts and outcomes with named tests, which is the standard this repo set.

### Minimal-Diff Rule
- Pass. Four new main files, four new or extended test files, one dependency line, and the factory arm plus its javadoc row and message. Every hunk traces to the issue. `AzureFileNames` is reused and does not appear in the diff.

### Correctness
- Traced `staged()`: the first block reuses the buffer `store()` filled, later iterations refill, a zero-length refill breaks before staging an empty block, and `last` flips on the first short read. The exact-boundary and empty-file cases are each covered by a test I read, not just named.
- `resolveAmbiguousCommit` compares this call's id list against `Get Block List`; a `null` (absent) or a previous blob's ids fail `equals`, so only this call's commit reads as success. The block-id base64 is 42 bytes, a multiple of three, so no padding can make the service's echoed ids differ from the ones sent. The same-size-overwrite trap test is the right wrong-but-well-formed input for the size heuristic this replaces.
- The in-memory double is genuinely strict: `sizeOrAbsent` ignores uncommitted blocks, `commit` rejects unstaged ids, `putBlob` discards uncommitted blocks, and `validate` throws on a raw `.`/`..`/`\`/empty component. A lenient fake would not pass `traversalFileNamesAreEncodedBeforeTheyReachTheContainer`.
- **One prose claim is false.** `AzureBlobContainer.java:331-332` says "Blob names are opaque to the service, so nothing is encoded on either side of this seam." This PR's own headline correction is that Blob names are not opaque: the service normalized `..`, a trailing dot and a backslash in the first live run. The adapter encodes *before* the seam, so the seam sees encoded names, but the sentence restates the disproved plan. Suggest: "Names reach this seam already encoded by the adapter; the service is not name-opaque (see the adapter javadoc)."
- **Two failure paths have no test.** `InMemoryAzureBlobContainer` declares `failPutBlobWith` (line 39) and `failCommittedBlockIdsWith` (line 43) and honors them, but no test sets either (grep across `src/test`). Untested as a result: the single `Put Blob` failure wrap in `store()`, and the branch of `resolveAmbiguousCommit` where the `Get Block List` probe itself throws and must be added as suppressed to the commit failure. Both are two-line tests in the existing pattern; the PR body does not claim them, so this is a gap, not a false claim.

### Security
- **The factory promises `https` and does not check it.** The `azure_blob` `sasUrl` branch's error text says the value "must be an https URL", but nothing tests the scheme. Probe with jshell against the module classpath: `new BlobContainerClientBuilder().endpoint("http://acct.blob.core.windows.net/attachments?sv=…&sig=…").buildClient()` returns a client with container `attachments` and URL `http://…`. An operator who pastes an `http://` SAS URL gets a working adapter whose SAS token travels in cleartext on every request (the service refuses it only if the account keeps "secure transfer required" on, which is the default but not guaranteed). Fix is one `startsWith("https://")` check that throws the same value-free message, plus a `MARKER` test. The pre-existing `azure` (Files) branch uses the same wording at line 112 and, as far as I can see, has the same gap; worth fixing both in one pass, but only the new branch is this PR's.
- No secrets in the diff; test connection strings use a placeholder key and `MARKER` values, and every factory branch that could echo a value throws value-free with no cause, asserted across the full cause chain.
- Name safety is a representation change (percent-encoding via the existing encoder), and `exists` builds names through the same method as `store`, so the two cannot disagree.
- Runtime `wrap()` includes the SDK exception's message, the same shape the Azure Files adapter uses at its lines 90, 174 and 238. One cheap assertion would close a question neither adapter answers: in the live `tamperedSignatureIsClassifiedAsWrongCredential`, assert the SAS `sig` value is absent from the full cause chain, so the service's `AuthenticationFailed` detail text is proven not to echo it.

### Regressions
- The factory's `s3` / `azure` / `gcs` arms are unchanged; the only changed string is the unknown-backend message, whose sole consumer is the updated test. `azure` still means Azure Files.
- The `sasUrl` branch has no happy-path test: every `azure_blob` SAS test in `RegisteringStorageFactoryTest` is a malformed or account-level case. Building a client is offline-safe, so a `https://acct.blob.core.windows.net/attachments?…` case asserting `AzureBlobAttachmentStorage` costs nothing and would have caught a scheme check that rejected too much.
- Doc sweep confirmed: `skills/connect-sdk-deploy/SKILL.md:36` still enumerates three adapters. The PR lists it as a follow-up under that skill's own revision convention, which is fine.

### Operational
- New dependency via the existing BOM; no new version pin. No ops-sidecar or LaunchAgent surface.
- **Probe coverage regressed relative to the sibling, by design.** The Azure Files probe is write-read-delete (its `verifyAccess`, lines 184-199), so a share SAS missing read fails at go-live. The Blob probe stages one block, so a container SAS missing `r` passes go-live and fails on the first `exists`. The adapter javadoc states this plainly and it matches the issue's design, but the factory javadoc table, where an operator configuring `azure_blob` will look, says nothing about required rights. Follow-up worth filing: after staging the probe block, call `Get Blob Properties` on the same `.verify-` name. Read allowed answers `BlobNotFound` (404), read denied answers `AuthorizationPermissionMismatch` (403), and nothing is committed either way. That validates `r` and `w` at zero cost and restores the sibling's guarantee. Until then, add the `rw` / `rcw` / Storage Blob Data Contributor requirement to the factory table row.

### Nits
- None beyond the javadoc sentence above.

shawn-tsanet:

Addressed in `0e5bb51`, pushed to this branch. Offline after the change: **170 run, 0 failures, 37 skipped** (four new tests over the 166 QA measured). Live class rerun against the same disposable container: **14 / 14**.

|---|---|---|
| Seam javadoc restates the disproved "names are opaque to the service" claim | **Fixed.** `AzureBlobContainer` now says names arrive already encoded by the adapter and that the service is not name-opaque. | `AzureBlobContainer.java`, class javadoc |
| `failPutBlobWith` and `failCommittedBlockIdsWith` declared but never exercised | **Fixed.** `putBlobFailureSurfacesThroughTheContractAndLeavesNothingVisible` (Put Blob failure wraps, `exists` false) and `ambiguousCommitProbeFailureRidesAsSuppressedOnTheCommitFailure` (commit failure is the cause, the probe failure is its one suppressed exception, store fails closed). | `AzureBlobAttachmentStorageTest`, both green |
| Factory promises https on `sasUrl` and does not check it | **Fixed.** Scheme checked case-insensitively before the builder runs, same value-free message, no cause. Probe: `azureBlobHttpSasUrlIsRejectedWithoutEchoingTheToken` feeds `http://…?sig=MARKER`, asserts the contract exception, `https` in the message, and no MARKER anywhere in the cause chain. The pre-existing `azure` (Files) branch has the same wording and the same gap; left untouched here as QA scoped it, follow-up below. | `RegisteringStorageFactory.azureBlob`, `RegisteringStorageFactoryTest` |
| No happy-path test for the `sasUrl` branch | **Fixed.** `azureBlobWithHttpsContainerSasUrlBuildsBlobAdapter` builds from an https container SAS URL and asserts the adapter type. It is the test that would have caught an over-eager scheme check. | `RegisteringStorageFactoryTest` |
| Assert the SAS `sig` is absent from the cause chain in the live tampered-signature test | **Fixed.** The test now extracts the tampered `sig` value and asserts it is absent from the full chain, raw and URL-decoded. Passed live: the service's `AuthenticationFailed` detail does not echo the signature. | `AzureBlobAttachmentStorageLiveTest.tamperedSignatureIsClassifiedAsWrongCredential` |
| Factory table says nothing about the rights `azure_blob` needs | **Fixed.** Paragraph under the table: `rw` (or `rcw`) SAS or Storage Blob Data Contributor, and the go-live probe validates write only. | `RegisteringStorageFactory` class javadoc |
| Probe coverage regressed relative to the Files sibling; proposed read check on the `.verify-` name | **Not done here, deliberately.** The issue specifies the Put Block probe and its `cw`-only acceptance; adding a `Get Blob Properties` read on the probe name would fail exactly that acceptance case. QA's design is sound (read allowed answers `BlobNotFound`, read denied answers `AuthorizationPermissionMismatch`, nothing committed either way) and I agree it restores the sibling's guarantee at zero cost. It needs the issue's acceptance amended, so it is a follow-up for a go, not a silent change. | `tsanetgit/Connect_SDK#71` acceptance, second bullet |
| Receipt caveat: QA's run used JDK 26 | Noted. My runs use the pinned JDK 21; same counts on both. | this comment |

## Pre-PR gate on the delta

Second and last gate run (the Coder cap is two): **CONCERNS**, both items answered rather than changed:

- *Confirm the BOM supplies `azure-storage-blob`.* It does: `azure-sdk-bom` 1.2.35 pins `azure-storage-blob` 12.30.0, which is what `mvn dependency:tree` resolves (QA's receipt above shows the same number), and the offline suite compiles and runs against it.
- *Document the `azure_blob` config keys.* The `RegisteringStorageFactory` javadoc table is the documented `storageProperties` schema, the same place the `s3`, `azure`, and `gcs` keys were documented in `tsanetgit/Connect_SDK#68`. There is no second config doc to update.

## Follow-ups this leaves open

- The `azure` (Files) `sasUrl` branch has the same missing https check as the one fixed here.
- The read-probe design above, if the acceptance on `tsanetgit/Connect_SDK#71` is amended.
- `skills/connect-sdk-deploy/SKILL.md` adapter list, and the `GoLiveVerifier` PASS wording, both from the PR body.

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass. All five hunks trace to prior QA notes: seam javadoc rewrite (opaque-name fix), factory https check + javadoc rights paragraph (security note), live-test sig-echo assertion + `fullChain` helper (SAS-echo note), two new offline tests for the previously-dead `failPutBlobWith`/`failCommittedBlockIdsWith` fields (untested-path note), and two new factory tests (happy-path + http-rejection).

**Correctness:**
The https guard (`RegisteringStorageFactory` L142–145): `sasUrl.regionMatches(true, 0, "https://", 0, 8)` — case-insensitive prefix match, length-8 region. `http://acct...` fails the match (`http://a` vs `https://` at position 4: `:` vs `s`) → throws the value-free message before the builder runs. `HTTPS://...` passes (case-insensitive). Correct. The happy-path test `azureBlobWithHttpsContainerSasUrlBuildsBlobAdapter` confirms the guard isn't over-eager.

`ambiguousCommitProbeFailureRidesAsSuppressedOnTheCommitFailure`: sets both `failCommitWith` and `failCommittedBlockIdsWith`, drives the staged path (`BLOCK_SIZE + 8`). Asserts the commit failure is the cause and the probe failure is its single suppressed — matches the fail-closed behavior the prior review traced through `resolveAmbiguousCommit`. Correct.

`putBlobFailureSurfacesThroughTheContractAndLeavesNothingVisible`: 16-byte payload → small-file `putBlob` path → `IllegalStateException` wrapped as `AttachmentStorageException` containing "put blob", `exists` false. Correct.

The live sig-echo assertion extracts the `sig` query value up to the next `&` and checks both raw and URL-decoded forms are absent from the full chain — closes the question the prior review raised without introducing a new claim.

**Security:**
The https check is the substantive change and it's right: cleartext SAS-token transport is now rejected at config time with a value-free message, MARKER-probed across the cause chain (`azureBlobHttpSasUrlIsRejectedWithoutEchoingTheToken`). No secrets in the diff — test values are `sig=Zm9v` (base64 "foo") and `MARKER` placeholders. The `fullChain` helper only concatenates class names and messages for assertion; not logged.

**Regression risk:**
`RegisteringStorageFactory.azureBlob` is the only production change; the added guard is a new early-throw inside the existing `sasUrl` branch, before the builder call — no other branch touched. `s3`/`azure`/`gcs` arms unchanged. The pre-existing `azure` (Files) branch carries the identical missing-https gap, correctly scoped out as a follow-up (it's not a regression this PR introduces).

**Carried forward:**
- Seam javadoc restates disproved "opaque" claim — RESOLVED
- `failPutBlobWith`/`failCommittedBlockIdsWith` declared but untested — RESOLVED
- Factory promises https, no scheme check — RESOLVED
- No happy-path test for `sasUrl` branch — RESOLVED
- Live tampered-sig test doesn't assert sig absent from chain — RESOLVED
- Factory table silent on required rights — RESOLVED (rights paragraph added)
- Probe coverage regressed vs Files sibling (read check) — FOLLOW-UP (needs issue acceptance amended)
- `azure` Files branch same https gap — FOLLOW-UP
- `skills/connect-sdk-deploy/SKILL.md` stale adapter list — FOLLOW-UP

**Recommendation:** Merge — every prior blocking/concern note is resolved with a cited test, and the three open items are legitimately post-merge follow-ups.

---

shawn-tsanet:
Follow-ups from this PR's review are filed as `tsanetgit/Connect_SDK#73` (all four: Files `sasUrl` https check, Blob probe read validation, deploy-skill adapter list, verifier PASS wording).

Issue #73, opened 2026-09-07
attachment-receiver: follow-ups from the Azure Blob adapter review (#72)


shawn-tsanet:
**For triage:** label `enhancement`; Project: `11-2026 Release` (project number 40). Parent: `tsanetgit/Connect_SDK#46`. Source: the review of `tsanetgit/Connect_SDK#72` (Azure Blob adapter, closed `tsanetgit/Connect_SDK#71`), which merged with these four items deliberately left out of the diff.

Four small, independent changes, one PR. Each is scoped to the file it names.

## 1. Enforce https on the Azure Files `sasUrl` branch

**Why.** `RegisteringStorageFactory.azure` rejects a malformed `sasUrl` with a message that says it "must be an https URL", but nothing checks the scheme. The SDK's `ShareClientBuilder.endpoint()` accepts `http://` and builds a working client whose SAS token, the credential, then travels in cleartext on every request. The service refuses it only while the account keeps "secure transfer required" on. `tsanetgit/Connect_SDK#72` closed the identical gap on the `azure_blob` branch (`regionMatches(true, 0, "https://", 0, 8)` before the builder runs, same value-free message, no cause); the Files branch was left untouched there because it was outside that PR's scope.

**Change.** The same scheme check on the `azure` branch, before `builder.endpoint(sasUrl)`.

**Acceptance.** A factory test feeding `http://…?sig=MARKER` asserts `AttachmentStorageException`, `https` in the message, and no `MARKER` anywhere in the cause chain; an https share SAS URL still builds `AzureFilesAttachmentStorage`.

## 2. Blob go-live probe validates read as well as write

**Why.** `AzureBlobAttachmentStorage.verifyAccess` stages one uncommitted block, per `tsanetgit/Connect_SDK#71`. That validates write only, while `exists` and the ambiguous-commit resolution are `Get Blob Properties` / `Get Block List`, which need read. A container SAS carrying only `cw` passes go-live and fails on the first `exists`; the Files sibling's write-read-delete probe would have failed it at go-live. Documented in the adapter and factory javadoc, but a documented regression is still a regression.

**Change.** After staging the probe block, call `Get Blob Properties` on the same `.verify-<uuid>` name. Read allowed answers `BlobNotFound` (404), which is the pass; read denied answers `AuthorizationPermissionMismatch` (403), classified as no permission. Nothing is committed either way, so the probe still leaves no blob behind.

[...]

PR #74, opened 2026-09-07
attachment-receiver: follow-ups from the Azure Blob adapter review (#73)

shawn-tsanet:
Closes `tsanetgit/Connect_SDK#73` (epic `tsanetgit/Connect_SDK#46`). The four items the review of `tsanetgit/Connect_SDK#72` left out of that diff, one PR, each scoped to the file it names.

**Account-hygiene rule, binding for this public repo:** no storage account, container, connection string, or subscription identifier appears here. Live results are counts and outcomes.

## The four changes

1. **Files `sasUrl` branch enforces https** (`RegisteringStorageFactory.azure`). The message already promised "an https URL"; nothing checked it, and `ShareClientBuilder` accepts `http://`, which would send the SAS token in cleartext on every request. Same case-insensitive scheme check the `azure_blob` branch got in #72, before the builder runs, same value-free message, no cause. Scheme only: the `$root` check stays a Blob concern because the two builders parse differently. Tests: `http://…?sig=MARKER` rejected with no `MARKER` anywhere in the cause chain, and an https share SAS URL still builds `AzureFilesAttachmentStorage`.
2. **Blob go-live probe validates read as well as write** (`AzureBlobAttachmentStorage.verifyAccess`). After staging the uncommitted block, a second try block reads the probe name's properties through the existing `sizeOrAbsent`. The answer, absent, is ignored: the probe tests permission, not state. A service refusal classifies through `classify(stage, e)`, so the operator reads "read denied" or "write denied". Nothing committed, nothing to delete, the seven-day expiry is untouched (a read does not move the `Put Block` clock). The javadoc that documented the gap in #72 now documents the fix, in the adapter class javadoc, the inline comment, and the factory note.
3. **`connect-sdk-deploy` skill** names all four adapters and both Azure backend ids (`azure` for Files, `azure_blob` for Blob) in its module table; the verified-date line moves to today. Nothing else in the skill changes. The installed copy under the skills directory is a symlink into this repo, so it follows the merge.
4. **`GoLiveVerifier` PASS wording** is adapter-neutral: "passed the storage access probe", in the message and the class javadoc. `GoLiveVerifierTest` asserts it. Repo-wide grep for `write-read-delete` after the change: one hit remains, the Blob adapter javadoc describing the *other* adapters' sentinel, which is still true.

## The case that flips

Before this PR a container SAS carrying only `cw` passed go-live and failed on the first `exists`. Now it fails go-live at the read stage. Proven live: the service answers the `Get Blob Properties` on the probe name with `AuthorizationPermissionMismatch` (403), which classifies as "no permission: read denied". A read-only SAS still fails at the write stage ("write denied"), and an `rw` SAS passes and leaves no committed blob, with the probe block enumerable only as uncommitted. This supersedes the `cw`-only acceptance line in `tsanetgit/Connect_SDK#71`, as #73 states.

## Rights parity, the check that gates the javadoc claim

The new javadoc says the probe validates "the same two rights the adapter needs at runtime". Sweep of every seam method the adapter calls (`grep -n 'container\.' AzureBlobAttachmentStorage.java` against `AzureBlobContainer`):

| seam method | REST operation | right |
|---|---|---|
| `putBlob` | Put Blob | write |
| `stageBlock` | Put Block | write |
| `commitBlockList` | Put Block List | write |
| `sizeOrAbsent` | Get Blob Properties | read |
| `committedBlockIdsOrAbsent` | Get Block List | read |

`SdkAzureBlobContainer` calls no delete and no list. `rw` is the complete runtime set, so an `rw` SAS that passes go-live cannot fail on a later operation for lack of a right.

## Verification

- Offline, `mvn -pl attachment-receiver -am test`: **174 run, 0 failures, 38 skipped** (was 170 / 0 / 37 at `main`). New: two factory tests for the Files https branch, one Blob test for read-denied-at-read-stage (with the double recording that the read hit the probe name), the verifier wording assertion, and one live test.
- Live, `AzureBlobAttachmentStorageLiveTest` against a disposable container in a private test account: **15 / 15**, about 46 s. The renamed `cwOnlySasIsClassifiedAsNoPermissionAtTheReadStage`, the new `rwSasPassesTheProbeAndLeavesNoCommittedBlob`, `readOnlySasIsClassifiedAsNoPermissionAtTheWriteStage`, and the fourteen cases from #72 unchanged.
- The final two-line wording commit (`3efeda7`) is javadoc and skill prose only; the affected test classes were rerun after it.

| claim | line |
|---|---|
| Probe validates read | `verifyAccess`, second try block calling `sizeOrAbsent(probeName)` |
| `cw`-only fails at the read stage | live `cwOnlySasIsClassifiedAsNoPermissionAtTheReadStage`; offline `verifyAccessClassifiesReadDeniedAsNoPermissionAtTheReadStage` |
| Still nothing committed | live `rwSasPassesTheProbeAndLeavesNoCommittedBlob` asserts the committed listing under `.verify-` is empty |
| The read hits the probe name | offline assertion on the double's `lastSizeOrAbsentName` |
| `rw` is the complete runtime rights set | the sweep table above |
| Files branch rejects `http://` value-free | `azureHttpSasUrlIsRejectedWithoutEchoingTheToken` |
| Verifier wording | `GoLiveVerifierTest.passesStorageAndSkipsCrmWhenUnconfigured` |

- **Guard vs representation.** The https check is a guard. The representation alternative, parsing the URL and requiring the scheme, is the same check with more code; the value-free rule forbids echoing the parsed parts anyway.
- **Third or later fix to one class.** This is the second https check in the factory (first was #72). Both branches now share the shape; a shared helper would be the third-time move.
- **Newly reachable failure.** The probe's read stage is a second network call that can fail on transient connectivity where there was none. It surfaces through `classify("read", …)` as the STORAGE check's FAIL line in `GoLiveVerifier`, the same handler as the write stage. Existing offline classification tests inject at `stageBlock` and their write-stage strings are byte-identical, so they still assert the correct stage without edits.
- **Abstraction / schema / stored shape.** None introduced or changed.

## Advisor and gate

Advisor consulted twice. Orientation: keep the `cw`-only live test flipped rather than deleted (it is the one case that discriminates old from new), thread the stage through every `classify` arm, treat the javadoc that documented the gap as part of item 2, keep item 1 scheme-only. Pre-done: the rights-parity sweep above was its one blocking check, plus the two wording trims in the second commit. Pre-PR gate: **LOOKS GOOD**, first run. Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
@@ -68,8 +68,8 @@ import java.util.UUID;
  * (tsanetgit/Connect-API-Code#140, question 2).
  *
  * <p><b>Go-live probe.</b> {@link #verifyAccess} stages one uncommitted block on a
- * {@code .verify-<uuid>} name (write), then reads that name's properties (read), expecting
- * {@code BlobNotFound}. Nothing is committed and nothing needs deleting, so it replaces
+ * {@code .verify-<uuid>} name (write), then reads that name's properties (read); the answer,
+ * absent, is ignored, since the probe tests permission, not state. Nothing is committed and nothing needs deleting, so it replaces
  * the other adapters' write-read-delete sentinel while validating the same two rights the
  * adapter needs at runtime: write for {@link #store}, read for {@link #exists} and the
  * ambiguous-commit resolution. Required rights are therefore SAS {@code rw} (or
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
@@ -68,13 +68,15 @@ import java.util.UUID;
  * (tsanetgit/Connect-API-Code#140, question 2).
  *
  * <p><b>Go-live probe.</b> {@link #verifyAccess} stages one uncommitted block on a
- * {@code .verify-<uuid>} name: it needs only write permission and leaves no committed blob,
- * replacing the other adapters' write-read-delete sentinel (this is the issue's design).
- * The consequence, stated rather than hidden: <b>the probe validates write only, while
- * operation needs read as well</b> — {@link #exists} and the ambiguous-commit probe are
- * {@code Get Blob Properties}. A SAS carrying only {@code cw} passes go-live and fails
- * {@code exists} at runtime. Required rights are therefore SAS {@code rw} (or {@code rcw})
- * on the container, or the Storage Blob Data Contributor role for an Entra identity.
+ * {@code .verify-<uuid>} name (write), then reads that name's properties (read), expecting
+ * {@code BlobNotFound}. Nothing is committed and nothing needs deleting, so it replaces
+ * the other adapters' write-read-delete sentinel while validating the same two rights the
+ * adapter needs at runtime: write for {@link #store}, read for {@link #exists} and the
+ * ambiguous-commit resolution. Required rights are therefore SAS {@code rw} (or
+ * {@code rcw}) on the container, or the Storage Blob Data Contributor role for an Entra
+ * identity. A SAS carrying only {@code cw} fails go-live at the read stage
+ * (tsanetgit/Connect_SDK#73; before it, such a SAS passed go-live and failed on the first
+ * {@code exists}).
  *
  * <p>Failure classification reads the service error code before the HTTP status, because
  * Azure Storage answers {@code AuthenticationFailed} with 403, not 401.
@@ -197,38 +199,46 @@ public final class AzureBlobAttachmentStorage implements AttachmentStorage {
         byte[] probe = "attachment-receiver go-live probe".getBytes(StandardCharsets.UTF_8);
         try {
             // One uncommitted block: proves write permission, commits nothing, and the
-            // service reclaims it in seven days. No read, no delete, nothing to clean up.
+            // service reclaims it in seven days. Nothing to delete.
             container.stageBlock(probeName, blockId(UUID.randomUUID().toString(), 0), probe, probe.length);
         } catch (RuntimeException e) {
-            throw new AttachmentStorageException(classify(e), e);
+            throw new AttachmentStorageException(classify("write", e), e);
+        }
+        try {
+            // Get Blob Properties on the same name proves read permission. The answer is
+            // "absent" (only uncommitted blocks exist) and is deliberately ignored: the
+            // probe tests permission, not state.
+            container.sizeOrAbsent(probeName);
+        } catch (RuntimeException e) {
+            throw new AttachmentStorageException(classify("read", e), e);
         }
     }
 
     /** Wrong-credential, no-permission, and wrong-target must read differently. */
-    private String classify(RuntimeException e) {
+    private String classify(String stage, RuntimeException e) {
         if (e instanceof BlobStorageException bse) {
             BlobErrorCode code = bse.getErrorCode();
             int status = bse.getStatusCode();
             if (BlobErrorCode.AUTHENTICATION_FAILED.equals(code)
                     || BlobErrorCode.INVALID_AUTHENTICATION_INFO.equals(code)
                     || BlobErrorCode.NO_AUTHENTICATION_INFORMATION.equals(code)) {
-                return "wrong credential: the container rejected the identity (verify write, " + code + ")";
+                return "wrong credential: the container rejected the identity (verify " + stage + ", " + code + ")";
             }
             if (BlobErrorCode.AUTHORIZATION_FAILURE.equals(code)
                     || BlobErrorCode.AUTHORIZATION_PERMISSION_MISMATCH.equals(code)
                     || BlobErrorCode.INSUFFICIENT_ACCOUNT_PERMISSIONS.equals(code)
                     || status == 403) {
-                return "no permission: write denied on container '" + containerName + "'"
+                return "no permission: " + stage + " denied on container '" + containerName + "'"
                         + (code == null ? "" : " (" + code + ")");
             }
             if (BlobErrorCode.CONTAINER_NOT_FOUND.equals(code) || status == 404) {
                 return "wrong target: container '" + containerName
-                        + "' (or its account) does not exist (verify write)";
+                        + "' (or its account) does not exist (verify " + stage + ")";
             }
-            return "verify write failed on container '" + containerName + "': " + code
+            return "verify " + stage + " failed on container '" + containerName + "': " + code
                     + " (HTTP " + status + ")";
         }
-        return "connectivity: cannot reach the container (verify write): " + e.getMessage();
+        return "connectivity: cannot reach the container (verify " + stage + "): " + e.getMessage();
     }
 
     private int fill(InputStream content, byte[] buffer, String name)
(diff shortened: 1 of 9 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `AzureBlobAttachmentStorage.java` class javadoc (lines 68–79) — rewrites the "validates write only" gap statement into the fix. Traces to item 2 (javadoc part).
- `AzureBlobAttachmentStorage.verifyAccess` (lines 199–215) — second try block reading `sizeOrAbsent(probeName)`, plus `classify` now takes a `stage` arg. Traces to item 2.
- `classify` message-string edits (five sites) — thread `stage` through every arm. Traces to item 2; these are required by the signature change, not incidental.
- `GoLiveVerifier.java` (javadoc + PASS message) — "write-read-delete" → "storage access probe". Traces to item 4.
- `RegisteringStorageFactory.java` javadoc (lines 61–66) — updates the blob probe description to "validates both". Traces to item 2's factory-note claim.
- `RegisteringStorageFactory.azure` (lines 110–123) — hoists `malformed` message to a variable and adds the https guard before `builder.endpoint`. Traces to item 1. The variable hoist is justified: the same message is now used in two throw sites.
- Live test rename + rw/cw/r reshuffle — traces to item 2's "case that flips".
- `AzureBlobAttachmentStorageTest.java` new read-denied test + probe-name assertion — traces to item 2 verification claims.
- `InMemoryAzureBlobContainer.lastSizeOrAbsentName` — traces to the "read hits the probe name" assertion.
- `GoLiveVerifierTest.java` wording assertion — traces to item 4.
- `SKILL.md` — module table gains both Azure backend ids, date bumps to 2026-09-07. Traces to item 3.

All hunks trace. No scope creep.

**Correctness:**

Trace of a `cw`-only SAS through the new path (the case that flips): `stageBlock` succeeds (write granted) → second try block calls `sizeOrAbsent(probeName)` → service returns 403 `AuthorizationPermissionMismatch` → caught as `RuntimeException`, `classify("read", e)` → `BlobStorageException` branch, status 403 → returns `"no permission: read denied on container '…' (…)"`. Matches the offline test assertion `contains("no permission: read denied")` and the live test. Correct.

The `stage` string is threaded through all five return paths in `classify` — auth-failed, permission, not-found, generic, and connectivity. The write-stage strings for the injected-at-`stageBlock` tests remain byte-identical (`"verify write"`), so pre-existing classification tests still assert the right stage without edits. Confirmed against the diff.

The read answer is ignored (`sizeOrAbsent` return value discarded), which is correct — the probe stages an uncommitted block, so `Get Blob Properties` on that name returns absent; the call is for permission, not the value. Nothing committed, so no cleanup and no clock reset. Consistent with the `rwSasPassesTheProbeAndLeavesNoCommittedBlob` assertion.

Item 1 guard: `sasUrl.regionMatches(true, 0, "https://", 0, 8)` — case-insensitive, length-8 prefix. `http://` (length 7) fails the match (index 4 = `:` ≠ `s`), rejected. `https://…` passes. Empty/blank already excluded by the enclosing `if (!blank(sasUrl))`. The rejection message contains `https` and does not embed the URL, so the `MARKER` token never reaches the cause chain — matches `azureHttpSasUrlIsRejectedWithoutEchoingTheToken`. Correct.

The rights-parity sweep in the PR body is the gating claim for the "same two rights" javadoc. I did not re-run the grep against `AzureBlobContainer` this session, so I take the table as the author's asserted sweep rather than an independently verified one — but the five listed seam methods and their read/write classification are internally consistent with the code shown, and no delete/list appears in the diff.

**Security:**

The item 1 fix is itself a security improvement — it stops a SAS token riding in cleartext over `http://`. The value-free message and the `MARKER`-absence test confirm the token is not echoed into the exception chain. No secrets, no injection, no credential logging introduced. None.

**Regression risk:**

`classify` is private to `AzureBlobAttachmentStorage`; the only consumers are the two call sites in `verifyAccess`, both updated. `GoLiveVerifier.storageCheck` calls `verifyAccess()` (unchanged signature) and only reads the PASS/FAIL message; the PASS wording change is asserted by `GoLiveVerifierTest`. The newly reachable failure — a transient read-stage network error — surfaces through the same STORAGE FAIL handler as the write stage, so no new unhandled path. The live `readOnlySasIsClassifiedAsNoPermissionAtTheWriteStage` still fails at write (read never reached), preserving that case. Item 1 touches only the Files `sasUrl` branch; the connectionString branch and the `azure_blob` branch are untouched. Low risk.

**Recommendation:** Merge — the diff is fully traced, the flipped case is correct and tested at both offline and live layers, and the one guard added is a genuine security fix.

---

shawn-tsanet (commented):
## QA review of attachment-receiver: follow-ups from the Azure Blob adapter review (#73)

Separate-session QA pass at head `3efeda7`, independent of the auto-QA comment above. This PR closes the four items my review of `tsanetgit/Connect_SDK#72` left for a follow-up; the rest of that review's notes landed in #72 itself (its disposition comment lists each with a receipt), so #73's scope is complete.

**Verdict:** ✅ Ready to merge

Receipts: `mvn -pl attachment-receiver -am test` in the local clone at `3efeda7`, whose merge-base with `origin/main` is main's head `613ff96`: **174 run, 0 failures, 0 errors, 38 skipped**, matching the PR body; per class, `AzureBlobAttachmentStorageTest` 30 / 0, `RegisteringStorageFactoryTest` 24 / 0, `GoLiveVerifierTest` 5 / 0. Same JDK 26 caveat as before. The live class (15 / 15 in the body) I could not re-run without container credentials.

### Minimal-Diff Rule
- Pass. Nine files, each traceable to one of the four numbered items in #73. The one hunk that is not literally an item, hoisting the Files branch's `malformed` message into a variable, is required because the message now has two throw sites.

### Correctness
- **Probe read stage.** Traced a `cw`-only SAS: `stageBlock` succeeds, `sizeOrAbsent(probeName)` is refused with 403 `AuthorizationPermissionMismatch`, `classify("read", e)` returns "no permission: read denied". Traced an `rw` SAS: the read answers `BlobNotFound`, `SdkAzureBlobContainer.sizeOrAbsent` maps only that code to `-1`, the value is discarded, nothing committed. The offline double records the name the read hit, and the test asserts it equals the staged probe name, so the read cannot silently target some other blob.
- **Rights parity.** The javadoc's "same two rights the adapter needs at runtime" is backed by the body's seam sweep; I re-ran the sweep on the PR head: `AzureBlobContainer` declares exactly the five methods in the table, and `grep -o 'container\.[a-zA-Z]*('` over the adapter finds calls to all five and nothing else, three writes and two reads, so `rw` is the complete runtime set.
- **Files `https` guard.** `regionMatches(true, 0, "https://", 0, 8)` runs before the builder; `http://` fails at index 4. Value-free, no cause, `MARKER` asserted absent across the full chain, and the new happy-path test builds from an https share SAS URL. This mirrors the `azure_blob` check #72 landed.
- `classify` threads `stage` through all five arms; the write-stage strings are byte-identical to before, so the pre-existing classification tests still assert the right stage without edits (confirmed by reading them against the diff).

### Security
- The Files `https` guard is itself the security item: a share SAS token no longer rides `http://`. No secrets in the diff; the only credential-shaped strings are `MARKER` and a base64 placeholder.

### Regressions
- `verifyAccess`'s signature is unchanged; `GoLiveVerifier.storageCheck` only reads PASS/FAIL, and the PASS wording change is asserted. The newly reachable failure, a transient error on the read stage, surfaces through the same STORAGE FAIL line as the write stage. The Files `connectionString` branch and the `azure_blob` branch are untouched.
- Repo-wide `git grep write-read-delete` on the PR head: two hits, not the one the body says. The Blob adapter javadoc (line 73) describing the other adapters' sentinel, which is still true, and the comment this PR adds in `GoLiveVerifierTest` (line 34), which is the correct use of the term. Nothing to change; the body's count is off by the comment it added.

### Operational
- `skills/connect-sdk-deploy/SKILL.md` now lists all four adapters and both Azure backend ids, with the verified-date line moved to today. The installed skill is a symlink into this repo, so the merge propagates it.
- No dependency, schema or LaunchAgent surface.

### Nits
- None.

PR #75, opened 2026-09-07
attachment-receiver: ambiguous-finish resolves by attempt identity in the S3 and GCS adapters (#69)

shawn-tsanet:
Closes `tsanetgit/Connect_SDK#69` (epic `tsanetgit/Connect_SDK#46`). Disclosed as a known parity caveat in `tsanetgit/Connect_SDK#68`; the Azure Blob adapter (`tsanetgit/Connect_SDK#72`) shipped the identity approach first, this brings S3 and GCS to the same guarantee.

**Account-hygiene rule, binding for this public repo:** no account, bucket, or credential identifier appears here.

## The gap, and the option taken

Both adapters resolved an ambiguous finalize (the client-side `CompleteMultipartUpload` or resumable `finish` fails after the server may have committed) with a probe the wrong object could satisfy: on S3, "abort answers `NoSuchUpload` and the key is visible"; on GCS, "the object is visible at exactly the byte count written". Because `store` overwrites a same-name object, an older object of the same size plus a finalize that never committed read as success while the stale bytes stayed.

Of the three options in the issue, this is **option 1, identity by a per-attempt marker**. Each multipart or resumable attempt is stamped with a UUID as object metadata (`tsanet-upload-attempt`) at initiation, which both services store on the completed object, and the probe requires the visible object's marker to equal this attempt's. Why not the others: option 2 (generation or ETag changed) still reports success when a *concurrent* writer committed instead of us, whereas identity fails closed there too; option 3 (checksum) breaks silently on SSE-KMS buckets, where the S3 ETag is not the content hash. Identity also mirrors the Blob adapter's block-list resolution, so all three adapters now answer the ambiguous case the same way.

## What changed

- **S3.** The UUID goes on `CreateMultipartUpload` as user metadata. `resolveAmbiguousComplete` keeps the abort step (still the cleanup call for a live session) and, on `NoSuchUpload`, requires a HEAD of the key to return this attempt's marker. The new `attemptMarkerOrAbsent` keeps `existsKey`'s 404 discrimination: an absent key is a definite "not this attempt", not a probe error. The key constant is lowercase because S3 returns user-metadata keys lowercased.
- **GCS.** The seam widens: `startResumable` carries the marker as custom metadata on the `BlobInfo` the session opens with; `attemptMarkerOrAbsent` reads it back with both null guards (absent object, no metadata); `sizeOrAbsent` stays for `exists`. `resolveAmbiguousFinish` compares identity and drops the size comparison. `SdkGcsBucket` stays logic-free.
- **Single-request objects stay unmarked on purpose** (`PutObject`, `create`). A later attempt's ambiguous finalize over one then finds no marker and fails closed, which is right. Both javadocs say so, and a test per adapter makes it true.
- **Doubles.** `StubS3Client` lowercases metadata keys at initiation, commits them on both commit branches (including commit-despite-failure), and returns them from `HeadObject`. `InMemoryGcsBucket` records the marker on both finish branches. Those are what keep the existing committed-ambiguous tests green with no edits.
- **Javadoc, three files.** The S3 policy paragraph, the GCS bullet including its "cannot distinguish a concurrent same-name writer" parenthetical (now false: it fails closed), and the Blob adapter's cross-reference that said S3 and GCS resolve by size.

## Verification

- Offline, `mvn -pl attachment-receiver -am test`: **182 run, 0 failures, 40 skipped** (was 174 / 0 / 38 at `main`; the extra skips are the new S3 and GCS live tests without their env vars). Four new tests, two per adapter: an older same-size object plus a finalize that never committed (S3 variant with abort answering `NoSuchUpload`, the expired-session shape), and an older *unmarked* single-request object plus the same. Each asserts the store throws and the older bytes are untouched. The existing committed-ambiguous tests are unchanged and green.
- **Red probe.** With the two old predicates temporarily restored (`existsKey(key)` on S3, `sizeOrAbsent(key) == total` on GCS), three tests fail with "expected `AttachmentStorageException`, nothing thrown": both S3 tests and the GCS same-size test (QA re-ran the probe and counted the same three). The GCS unmarked test stays green under the old predicate because its 64-byte older object already differs in size from a multi-buffer attempt; it pins the unmarked policy rather than the mechanism. The patch was then re-applied. The tests discriminate the defect, not a proxy.
- **S3, run live.** `S3AttachmentStorageLiveTest` against a disposable bucket in a private test account: **9 / 9**. The 7 shared-contract tests, abort-after-partial-commit, and a new test, `completedMultipartObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot`, which is the fact this design rests on observed against the service: user metadata given at `CreateMultipartUpload` comes back from `HeadObject` on the completed object under the lowercase key, and a `PutObject` object carries no marker. The doc sentences it confirms, from the S3 multipart overview read today: "If you want to provide metadata describing the object being uploaded, you must provide it in the request to initiate the multipart upload" and "Amazon S3 associates that metadata with the object."
- **GCS, not run live, stated plainly.** There are no GCS credentials on the build machine. The propagation fact is cited from the GCS resumable-upload guide (the initial POST of the session carries the object metadata) and the SDK surface confirmed by `javap` (`BlobInfo.Builder.setMetadata(Map)`, `BlobInfo.getMetadata()`). The env-gated mirror test `completedResumableObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot` is now in `GcsAttachmentStorageLiveTest`, so the follow-up is "set `GCS_CONTRACT_TEST_BUCKET` and run".

## Observable side effect

Every multipart or resumable object now permanently carries the `tsanet-upload-attempt` metadata entry. It is a non-secret UUID; anything downstream that enumerates object metadata will see it.

| claim | line |
|---|---|
| Same-size older object fails closed | `ambiguousCompleteOverAPreExistingSameSizeObjectThrows`, `ambiguousFinishOverAPreExistingSameSizeObjectThrows` (both red-probed) |
| Unmarked single-request object fails closed | `ambiguousCompleteOverAnUnmarkedSingleRequestObjectThrows`, `ambiguousFinishOverAnUnmarkedSingleRequestObjectThrows` |
| Committed-ambiguous still succeeds | existing `ambiguousCompleteThatActuallyCommittedIsReportedAsSuccess`, `ambiguousFinishThatActuallyCommittedIsReportedAsSuccess`, unchanged |
| Marker is on the completed object | S3: live `completedMultipartObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot`; GCS: vendor doc above; the doubles model it on both commit branches |
| S3 keys come back lowercased | constant is lowercase; the stub lowercases at initiation so stub and live agree |
| Absent key is not a probe error | `attemptMarkerOrAbsent` 404 branch returns null, same as `existsKey` |

- **Guard vs representation.** This *is* the representation change: the probe now compares an identity the attempt owns instead of a property (size, visibility) any object can have.
- **Third or later fix to one class.** First fix; the caveat was disclosed in #68 and tracked as #69.
- **Newly reachable failure.** The marker read is a new HEAD (S3) or get (GCS) on the ambiguous path only; failures ride as suppressed on the finalize failure and surface as `AttachmentStorageException` toward the ingest path, the same handler as before.
- **Abstraction / schema / stored shape.** `GcsBucket` gained a parameter and a method; both implementations are in the diff and the seam is package-private with no other consumers (`grep -rn 'startResumable\|GcsBucket' src` shows only the adapter, the SDK implementation, and the double). Object metadata is additive; pre-existing objects have no marker and are treated as "not this attempt", which is the fail-closed outcome.

## Advisor and gate

Advisor consulted twice. Orientation: confirmed option 1 over 2 and 3 with the reasons above, keep the S3 abort as cleanup, lowercase the S3 key and make the stub lowercase too, commit the marker on both commit-despite-failure branches, update the GCS concurrent-writer parenthetical. Pre-done: required the red probe to be observed rather than expected, and the two unmarked-object tests to make the javadoc claim true; both done. Pre-PR gate: **LOOKS GOOD**, first run. Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/azure/AzureBlobAttachmentStorage.java
@@ -53,10 +53,9 @@ import java.util.UUID;
  *       truncated object — and there is deliberately no cleanup round-trip that could
  *       itself fail (the GCS adapter's shape, unlike S3's abort call).</li>
  *   <li><b>Ambiguous commit, resolved exactly.</b> A {@code Put Block List} that fails
- *       client-side may have committed server-side. The S3 and GCS adapters resolve that
- *       by size, which a same-size overwrite or a concurrent writer can fool
- *       (tsanetgit/Connect_SDK#69). Here the block ids carry a UUID minted for this store
- *       call, so the resolution is identity, not size: the blob's committed block list
+ *       client-side may have committed server-side. As in the S3 and GCS adapters since
+ *       tsanetgit/Connect_SDK#69, the resolution is identity, not size. Here the block ids
+ *       carry a UUID minted for this store call: the blob's committed block list
  *       ({@code Get Block List}) equal to the ids this call staged means this commit
  *       landed, and nothing else can produce that list. Any other shape throws.</li>
  * </ul>
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsAttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsAttachmentStorage.java
@@ -36,11 +36,15 @@ import java.util.UUID;
  *       {@code abandon()}ed, not finished — finishing a partial upload would finalize a
  *       truncated object. There is deliberately no cleanup round-trip that could itself
  *       fail, unlike S3's {@code AbortMultipartUpload}.</li>
- *   <li><b>Ambiguous finish.</b> A {@code finish} that fails client-side may have committed
- *       server-side. When the object is then visible at exactly the byte count written, the
- *       store is treated as committed and reported as success, mirroring the S3 adapter's
- *       ambiguous-complete policy. (Like both twins, this cannot distinguish a concurrent
- *       same-name writer's object.)</li>
+ *   <li><b>Ambiguous finish, resolved by identity.</b> A {@code finish} that fails
+ *       client-side may have committed server-side. Each resumable attempt is stamped with
+ *       a UUID as custom metadata ({@value #ATTEMPT_METADATA_KEY}) on the {@code BlobInfo}
+ *       the session opens with, which GCS stores on the finalized object. The visible
+ *       object carrying this attempt's UUID means this finish landed; a previous object of
+ *       the same name (same size or not), a concurrent writer's, or nothing at all fails
+ *       closed and throws (tsanetgit/Connect_SDK#69). Single-request objects written by
+ *       {@code create} are deliberately unmarked: a later attempt's ambiguous finish over
+ *       one then fails closed, which is right.</li>
  * </ul>
  *
  * <p>Required IAM on the bucket/prefix: {@code storage.objects.create},
@@ -54,6 +58,9 @@ public final class GcsAttachmentStorage implements AttachmentStorage {
 
     /** Stream read granularity; a 256 KiB multiple, so it is also a valid resumable chunk size. */
     static final int BUFFER_SIZE = 5 * 1024 * 1024;
+
+    /** Custom-metadata key carrying the per-attempt UUID on resumable objects. */
+    static final String ATTEMPT_METADATA_KEY = "tsanet-upload-attempt";
 
     private final GcsBucket bucket;
     private final String bucketName;
@@ -94,9 +101,10 @@ public final class GcsAttachmentStorage implements AttachmentStorage {
     private StoredAttachment resumable(IncomingAttachment attachment, InputStream content,
                                        String key, byte[] firstBuffer)
             throws AttachmentStorageException {
+        String attempt = UUID.randomUUID().toString();
         GcsBucket.Upload upload;
         try {
-            upload = bucket.startResumable(key, attachment.contentType());
+            upload = bucket.startResumable(key, attachment.contentType(), attempt);
         } catch (RuntimeException e) {
             throw wrap("resumable start", key, e);
         }
@@ -132,21 +140,22 @@ public final class GcsAttachmentStorage implements AttachmentStorage {
         try {
             upload.finish();
         } catch (RuntimeException finishFailure) {
-            return resolveAmbiguousFinish(key, total, finishFailure);
+            return resolveAmbiguousFinish(key, attempt, total, finishFailure);
         }
         return new StoredAttachment(key, total);
     }
 
     /**
-     * A failed finish is ambiguous: GCS may have committed server-side. The object visible
-     * at exactly the byte count written means exactly that — report the commit as the
-     * success it was. Every other shape throws.
+     * A failed finish is ambiguous: GCS may have committed server-side. The visible object
+     * carrying this attempt's UUID means exactly that — report the commit as the success
+     * it was. Another marker or none (a previous same-name object, a concurrent writer's,
+     * nothing committed) fails closed. Every other shape throws.
      */
-    private StoredAttachment resolveAmbiguousFinish(String key, long total,
+    private StoredAttachment resolveAmbiguousFinish(String key, String attempt, long total,
                                                     RuntimeException finishFailure)
             throws AttachmentStorageException {
         try {
-            if (bucket.sizeOrAbsent(key) == total) {
+            if (attempt.equals(bucket.attemptMarkerOrAbsent(key))) {
                 return new StoredAttachment(key, total);
             }
         } catch (RuntimeException probeFailure) {
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsBucket.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/GcsBucket.java
@@ -29,12 +29,20 @@ interface GcsBucket {
      * Large-file path: start a resumable upload. The object does not become visible until
      * {@link Upload#finish()}; an {@link Upload#abandon() abandoned} upload never produces a
      * visible object and the session expires server-side, which is how the SPI's
-     * no-partial-visibility rule holds without an abort call.
+     * no-partial-visibility rule holds without an abort call. {@code attemptMarker} is
+     * stored as custom metadata on the finalized object, so the adapter can later tell
+     * this attempt's object from any other.
      */
-    Upload startResumable(String key, String contentType);
+    Upload startResumable(String key, String contentType, String attemptMarker);
 
     /** The object's byte size, or {@code -1} when it does not exist; throws on access error. */
     long sizeOrAbsent(String key);
+
+    /**
+     * The visible object's attempt marker, or {@code null} when the object is absent or
+     * carries none (a {@link #create single-request} object); throws on access error.
+     */
+    String attemptMarkerOrAbsent(String key);
 
     byte[] download(String key);
 
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/SdkGcsBucket.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/gcs/SdkGcsBucket.java
@@ -9,6 +9,7 @@ import com.google.cloud.storage.StorageException;
 
 import java.io.IOException;
 import java.nio.ByteBuffer;
+import java.util.Map;
 
 /**
  * The logic-free SDK implementation of {@link GcsBucket}: every method is one or two SDK
@@ -28,12 +29,14 @@ final class SdkGcsBucket implements GcsBucket {
 
     @Override
     public void create(String key, String contentType, byte[] bytes) {
-        storage.create(blobInfo(key, contentType), bytes);
+        storage.create(blobInfo(key, contentType, null), bytes);
     }
 
     @Override
-    public Upload startResumable(String key, String contentType) {
-        WriteChannel channel = storage.writer(blobInfo(key, contentType));
+    public Upload startResumable(String key, String contentType, String attemptMarker) {
+        // The BlobInfo opens the resumable session, and its metadata lands on the object
+        // when the session finalizes.
+        WriteChannel channel = storage.writer(blobInfo(key, contentType, attemptMarker));
         channel.setChunkSize(GcsAttachmentStorage.BUFFER_SIZE);
         return new Upload() {
             @Override
@@ -70,6 +73,15 @@ final class SdkGcsBucket implements GcsBucket {
     }
 
     @Override
+    public String attemptMarkerOrAbsent(String key) {
+        Blob blob = storage.get(BlobId.of(bucket, key));
+        if (blob == null || blob.getMetadata() == null) {
+            return null;
+        }
+        return blob.getMetadata().get(GcsAttachmentStorage.ATTEMPT_METADATA_KEY);
+    }
+
+    @Override
     public byte[] download(String key) {
         return storage.readAllBytes(BlobId.of(bucket, key));
     }
@@ -79,10 +91,13 @@ final class SdkGcsBucket implements GcsBucket {
         storage.delete(BlobId.of(bucket, key));
     }
 
-    private BlobInfo blobInfo(String key, String contentType) {
+    private BlobInfo blobInfo(String key, String contentType, String attemptMarker) {
         BlobInfo.Builder builder = BlobInfo.newBuilder(BlobId.of(bucket, key));
         if (contentType != null) {
             builder.setContentType(contentType);
+        }
+        if (attemptMarker != null) {
+            builder.setMetadata(Map.of(GcsAttachmentStorage.ATTEMPT_METADATA_KEY, attemptMarker));
         }
         return builder.build();
     }
--- a/attachment-receiver/src/main/java/com/tsanet/receiver/storage/s3/S3AttachmentStorage.java
+++ b/attachment-receiver/src/main/java/com/tsanet/receiver/storage/s3/S3AttachmentStorage.java
@@ -17,6 +17,7 @@ import java.io.InputStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
+import java.util.Map;
 import java.util.Objects;
 import java.util.UUID;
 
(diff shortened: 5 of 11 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `AzureBlobAttachmentStorage.java` javadoc — updates the cross-reference that claimed S3/GCS resolve by size. Traces to "Javadoc, three files" and is now factually required (the old text is false after this PR). ✓
- `GcsAttachmentStorage.java` javadoc + `ATTEMPT_METADATA_KEY` constant + threading `attempt` through `resumable`/`resolveAmbiguousFinish` — the core GCS identity change. Traces to "GCS" section. ✓
- `GcsBucket.java` interface — `startResumable` gains `attemptMarker`, new `attemptMarkerOrAbsent`. Traces to "the seam widens". ✓
- `SdkGcsBucket.java` — `blobInfo` gains marker param, new `attemptMarkerOrAbsent`. Traces to "SdkGcsBucket stays logic-free". ✓
- `S3AttachmentStorage.java` javadoc + constant + `metadata(Map.of(...))` on create + `attempt` threading + new `attemptMarkerOrAbsent`. Traces to "S3" section. ✓
- Test files (`GcsAttachmentStorageTest`, `S3AttachmentStorageTest`, `InMemoryGcsBucket`, `StubS3Client`) — new tests and double plumbing. Trace to "Doubles" and "Verification". ✓

Every hunk traces. No incidental tidying spotted.

**Correctness:**

Traced the S3 same-size false-positive path (`ambiguousCompleteOverAPreExistingSameSizeObjectThrows`):
1. First store of `older` (size `PART_SIZE + 5` > 5 MiB) goes multipart, stamps a UUID, commits, `objectMetadata["01234567/same-size.bin"]` = attempt-A's marker.
2. Second store: new UUID attempt-B on `createMultipartUpload`, `pendingMetadata[uploadId]` = attempt-B. Complete fails (`SdkClientException`), `completeCommitsDespiteFailure=false` so no commit — `objectMetadata` still holds attempt-A. Abort throws `NoSuchUpload`. `attemptMarkerOrAbsent(key)` HEADs the key, returns attempt-A's marker. `attempt.equals(...)` is `"B".equals("A")` → false → falls through to `throw wrap("complete", ...)`. Older bytes untouched. ✓ Matches the assertion.

The unmarked-object path: `PutObject` removes metadata, HEAD returns `Map.of()`, `.get(ATTEMPT_METADATA_KEY)` returns null, `attempt.equals(null)` false → throws. ✓

One correctness note on `attemptMarkerOrAbsent` in S3 (line ~208): if `headObject` throws a non-404 `S3Exception`, it's wrapped and rethrown as `AttachmentStorageException`. In `resolveAmbiguousComplete` this is caught by `catch (Exception headFailure)` and the original `completeFailure` is thrown instead — so a probe error surfaces as the finalize failure, not the marker-read failure. That is the pre-existing behavior with `existsKey` and is the correct fail-closed outcome. No new issue.

Committed-ambiguous still-succeeds path relies on the doubles committing metadata on the commit-despite-failure branch. `StubS3Client.commit` writes `objectMetadata` unconditionally; `InMemoryGcsBucket.commit()` writes `attemptMarkers` on both branches. Verified the double models both branches. ✓

**Security:**

None. The marker is a random UUID (non-secret, disclosed as such). Metadata key is a fixed constant, values are UUIDs — no injection surface. Metadata goes through the SDK's `metadata(Map)` / `setMetadata(Map)` typed paths, not string-built. No credentials in the diff (account-hygiene rule held).

**Regression risk:**

- `GcsBucket.startResumable` signature changed (positional contract). Consumers: `SdkGcsBucket` (updated in diff) and `InMemoryGcsBucket` (updated in diff). PR claims `grep -rn 'startResumable\|GcsBucket' src` shows only adapter, SDK impl, and double. I did not re-run the grep this session, but the interface is package-private and the diff updates both implementors; a missed consumer would fail compilation, and the PR reports a green build (178 run). Low risk.
- `resolveAmbiguousComplete` / `resolveAmbiguousFinish` are private — no external consumers.
- New `attemptMarkerOrAbsent` on the interface: `sizeOrAbsent` is retained for `exists`, so no behavior on the existing-check path changed.
- Pre-existing objects carry no marker → treated as "not this attempt" → fail closed. This is a behavior change from the old size/visibility probe (some previously-succeeding ambiguous finalizes over pre-existing objects would now throw), but that is exactly the defect being fixed, and it fails closed (safe direction). ✓
- Observable side effect: every multipart/resumable object now carries `tsanet-upload-attempt` metadata. Disclosed. Anything enumerating object metadata downstream will see it — additive, non-secret.

**Recommendation:** Merge — the identity change is correct, traces cleanly, fails closed on every off-path, and the one unverified claim (the grep for `GcsBucket` consumers) is backstopped by the green compile; note the PR is still marked Draft and the live confirmation remains the stated follow-up.

---

shawn-tsanet:
Second commit adds a live S3 test for the fact the identity resolution rests on, and the S3 live class ran **9 / 9** against a disposable bucket: metadata given at `CreateMultipartUpload` comes back from `HeadObject` on the completed object under the lowercase key, and `PutObject` objects stay unmarked. PR body updated; GCS remains doc-cited (no credentials).

shawn-tsanet (commented):
## QA review of attachment-receiver: ambiguous-finish resolves by attempt identity in the S3 and GCS adapters (#69)

Separate-session QA pass at head `0bdfe1e`, independent of the auto-QA comment above (which reviewed `386acf5`, before the live S3 test was added).

The fix is correct and the discriminating tests are real; I reproduced the red probe myself. Two test-adequacy notes are worth a follow-up, neither holds the merge.

Receipts: `mvn -pl attachment-receiver -am test` in the local clone at `0bdfe1e`, whose merge-base with `origin/main` is main's head `7e662d3`: **179 run, 0 failures, 0 errors, 39 skipped**, matching the PR body; per class, `S3AttachmentStorageTest` 16 / 0, `GcsAttachmentStorageTest` 19 / 0, `AzureBlobAttachmentStorageTest` 30 / 0. Same JDK 26 caveat as the previous reviews. **Red probe re-run by me** in a scratch worktree of the PR head: with `existsKey(key)` restored on S3 and `sizeOrAbsent(key) == total` restored on GCS, `mvn test` on the two adapter classes fails **3** tests, each with "Expected AttachmentStorageException to be thrown, but nothing was thrown": both S3 tests (same-size and unmarked) and the GCS same-size test. The GCS unmarked test stays green under the old predicate because its older object is 64 bytes against a multi-buffer attempt, so size alone already discriminates it there; it pins the unmarked policy rather than the mechanism. The body says "exactly the two same-size tests failed"; it is three, the stronger result. The S3 live claim (9 / 9, marker on the completed object under the lowercase key) I could not re-run without bucket credentials; GCS is doc-cited, as the body states plainly.

### Minimal-Diff Rule
- Pass. Ten files, every hunk traceable to #69: the two resolvers and their seams, the doubles that model the marker on both commit branches, four offline tests, one live S3 test, and the three javadoc paragraphs that were false or incomplete after the change (S3 policy, GCS bullet with its concurrent-writer parenthetical, the Blob cross-reference).

### Correctness
- **S3 path traced.** The UUID goes on `CreateMultipartUpload` as user metadata. On a failed complete, abort runs first (still the cleanup for a live session); only `NoSuchUpload` proceeds to the HEAD, and only `attempt.equals(marker)` reports success. `null` from an absent key or an unmarked object fails `equals`, so a previous same-name object, a `PutObject` object, or an expired session with nothing committed all throw, and the older bytes are asserted untouched. The 404 discrimination in `attemptMarkerOrAbsent` mirrors `existsKey`, so an absent key is a definite "not this attempt", and any other `S3Exception` or `SdkException` becomes a probe failure that rides as suppressed on the original complete failure, the pre-existing fail-closed shape.
- **GCS path traced.** The marker rides on the `BlobInfo` that opens the resumable session and is read back with both null guards (absent blob, null metadata map). `sizeOrAbsent` stays for `exists`, so the existing-check path is untouched; the only `failSizeOrAbsent` test is `existsThrowsWhenTheBackendCannotAnswer`, which still tests what it says. `create` passes `null` metadata, so single-request objects are unmarked and a later ambiguous finish over one fails closed.
- **Doubles model the real services on the branch that matters.** `StubS3Client` lowercases metadata keys at initiation and commits them on both the normal and commit-despite-failure branches; `InMemoryGcsBucket` records the marker in one `commit()` used by both branches and drops it on `create`. That is why the pre-existing committed-ambiguous tests stay green with no edits, and the red probe shows the new tests exercise the resolvers rather than the doubles.
- **Untested branch on both resolvers: the marker read itself failing.** `InMemoryGcsBucket` gains `failAttemptMarker` (line 27) and honors it (line 94), but no test sets it, so `resolveAmbiguousFinish`'s `probeFailure` catch is never exercised. On S3 the only `failHeadWith` uses are in `exists404IsFalseButOtherErrorsThrow` (lines 204 and 209), the `exists` path, so `attemptMarkerOrAbsent`'s non-404 branches and the resolver's `headFailure` catch are likewise untested. The S3 gap predates this PR (the old `existsKey` probe had the same), the GCS hook is new and unused. One test per adapter in the shape of the Blob adapter's `ambiguousCommitProbeFailureRidesAsSuppressedOnTheCommitFailure` (from #72's disposition) closes both: probe throws, store fails closed, the finalize failure is the cause and the probe failure its suppressed exception.
- Option choice is sound: identity fails closed on a concurrent writer where a generation/ETag check would report success, and it does not depend on the ETag being a content hash (SSE-KMS). The body and both javadocs say "fails closed" for the concurrent case, not "resolved", which is the accurate claim: a writer that overwrites between our commit and our probe still makes us throw.

### Security
- The marker is a random UUID in typed metadata paths (`metadata(Map)`, `setMetadata(Map)`), not string-built. No credentials in the diff. No new IAM: the S3 HEAD needs `s3:GetObject`, already required; the GCS get needs `storage.objects.get`, already required for `exists`.

### Regressions
- `GcsBucket.startResumable` gained a parameter: the seam is package-private and both implementors are in the diff; a missed consumer would not compile, and the build is green. `resolveAmbiguousComplete` / `resolveAmbiguousFinish` are private.
- Happy paths unchanged: store, exists and verifyAccess do not touch the marker except to write it at initiation.
- Behavior change, disclosed and in the safe direction: objects stored before this PR carry no marker, so an ambiguous finalize over one now throws where it might previously have reported success. Every multipart or resumable object now carries the `tsanet-upload-attempt` metadata entry permanently; non-secret, additive, stated in the body.

### Operational
- No dependency, schema or LaunchAgent surface.
- **The GCS propagation fact can be pinned now, not only when credentials arrive.** `GcsAttachmentStorageLiveTest` already exists and is env-gated. A mirror of the S3 `completedMultipartObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot` test added there costs nothing today, is skipped like the rest of the class, and turns the stated follow-up into "set the env var and run" instead of "write a test and run".

### Nits
- Body accuracy: the red probe fails three tests, not two (above). Nothing to change in code.

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass. The single new hunk adds `completedMultipartObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot` plus its three imports (`ByteArrayInputStream`, `assertNotNull`, `assertNull`) to the S3 live test — traces directly to "S3, run live" in the PR body.

**Correctness:**
The new test asserts the design's load-bearing fact against the service: a multipart store (`PART_SIZE + 1024` bytes forces multipart) HEADs back a non-null, well-formed UUID under `ATTEMPT_METADATA_KEY`; a 16-byte `PutObject` store HEADs back null. `UUID.fromString(marker)` throws on a malformed marker, so it also pins the value shape, not just presence. Live-gated like the rest of the class (skipped without the bucket env var — consistent with the +1 skip count the PR body reports). Correct.

**Security:**
None. Marker is a UUID read back through the typed `.metadata()` map; no credentials, no injection surface.

**Regression risk:**
None. Additive test only; no production code touched in this delta.

**Carried forward:**
- S3 grep for GcsBucket consumers — FOLLOW-UP (backstopped by green compile; interface is package-private).
- Marker-read-failure branch untested (both resolvers; `failAttemptMarker`/`failHeadWith` unused on the probe path) — FOLLOW-UP.
- GCS propagation not pinned live (mirror test could be added env-gated now) — FOLLOW-UP.
- Body says red probe fails two tests, actually three — FOLLOW-UP (prose only, no code change).

**Recommendation:** Merge — this delta only adds the live S3 propagation test the prior reviews wanted; nothing new introduced, prior findings all follow-ups.

---

shawn-tsanet:

Addressed in the third commit, pushed. Offline after it: **182 run, 0 failures, 40 skipped** (two new offline tests, one new env-gated live test).

|---|---|---|
| Marker read failing is untested on both resolvers (`failAttemptMarker` unused; `failHeadWith` only on the `exists` path) | **Fixed.** `ambiguousCompleteProbeFailureRidesAsSuppressedOnTheCompleteFailure` (S3: HEAD answers 500, store fails closed, the complete failure is the cause, the wrapped `attempt marker read` failure rides as suppressed alongside `NoSuchUpload`) and `ambiguousFinishProbeFailureRidesAsSuppressedOnTheFinishFailure` (GCS: same shape, one suppressed exception). | both green offline |
| GCS propagation can be pinned now with an env-gated mirror test | **Fixed.** `completedResumableObjectCarriesThisAttemptsMarkerAndSingleRequestObjectsDoNot` in `GcsAttachmentStorageLiveTest`, skipped without `GCS_CONTRACT_TEST_BUCKET` like the rest of the class. The follow-up is now "set the env var and run". | 40 skips, up from 39 |
| Body says the red probe fails two tests; it fails three | **Fixed in the body.** Three, with the note that the GCS unmarked test pins the policy rather than the mechanism, as QA observed. | PR body, Verification |
| Auto-QA could not re-run the `GcsBucket` consumer grep | **Receipt.** `grep -rn 'startResumable\|GcsBucket\b' attachment-receiver/src --include='*.java'` hits five files only: the adapter, the seam, the SDK implementation, the double, and the offline test. All in the diff. | this comment |
| JDK 26 caveat on QA's run | Noted; my runs use the pinned JDK 21, same counts. | this comment |

Issue #76, opened 2026-09-07
connect-library: V2 attachment client (grant, upload, complete, abandon) for the direct delivery path


shawn-tsanet:
**For triage:** label `enhancement`; Project: `09-2026 Release` (project number 39). Serves: `tsanetgit/Connect-API-Code#147` (the V2 grant and complete contract). Parent: none; standalone under the release project. Related: `tsanetgit/Connect-API-Code#148` (the MongoDB mock this is developed against), `tsanetgit/Connect-API-Code#141` (false success on V1 forward), `tsanetgit/Connect_SDK#46` (the receive-side epic; unrelated code, same design vocabulary).

## Why

The September direction for attachments is the direct path: the sender uploads straight into the receiver's store under instructions the platform returns, and the platform seals and verifies the upload. Nothing passes through the core or through Cloudflare, so file size is bounded by the receiver's store, not by ingress. `tsanetgit/Connect-API-Code#147` defines the sender's side of that as four calls on `/v2/collaboration-requests/{token}/attachments`: **grant** (declare one file, receive an `upload` block), **upload** (execute the block verbatim in its `mode`: `single`, `multipart`, `resumable`, or `relay`), **complete** (return the provider's part receipts and the final size), and **abandon**. The September plan (v0.2, 2026-09-04) makes the Connect SDK the reference implementation of those calls, so that a connector adopts V2 by upgrading `connect-library`, and so that a member writing its own client has a working implementation to read. The plan sets the client's completion at 2026-09-16.

The contract is a draft until `tsanetgit/Connect-API-Code#147` is agreed; the mock in `tsanetgit/Connect-API-Code#148` already implements the receiver side of it (single up to 5,000,000,000 bytes, multipart above, `Content-Length` signed, same-name 409, grant TTL, sweep), which is what this client develops and tests against.

## Change

New in `connect-library`, alongside the existing V1 `AttachmentsFacade` (which stays; V1 is deprecated in the contract, not removed):

- **`AttachmentsV2Facade`** with `grant(token, GrantRequest)`, `upload(Grant, source)`, `complete(token, grantId, CompleteRequest)`, `abandon(token, grantId)`, and a convenience `send(token, file)` that runs all of them and returns the recorded outcome. `source` is a `Path` or an `InputStream` plus declared size; the facade never holds more than one part in memory.
[...]

PR #77, opened 2026-09-07
connect-library: V2 attachment client (grant, upload, complete, abandon) with demo and console (#76)

shawn-tsanet:
Closes `tsanetgit/Connect_SDK#76`. Serves `tsanetgit/Connect-API-Code#147` (the V2 grant and complete contract, a draft until agreed). The receiver-side stand-in this was designed against is `tsanetgit/Connect-API-Code#148`; nothing of that receiver's interface appears here, this repo being public.

## What this adds

The sender's side of the V2 attachment contract, so a connector adopts the direct delivery path by upgrading `connect-library`, and a member writing its own client has a working implementation to read.

**Library.** `AttachmentsV2Facade` on `TsaNetApiSession` (`attachmentsV2()`): `grant`, `upload`, `complete`, `abandon`, and `send`, which runs the flow. Behind it:

- `ConnectApiAttachmentsV2Api` makes the three core calls through the generated `ApiClient.invokeAPI` with the `bearerAuth` scheme, so the bearer supplier and the 401 re-auth interceptor apply exactly as they do to V1. The V2 paths are not in the OpenAPI spec yet, which is the only reason the DTOs are hand-written records; they mirror the `#147` schemas field for field and sit in one package for the swap to generated classes.
- `AttachmentUploadExecutor` executes the grant's `upload` block verbatim on the JDK `HttpClient`, all four modes. It is not on `RestTemplate` because the library's template wraps a `BufferingClientHttpRequestFactory`, which would hold a whole part in memory. `Content-Length` is a restricted header in the JDK client, so it is filtered from the pass-through and guaranteed by a `FileRegionPublisher` with an exact `contentLength`; a body without one goes chunked and a store that signs the length refuses it. The wire header is the value the grant carried.
- `ConnectApiAttachmentsV2Gateway` owns the flow rules: abandon on any upload failure (best effort, the upload failure stays primary), complete retried on transient failure because it is idempotent on the grant id, one re-grant on a complete-time `attachment/grant-expired`, and the platform's recorded outcome returned unchanged.
- `AttachmentV2Exception` is value-free by construction: status, problem type, and the platform's own title and detail. Every record carrying a URL or headers has a redacted `toString`, and tests assert no URL, host or signature reaches a message.

**Demo.** A "Deliver directly (V2)" form on the case page. The file is handed to the SDK on a virtual thread; the page polls a state endpoint (login required, like every case endpoint) and shows the grant's mode, parts done of total, bytes sent of total (byte-level while a single or relay body streams, per part otherwise), then the platform's outcome in the platform's words. The facade is resolved on the request thread so an account switch mid-flight cannot redirect a delivery; the browser's file name is reduced to its last segment before it becomes a temp path; the multipart cap is raised to 20 GB so a large file reaches the SDK at all.

**Console.** `deliver-attachment` with the same progress and outcome lines; `help` lists it through the registry.

**Doc.** `docs/attachments-v2-client.md`: one full sequence per mode for members implementing the three calls without the SDK, shapes taken from the SDK's test double. The deploy skill's facade list gains `attachmentsV2()`.

## Policies worth knowing

| Situation | What the client does |
|---|---|
| I/O failure, 5xx, 429 on a signed URL | retry, three attempts, exponential backoff with jitter, `Retry-After` honored and capped |
| any other 4xx, including a 403 from an expired signed URL | terminal; abandon, throw; the caller decides whether to run `send` again |
| relay PUT | retried only when the connection never opened; once the receiver answered, never re-sent |
| resumable 308 without `Range` | counts as no progress and the chunk is resent; a session that never reports `Range` fails closed after the budget. The contract draft does not pin chunk semantics; this is the fail-closed reading and it is stated in the doc |
| resumable chunk answered 5xx | re-PUT with the same `Content-Range`; if the session had committed before failing the re-PUT is refused terminally. Fine while `#147` keeps chunk commits transactional |
| complete answers 5xx or connectivity | retried up to three times with backoff (idempotent on `grantId`) |
| complete answers `attachment/grant-expired` | one fresh grant, one more upload; a second expiry throws |
| `verification.mode` is `none` | the platform records `DELIVERED_UNVERIFIED`; the client returns that word unchanged and never branches on the mode |

## Verification

- Reactor, `mvn test`: **connect-library 131 run, 0 failures, 1 skipped** (35 new offline tests and the one env-gated live test; the rest unchanged), **attachment-receiver 182 / 0 / 40**, **demo-ui 15 / 0**, **console 33 / 0**, **demo 6 / 0**. The demo script passes a syntax check.
- `ConnectApiAttachmentsV2ApiTest` (9): the three calls through the real `ApiClient` converters against `MockRestServiceServer`, plus a real `RestTemplate` against a closed port asserting a connectivity failure never echoes the case token that rides in the URL. Golden JSON both ways with strict matching, `Authorization` and `Idempotency-Key` on the wire, optional fields omitted, single-mode complete carrying only the size, problem details mapped to a typed exception, a non-problem error body yielding the status alone with nothing echoed, a grant without an upload block rejected.
- `AttachmentUploadExecutorTest` (17) against the JDK's own `HttpServer`: headers received verbatim, wire `Content-Length` equal to the declared or part size, part order and bytes, ETags forwarded intact, a 503 part retried with a fresh body, a 403 terminal with one attempt and no URL in the message, budget exhaustion, a target that never answers surfacing as `UPLOAD_UNREACHABLE` through the facade exception, relay not retried after a 503, the three-chunk resumable flow, resumable recovery after a dropped connection (`bytes */total` query, resend), unsupported mode and part-size mismatch before any request, the publisher reading exactly its region, `Retry-After` honored and capped, relay retrying when the connection never opens, a resumable session that never reports `Range` failing closed, a 5xx resumable chunk re-PUT with the same `Content-Range`, duplicate part numbers rejected, and byte-level progress on a single upload.
- `ConnectApiAttachmentsV2GatewayTest` (10) with the seam and executor mocked: the flow rules above, digest computed only on request and forwarded on grant and complete, login required, missing file rejected before granting.
- **Live coverage, stated plainly.** The core's V2 endpoints do not exist yet, so grant, complete and abandon have no live target; they are proven through the real client converters against the mock server. The upload executor has an env-gated live test that reads an `AttachmentGrantDTO` envelope from a file (`ATTACHMENTS_V2_GRANT_FILE`: grant, source path, receipts output) and executes it against real signed URLs; whatever issued the grant seals and verifies on its side from the receipts. That split keeps every receiver's interface out of this repository. Not run in this PR: no grant issuer is wired to this machine yet.

| claim | line |
|---|---|
| wire `Content-Length` equals the declared size, never chunked | `FileRegionPublisher.contentLength()`; executor test asserts the header and no `Transfer-Encoding` |
| headers sent verbatim, restricted ones filtered | `passThrough`; single-PUT test asserts custom headers present and `Host` rewritten |
| 403 terminal, one attempt | `sendWithRetry` non-transient branch; `aRejectedPartIsNeverRetried…` |
| I/O exhaustion surfaces as `UPLOAD_UNREACHABLE` | `execute()` boundary catch; `aTargetThatNeverAnswers…` |
| relay never re-sent after an answer | `RetryScope.CONNECT_ONLY` excluded from the transient-status branch; relay test |
| 308 without `Range` is no progress | `committedOffset(response, offset)` plus the stall counter; `aResumableSessionThatNeverReportsRangeFailsClosed` |
| a 5xx resumable chunk is re-PUT with the same `Content-Range` | `TRANSIENT_STATUS_ONLY` keeps transient statuses in the retry branch; `aResumableChunkAnsweredFiveHundredIsRePut…` |
| a connectivity failure names a cause type, never the URL | `invoke`'s `ResourceAccessException` branch; `aConnectivityFailureNeverEchoesTheCaseToken…` |
| abandon on upload failure, failure stays primary | `uploadOrAbandon`; two gateway tests |
| one re-grant on grant-expired | `send`; `aGrantExpiredOnCompleteRegrantsExactlyOnce`, `aSecondGrantExpiredIsNotRetriedAgain` |
| outcome is the platform's | `send` returns `api.complete`'s result; gateway test returns `FAILED` unchanged |
| no URL or token in any message | redacted `toString`s; `doesNotContain("SECRET")` assertions in both test classes |

- **Positional contract.** `DefaultTsaNetApiSession`'s constructor gained a parameter. Consumers by command, `grep -rn 'new DefaultTsaNetApiSession('`: the runtime and one test, both updated. `TsaNetApiSession` gained a method; implementors by command, `grep -rln 'implements TsaNetApiSession'`: the default and account-scoped sessions, both updated.
- **Vocabulary gained a member.** The facade list in the deploy skill's reference now names `attachmentsV2()`; the console `help` output is generated from the registry. Repo grep for other enumerations of facades or commands: none.
- **Guard vs representation.** The demo's filename sanitization is a guard on client input; the representation alternative (never using the client name at all) would lose the name the partner sees. The `Content-Length` handling is a representation change: the length lives on the body publisher instead of a header.
- **Newly reachable failures.** Every new failure surfaces as `AttachmentV2Exception` toward the caller; the demo renders it as a failed state, the console as an error line. The one pre-existing catch-all in the console command still catches anything else.
- **Schema / stored shape.** None. No SQLite table, no new dependency; the JDK client and its test server are in the platform.
- **Abstraction introduced.** `AttachmentsV2Api` centralizes the three core calls; there was no old form to sweep, the endpoints are new.
- **URL scheme on upload URLs, a decision rather than a check.** The executor accepts what the platform returns, `http` included, because the contract's rule is verbatim execution and the URLs are per-grant, short-lived and issued by the authenticated platform, unlike the operator-pasted SAS URLs this repo rejects on `http` in `#72` and `#74`. Stated in the executor javadoc.

## Gate dispositions (CONCERNS, first run)

- *File size read twice, once for the grant and once in the executor.* Noted. A file that changes between the two reads fails at the platform's signed-length enforcement or at complete's size check; the executor's own part-sum check is against the file it is about to send, which is the property that matters for the PUT. No change.
- *`cancel()` closing the channel while a read is in flight.* Noted. The outcome is a spurious `onError` after cancel, which the JDK client tolerates. No change; revisit if flakiness ever appears.
- *The 20 GB multipart cap applies to every demo form, not only V2.* Correct, and stated here. The V1 form benefits equally; the demo has no other multipart endpoint.

- **Pre-existing in the demo, security-adjacent:** the V1 `forward` endpoint resolves `file.getOriginalFilename()` under its temp directory without the last-segment reduction this PR gives the V2 endpoint. Same one-line fix; out of this PR's scope.
- Replace the hand-written records and the `AttachmentsV2Api` seam with generated classes once `#147` lands in `openapi.yaml`.
- Live gateway coverage (grant, complete, abandon) once the core endpoint exists; wire a grant issuer to the env-gated executor test in the private harness.
- The demo's in-memory upload state map is unbounded; fine for a demo, a TTL if it ever hosts more than demos.

## Advisor and gate

Advisor consulted twice. Orientation: implement resumable rather than defer it, put the progress callback in the facade, the JDK client is mandatory not optional, the `Content-Length` restricted-header trap, relay connect-only retry, one PR. Pre-done: found the internal I/O signal escaping the executor for three of four modes (fixed at the `execute` boundary, with a test that drops every attempt) and the demo's unsanitized filename (fixed), and asked for the 308-without-`Range` reading to be resolved fail-closed rather than from memory (done, stated above). Pre-PR gate: **CONCERNS** on the first run, dispositions above; second and last run on the QA delta: **CONCERNS** with two notes answered in the disposition comment (the adversarial one-byte-progress stall, and the implementer set, which `git grep -ln "implements TsaNetApiSession"` settles at two, both in the diff). Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CollaborationRequestAttachmentDeliverExecutor.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/CollaborationRequestAttachmentDeliverExecutor.java
@@ -1,0 +1,84 @@
+package com.tsanet.facade.cli;
+
+import com.tsanet.api.TsaNetApiSession;
+import com.tsanet.api.attachments.v2.AttachmentCompleteResult;
+import com.tsanet.api.attachments.v2.AttachmentV2Exception;
+import com.tsanet.api.connectapi.dto.CollaborationRequestStatusDto;
+import java.io.IOException;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.Arrays;
+import java.util.List;
+import java.util.Scanner;
+import org.springframework.stereotype.Component;
+
+/**
+ * {@code deliver-attachment}: one file to the partner on the direct path (V2). Prints the
+ * grant's mode and per-part progress as the upload runs, then the platform's recorded
+ * outcome, in the platform's own words.
+ */
+@Component
+public class CollaborationRequestAttachmentDeliverExecutor {
+    private final TsaNetApiSession session;
+    private final CollaborationRequestResolver requestResolver;
+
+    public CollaborationRequestAttachmentDeliverExecutor(
+        TsaNetApiSession session,
+        CollaborationRequestResolver requestResolver
+    ) {
+        this.session = session;
+        this.requestResolver = requestResolver;
+    }
+
+    public void execute(String[] args, Scanner scanner, CliRunContext cliRunContext) {
+        requestResolver.requireAuthentication();
+        CollaborationRequestStatusDto request = requestResolver.resolve(args);
+
+        List<String> files = CliArgs.files(args);
+        if (files.size() != 1) {
+            System.out.println(EntityPrinter.error(cliRunContext, "Provide exactly one --file PATH (one grant delivers one file)"));
+            return;
+        }
+        Path file = Path.of(files.get(0));
+        if (!Files.isRegularFile(file)) {
+            System.out.println(EntityPrinter.error(cliRunContext, "Attachment path is not a regular file: " + file));
+            return;
+        }
+        String description = CliArgs.description(args).orElse(null);
+        boolean sha256 = Arrays.asList(args).contains("--sha256");
+        String contentType = probeContentType(file);
+
+        System.out.println(EntityPrinter.info(cliRunContext, "Delivering " + file.getFileName()
+            + " to the partner on request id=" + request.id() + " token=" + request.token()));
+        try {
+            AttachmentCompleteResult outcome = session.attachmentsV2().send(
+                request.token(),
+                file,
+                contentType,
+                description,
+                sha256,
+                progress -> System.out.println(EntityPrinter.info(cliRunContext, progress.mode() + ": "
+                    + progress.partsDone() + "/" + progress.partsTotal() + " parts, "
+                    + progress.bytesSent() + " of " + progress.bytesTotal() + " bytes"))
+            );
+            System.out.println(EntityPrinter.info(cliRunContext, "Platform outcome: " + outcome.status()
+                + (outcome.verification() != null && outcome.verification().method() != null
+                    ? " (verified by " + outcome.verification().method() + ")" : "")
+                + (outcome.noteId() != null ? ", case note " + outcome.noteId() : "")
+                + (outcome.message() != null ? " - " + outcome.message() : "")));
+        } catch (AttachmentV2Exception ex) {
+            System.out.println(EntityPrinter.error(cliRunContext, "Delivery failed: " + ex.getMessage()));
+        } catch (IllegalArgumentException | IllegalStateException ex) {
+            System.out.println(EntityPrinter.error(cliRunContext, ex.getMessage()));
+        }
+    }
+
+    private static String probeContentType(Path file) {
+        try {
+            String probed = Files.probeContentType(file);
+            return probed == null ? "application/octet-stream" : probed;
+        } catch (IOException e) {
+            return "application/octet-stream";
+        }
+    }
+}
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/commands/DeliverAttachmentCommand.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/cli/commands/DeliverAttachmentCommand.java
@@ -1,0 +1,40 @@
+package com.tsanet.facade.cli.commands;
+
+import com.tsanet.facade.cli.CliRunContext;
+import com.tsanet.facade.cli.CollaborationRequestAttachmentDeliverExecutor;
+import com.tsanet.facade.cli.EntityPrinter;
+import java.util.Scanner;
+import org.springframework.stereotype.Component;
+
+@Component
+public class DeliverAttachmentCommand implements Command {
+    private final CollaborationRequestAttachmentDeliverExecutor deliverExecutor;
+    private final CliRunContext cliRunContext;
+
+    public DeliverAttachmentCommand(
+        CollaborationRequestAttachmentDeliverExecutor deliverExecutor,
+        CliRunContext cliRunContext
+    ) {
+        this.deliverExecutor = deliverExecutor;
+        this.cliRunContext = cliRunContext;
+    }
+
+    @Override
+    public String name() {
+        return "deliver-attachment";
+    }
+
+    @Override
+    public String description() {
+        return "Deliver one file to the partner on the direct path, V2 (--id/--token, --file PATH, --description TEXT, --sha256)";
+    }
+
+    @Override
+    public void execute(String[] args, Scanner scanner) {
+        try {
+            deliverExecutor.execute(args, scanner, cliRunContext);
+        } catch (Exception ex) {
+            System.out.println(EntityPrinter.error(cliRunContext, "Failed: " + ex.getMessage()));
+        }
+    }
+}
--- a/connect-library/src/main/java/com/tsanet/api/TsaNetApiSession.java
+++ b/connect-library/src/main/java/com/tsanet/api/TsaNetApiSession.java
@@ -1,6 +1,7 @@
 package com.tsanet.api;
 
 import com.tsanet.api.facade.AttachmentsFacade;
+import com.tsanet.api.facade.AttachmentsV2Facade;
 import com.tsanet.api.facade.AuthFacade;
 import com.tsanet.api.facade.CaseNotesFacade;
 import com.tsanet.api.facade.CaseResponsesFacade;
@@ -25,4 +26,7 @@ public interface TsaNetApiSession {
     PartnersFacade partners();
 
     AttachmentsFacade attachments();
+
+    /** The V2 direct-delivery attachment client (tsanetgit/Connect-API-Code#147). */
+    AttachmentsV2Facade attachmentsV2();
 }
--- a/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentCompleteRequest.java
+++ b/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentCompleteRequest.java
@@ -1,0 +1,25 @@
+package com.tsanet.api.attachments.v2;
+
+import com.fasterxml.jackson.annotation.JsonInclude;
+import java.util.List;
+
+/**
+ * Body of {@code POST .../grants/{grantId}/complete}: what the client uploaded, from the
+ * {@code AttachmentCompleteRequestDTO} schema (tsanetgit/Connect-API-Code#147). Empty
+ * collections and nulls are omitted on the wire, so a single-mode complete carries only
+ * {@code sizeBytes}.
+ *
+ * @param sizeBytes the byte count actually sent
+ * @param sha256    the digest supplied on grant, if any
+ * @param parts     multipart mode only: the receipts the provider returned, in part order
+ */
+@JsonInclude(JsonInclude.Include.NON_EMPTY)
+public record AttachmentCompleteRequest(long sizeBytes, String sha256, List<PartReceipt> parts) {
+    public AttachmentCompleteRequest {
+        parts = parts == null ? List.of() : List.copyOf(parts);
+    }
+
+    /** An ETag (S3), block id (Azure) or the provider's equivalent, forwarded exactly as received. */
+    public record PartReceipt(int partNumber, String receipt) {
+    }
+}
--- a/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentCompleteResult.java
+++ b/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentCompleteResult.java
@@ -1,0 +1,40 @@
+package com.tsanet.api.attachments.v2;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.time.OffsetDateTime;
+import java.util.UUID;
+
+/**
+ * The platform's recorded outcome for a grant, from the {@code AttachmentCompleteResultDTO}
+ * schema (tsanetgit/Connect-API-Code#147). {@code status} is the platform's word, never the
+ * client's: {@code DELIVERED}, {@code DELIVERED_UNVERIFIED}, {@code FAILED} or {@code EXPIRED}.
+ * A client that uploaded every byte without error still reports whatever this says.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record AttachmentCompleteResult(
+    UUID grantId,
+    String fileName,
+    String status,
+    VerificationResult verification,
+    Long noteId,
+    String message
+) {
+    public static final String DELIVERED = "DELIVERED";
+    public static final String DELIVERED_UNVERIFIED = "DELIVERED_UNVERIFIED";
+    public static final String FAILED = "FAILED";
+    public static final String EXPIRED = "EXPIRED";
+
+    /** True only for a platform-verified delivery. */
+    public boolean delivered() {
+        return DELIVERED.equals(status);
+    }
+
+    /** True when the sender reported completion and the platform could not verify arrival. */
+    public boolean deliveredUnverified() {
+        return DELIVERED_UNVERIFIED.equals(status);
+    }
+
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public record VerificationResult(String method, OffsetDateTime verifiedAt, Boolean sizeMatched, Boolean checksumMatched) {
+    }
+}
--- a/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentGrant.java
+++ b/connect-library/src/main/java/com/tsanet/api/attachments/v2/AttachmentGrant.java
@@ -1,0 +1,69 @@
+package com.tsanet.api.attachments.v2;
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties;
+import java.time.OffsetDateTime;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+
+/**
+ * A grant: permission plus upload instructions for one file on one case, from the
+ * {@code AttachmentGrantDTO} schema of the V2 contract draft (tsanetgit/Connect-API-Code#147).
+ * The client executes {@link Upload} verbatim and never branches on the receiver.
+ *
+ * <p>{@code toString} on every record that carries a URL or headers is redacted: signed URLs
+ * and the headers block are credentials, and a record's default {@code toString} would put
+ * them into any log line or exception message that interpolates the grant.
+ */
+@JsonIgnoreProperties(ignoreUnknown = true)
+public record AttachmentGrant(
+    UUID grantId,
+    String fileName,
+    OffsetDateTime expiresAt,
+    Receiver receiver,
+    Upload upload,
+    Verification verification
+) {
+    @Override
+    public String toString() {
+        return "AttachmentGrant[grantId=" + grantId + ", fileName=" + fileName + ", expiresAt=" + expiresAt
+            + ", receiver=" + receiver + ", upload=" + upload + ", verification=" + verification + "]";
+    }
+
+    /** What the platform echoes about the receiver so a client can fail fast. */
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public record Receiver(Long companyId, String targetKind, Long maxSizeBytes) {
+    }
+
+    /**
+     * The upload instructions. {@code mode} is one of {@code single}, {@code multipart},
+     * {@code resumable}, {@code relay}; {@code headers} are sent verbatim on every request.
+     */
+    @JsonIgnoreProperties(ignoreUnknown = true)
+    public record Upload(String mode, String method, String url, Map<String, String> headers, List<Part> parts) {
+        public Upload {
+            headers = headers == null ? Map.of() : Map.copyOf(headers);
+            parts = parts == null ? List.of() : List.copyOf(parts);
+        }
+
+        @Override
+        public String toString() {
+            return "Upload[mode=" + mode + ", method=" + method + ", url=<redacted>, headers="
+                + headers.keySet() + ", parts=" + parts.size() + "]";
+        }
+    }
+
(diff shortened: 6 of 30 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**

Walking the hunks — all trace to the PR body:
- `TsaNetApiSession` / `AccountScopedTsaNetApiSession` / `DefaultTsaNetApiSession` / `TsaNetApiRuntime` — the `attachmentsV2()` positional/interface addition. Traces to "Library. `AttachmentsV2Facade` on `TsaNetApiSession`". Constructor param + one test updated; consumer commands are in the PR body.
- `AttachmentGrant*`, `AttachmentComplete*`, `UploadProgress*`, `UploadReceipts`, `AttachmentV2Exception` — hand-written DTOs. Traces to "the DTOs are hand-written records."
- `AttachmentUploadExecutor`, `AttachmentsV2Api`, `ConnectApiAttachmentsV2Api`, `ConnectApiAttachmentsV2Gateway` — the four flow layers, all named in the body.
- `CollaborationRequestAttachmentDeliverExecutor` / `DeliverAttachmentCommand` — console `deliver-attachment`. Traces.
- `AttachmentsController` / `app.js` / `index.html` / `application.yml` — demo form + 20GB cap. Traces.
- `docs/attachments-v2-client.md` / `consume-artifact.md` — doc. Traces.
- Test files — traces.

No untraced hunks. The 20GB multipart cap is repo-wide, not V2-scoped — the PR flags this explicitly as a disposition, so it's disclosed, not creep.

**Correctness:**

Traced the `send` happy path (multipart, sha256): `grant` → `uploadOrAbandon` → `completeWithRetry`. Digest computed once via `sha256Of` before grant, forwarded on both grant and complete (gateway test `sendComputesTheDigestOnlyWhenAskedAndForwardsItOnGrantAndComplete` asserts this). Receipts flow from executor ETags into `AttachmentCompleteRequest`. Correct.

Two real concerns, neither blocking:

1. **`ConnectApiAttachmentsV2Gateway.send` size/grant divergence on grant-expired re-grant** (line ~`send`): the re-grant path reuses the *same* `request` object built from the first `sizeOf(file)` read. If the file changed between the two uploads, the second grant carries the stale `sizeBytes` but the executor re-reads the file. This is the "file size read twice" disposition, and the platform's signed-length enforcement catches it — acceptable, and disclosed. No change needed.

2. **`resumable` stall counter reset semantics** (`resumable`, `sendWithRetry` with `TRANSIENT_STATUS_ONLY`): a 5xx on a resumable chunk is *not* retried inside `sendWithRetry` (scope excludes 5xx from the transient branch, `TRANSIENT_STATUS_ONLY` → `throw rejected`). Trace: a resumable chunk answering 503 → `sendWithRetry` returns neither 2xx nor 308 → `throw rejected(what, 503)` — an `AttachmentV2Exception`, *not* `UploadIoFailure`. So it escapes the `try/catch (UploadIoFailure)` in `resumable` and terminates the whole upload. The PR's policy table says "resumable chunk answered 5xx → re-PUT with the same Content-Range." I don't see that re-PUT in the code — a 5xx on a chunk is terminal, not retried. This is a **gap between the stated policy and the code**, and there's no test exercising a 5xx-on-resumable-chunk path (the resumable tests cover 308 flow and dropped-connection, not 5xx). Worth confirming: is the policy-table row aspirational, or is the terminal-on-5xx behavior intended? If intended, the doc/table overstates. Flagging as a follow-up, not a blocker — terminal-on-5xx fails closed, which is safe.

Edge case checked: `committedOffset` with malformed `Range` returns `fallback` → treated as no progress → stall counter increments → fails closed after `MAX_ATTEMPTS`. Correct and matches the doc.

`FileRegionPublisher` cancel-during-read: disclosed disposition (spurious `onError` tolerated by JDK client). Reasonable.

**Security:**

No secrets in the diff. Redaction is thorough and *tested*: `AttachmentGrant.Upload`/`Part` have redacted `toString`, and `aRejectedPartIsNeverRetriedAndTheMessageCarriesNoUrl` + `aTargetThatNeverAnswers...` assert `doesNotContain("SECRET")` and `doesNotContain("127.0.0.1")` against URLs that literally carry `?sig=SECRET`. `ConnectApiAttachmentsV2Api.translate` echoes only platform `type`/`title`/`detail`, never request content — `aNonProblemErrorBodyYieldsTheStatusAlone` asserts no `SECRET` leaks from an HTML error body.

Demo filename sanitization: `Path.of(originalFilename).getFileName().toString()` reduces to last segment before temp-path use. Correct guard. The pre-existing V1 `forward` endpoint lacking the same reduction is called out as an out-of-scope follow-up — agreed, that's a real security-adjacent gap but not introduced here.

`RESTRICTED_HEADERS` filtering prevents header injection of reserved names; the JDK client rewrites `Host`. No injection surface in the URL path (path params go through `ApiClient.invokeAPI`, not string concat).

**Regression risk:**

- `DefaultTsaNetApiSession` constructor gained a positional param. Consumers by command (`new DefaultTsaNetApiSession(`): runtime + `DefaultTsaNetApiSessionCreateRequestTest`, both updated in the diff. Confirmed the test hunk adds the mock at the correct position (after `ConnectApiAttachmentsGateway`, matching constructor order).
- `TsaNetApiSession` gained `attachmentsV2()`. Implementors (`implements TsaNetApiSession`): `DefaultTsaNetApiSession` (via `implements TsaNetApiSession, AuthFacade, ...`) and `AccountScopedTsaNetApiSession`, both updated. Confirmed both hunks present.
- New gateway is additive — no existing endpoint, DTO, or facade method modified. V1 attachments untouched.
- Demo `AttachmentsController`: new endpoints (`/v2`, `/api/uploads/{id}`) additive; existing `getConfig` unchanged. The 20GB multipart cap change is app-wide and affects the V1 form too — disclosed, benign for a demo.
- `deliveries` ExecutorService with `@PreDestroy shutdownNow()` — clean lifecycle. The unbounded `uploads` map is a disclosed demo follow-up.

No mechanical detector fired; positional and vocabulary edges are enumerated in the PR body with commands. Verified the claims match the diff.

**Recommendation:** Merge — confirm whether the resumable-5xx policy-table row matches the terminal-on-5xx code (either add the re-PUT+test or correct the doc) as a post-merge follow-up; nothing here is wrong runtime behavior on the tested paths, and the PR is a draft anyway.

---

shawn-tsanet (commented):
## QA review of connect-library: V2 attachment client (grant, upload, complete, abandon) with demo and console (#76)

Separate-session QA pass at head `0b99033`.

**Verdict:** ❌ Needs changes

One item blocks, and it is one line: the Connect API connectivity branch echoes the case token into the exception message, contradicting that class's own javadoc. Everything else below is a note; nothing else would hold the merge.

Receipts: `mvn test` in the local clone at `0b99033`, whose merge-base with `origin/main` is main's head `2788927`: **connect-library 124 run, 0 failures, 1 skipped**, attachment-receiver 182 / 0 / 40, demo-ui 15 / 0, console 33 / 0, demo 6 / 0, all matching the body. Per class: `ConnectApiAttachmentsV2ApiTest` 8, `AttachmentUploadExecutorTest` **11**, `ConnectApiAttachmentsV2GatewayTest` 10. Same JDK 26 caveat as the previous reviews. The live executor test is the one skip.

### Blocking
1. **The case token reaches the exception message on any Connect API connectivity failure.** `ConnectApiAttachmentsV2Api.invoke`, the `catch (ResourceAccessException e)` branch, interpolates `e.getMessage()`. Probe, with the module's own classpath and Spring's real `RestTemplate` against a dead port:

   ```
   getMessage(): I/O error on POST request for
     "http://127.0.0.1:1/v2/collaboration-requests/CASETOKEN-MARKER/attachments/grants": Connection refused
   contains token marker: true
   ```

   Spring formats that message with the expanded URL, and the V2 paths carry `{token}`. That string becomes the `AttachmentV2Exception` message, which the console prints (`"Delivery failed: " + ex.getMessage()`) and the demo stores as `state.error` and renders in the browser. It contradicts two statements in this diff: the class javadoc's "nothing from the request is echoed", and `AttachmentV2Exception`'s "The message is value-free by construction ... never a signed URL, a header value or a token". The adjacent `RestClientException` branch already does the right thing with `e.getClass().getSimpleName()`; the same treatment here is the fix. The branch has no test, which is why the existing `SECRET`-absence assertions did not catch it. Worth adding a marker test in the shape the factory branches use.

### Minimal-Diff Rule
- Pass. Thirty files, all traceable to issue 76: the DTO package, the API seam and its Connect implementation, the executor, the gateway, the facade and its three wiring points, tests, the demo form and controller, the console command, the member doc, and the one-line facade list in the deploy skill.
- Positional contract verified rather than taken from the body: `git grep -ln "implements TsaNetApiSession"` returns exactly the default and account-scoped sessions, both updated; `git grep -n "new DefaultTsaNetApiSession("` returns the runtime and one test, both updated.

### Correctness
- **Executor traced and sound.** The restricted-header filter matches the JDK's reserved set, and the length is guaranteed by `FileRegionPublisher.contentLength()` rather than a header, which is what keeps a store that signs the length from refusing the PUT. Each retry re-subscribes the publisher and opens a fresh channel, so the "fresh body on retry" claim holds. The 403-is-terminal, relay-not-retried-after-an-answer, and resumable offset paths (including the `bytes */total` query and the `Range` parse) all read correctly, and the wire assertions in `AttachmentUploadExecutorTest` are real: a live `HttpServer` records headers, bytes and order.
- **The client never overrides the platform's outcome**, which is the right reading of the no-false-success rule. That makes one row of the body's policy table imprecise: "`verification.mode` is `none` → reported as `DELIVERED_UNVERIFIED`" is something the *platform* does. No client code branches on `verification.mode`, and it should not. Reword the row rather than the code.
- **`completeWithRetry` retries three times with no backoff**, unlike every other retry in the PR. A 502 from the Connect API produces three immediate calls.
- **Multipart trusts the grant's part numbering.** The parts are sorted and their sizes summed against the file, but duplicate `partNumber` values with sizes that happen to total correctly would upload the wrong regions and report duplicate receipts. Platform-side bug only, but the sum check reads like it covers this and does not.
- `FileRegionPublisher.fail()` checks `terminated` but not `cancelled`, so a read error after `cancel()` still calls `onError`. This is the cancel race already dispositioned in the body; noting it because the disposition described a different mechanism.

### Security
- **Verified clean, contrary to what I expected going in:** the executor's `UPLOAD_UNREACHABLE` cause chain does not carry the URL or host. Probe against an unresolvable signed URL: `java.net.ConnectException: null` → `java.net.ConnectException: null` → `java.nio.channels.UnresolvedAddressException: null`. So attaching the `IOException` as cause leaks nothing here, and the message assertions are sufficient.
- Redacted `toString` on `Upload` and `Part` is real and asserted, including on the grant returned through the real converters.
- The demo's filename handling holds: `Path.of(name).getFileName()` cannot produce a separator, and the one interesting input, `..`, resolves to an existing directory and fails the transfer rather than escaping. A name of exactly `/` returns null from `getFileName()` and would NPE into a 500; harmless, one guard.
- **No scheme check on platform-supplied upload URLs.** `requireUrl` accepts `http://`, which the executor's own tests rely on. Worth a decision rather than a change: this repo added https enforcement in `tsanetgit/Connect_SDK#72` and `#74` for operator-supplied SAS URLs because the credential rides in the URL, and the same reasoning applies to a grant's signed URL and its `headers` block. The difference is that these URLs come from the authenticated platform rather than from config, so the trust argument is defensible. State it in the executor javadoc either way.
- Demo-grade, disclosed in the body and fine to leave: the `uploads` map is unbounded, and `GET /api/uploads/{uploadId}` has no session guard, so the random UUID is the only protection.

### Regressions
- None outside the new surface. The two interface changes have their full consumer sets updated, and the V1 attachment path is untouched.

### Operational
- No new dependency; the JDK client and its test server are in the platform.
- **Single and relay report progress once, after the whole PUT returns.** For a 5 GB single upload the demo shows "Requesting a grant for X..." for the entire transfer and then jumps to the outcome, and the console prints one line at the end despite its javadoc promising "per-part progress as the upload runs". Multipart and resumable are fine. Given that this demo is the large-file story, a byte-level callback from the publisher would be worth the follow-up.

### Test gaps
Each of these is claimed in the body's policy table or javadoc and has no assertion: `Retry-After` honored and capped (`git grep -c "Retry-After"` over the test tree returns nothing); the relay `CONNECT_ONLY` retry actually retrying on a `ConnectException`, as opposed to the tested not-retrying-after-a-503; the 308-without-`Range` stall counter exhausting to `UPLOAD_UNREACHABLE`; a 5xx on a resumable chunk; and the connectivity branch in finding 1.

### Nits
- Body counts: `AttachmentUploadExecutorTest` is 11 tests, not 10, so the new offline total is 29 rather than 28.
- The client-side null checks in `createGrant` and `complete` hardcode status `200` into the exception even when the response was `201`.

shawn-tsanet:

Addressed in `6c59ac1`, pushed. After it: **connect-library 131 run, 0 failures, 1 skipped** (seven more tests), demo-ui 15 / 0, console 33 / 0.

|---|---|---|
| **Blocking:** connectivity failures echo Spring's URL-bearing message, and the V2 paths carry the case token | **Fixed.** The `ResourceAccessException` branch names the root cause's class only. Probe as QA's: a real `RestTemplate` against a closed port with token `CASETOKEN-MARKER`; the message contains neither the marker nor the host. | `aConnectivityFailureNeverEchoesTheCaseTokenThatRidesInTheUrl` |
| `verification.mode: none → DELIVERED_UNVERIFIED` is the platform's doing, not the client's | **Reworded in the body.** No client code branches on the mode, and none should. | policy table |
| `completeWithRetry` retries with no backoff | **Fixed.** Linear backoff per attempt, configurable, zero in the gateway tests. | `ConnectApiAttachmentsV2Gateway.pause` |
| duplicate `partNumber` values pass the sum check | **Fixed.** Rejected before any request. | `duplicatePartNumbersAreRejectedBeforeAnyRequest` |
| `FileRegionPublisher.fail()` ignores `cancelled` | **Fixed.** A read error after cancel is dropped. | `fail()` |
| no scheme check on platform-supplied upload URLs | **Decision, stated.** Verbatim execution is the contract's rule; the URLs are per-grant, short-lived and platform-issued, unlike operator-pasted SAS URLs. In the executor javadoc and the body. | javadoc |
| `getFileName()` is null for a bare separator | **Fixed.** Defaults to `attachment`. | `lastSegmentOrDefault` |
| `GET /api/uploads/{id}` has no session guard | **Fixed.** Same login requirement as the case endpoints. | `uploadState` |
| single and relay report progress once at the end | **Fixed.** The publisher reports bytes read per whole MiB while the body streams; a retry truthfully restarts from zero. | `singleModeReportsByteLevelProgressWhileTheBodyStreams` |
| test gaps: `Retry-After`, relay connect retry, 308-without-`Range` exhaustion, 5xx resumable chunk | **Fixed, four tests.** The 5xx case settles the auto-QA's reading: `TRANSIENT_STATUS_ONLY` keeps 5xx and 429 in the retry branch, so the chunk is re-PUT with the same `Content-Range`, as the policy table says. | `retryAfterIsHonoredAndCapped`, `relayRetriesWhenTheConnectionNeverOpens`, `aResumableSessionThatNeverReportsRangeFailsClosed`, `aResumableChunkAnsweredFiveHundredIsRePutWithTheSameContentRange` |
| precondition exceptions hardcode status 200 | **Fixed.** The response's real status. | `aGrantWithoutAnUploadBlockIsRejectedClientSideWithTheRealStatus` asserts 201 |
| body counts (11, 29) | **Fixed in the body**, now 17 and 35 after the additions. | body |
| unbounded `uploads` map | Left, demo-grade, already in the body's follow-ups. | body |

## Second gate run (the cap)

**CONCERNS**, two notes:

- *A store answering 308 with a stale but advancing `Range` could make one-byte progress forever.* Acknowledged as adversarial-only: the store is one the authenticated platform chose, and the per-request timeout still bounds each round trip. A total-attempt ceiling per resumable upload is a one-line follow-up if it is ever wanted; not added here to keep the change to the QA set.
- *Out-of-diff `TsaNetApiSession` implementers.* Settled by command earlier and again now: `git grep -ln "implements TsaNetApiSession"` returns the default and account-scoped sessions only, both in the diff.

## Follow-ups this leaves open

- The V1 demo `forward` endpoint's unsanitized file name (pre-existing, one line).
- Swap the hand-written records and the seam for generated classes when `#147` lands in the spec; wire a grant issuer to the env-gated executor test; the adversarial resumable ceiling above.

PR #78, opened 2026-09-07
connect-library + apps: connection hardening — classified API errors, password-mode renewal, /v1/me on login, https default (#20)

### LINKED ISSUE #20: Connection hardening: TLS default, re-auth on expiry/401, /v1/me, ProblemDetail mapping
**Re-scoped 2026-09-07** to what is still undone on `main` at `85e3b5c`, verified in the code (status table in the comment of the same date). The original scope had four items; the parts that shipped are recorded under "Already done" so this issue is only the remaining work.

## Context
From the 2026-06-01 evaluation note: the working connection should be production-honest before more is layered on it. The 2026-06-12 note still applies: not required for pre-GA, but required before publication, and v1.0.0 is published on GitHub Packages, so this is overdue rather than optional.

## Already done (not in this issue)
- **Token lifecycle for client credentials.** `TokenManager.ensureValidAccessToken` refreshes on expiry and `OAuth401RetryInterceptor` retries once on a 401, wired when the mode is `CLIENT_CREDENTIALS` (`tsanetgit/Connect_SDK#27`).
- **`/v1/me` is reachable.** `ConnectApiUserGateway` fetches `UserContextDTO` through the session's `users()` facade.

## Remaining scope, three items

1. **TLS default.** `TSANet-integration-app/src/main/resources/application.yml` ships `base-url: "http://localhost:8080"`. The shipped default must not invite sending a bearer over plaintext: default to an https URL (the demo-ui's environments are already `https://`), or ship no default and fail fast with a clear message when the base URL is absent or not https, with an explicit opt-out for a local mock.
2. **Token lifecycle for password mode.** `TokenManager.refreshAccessToken` throws for `CONNECT1_PASSWORD`, and the 401 interceptor is not wired for it, so an expired password-mode token surfaces as an opaque failure. The platform has no refresh endpoint, so re-login is the strategy: on expiry or a 401, re-login transparently and retry once; repeated failure surfaces a clear error. As part of validating the session against a real protected call, the login path should fetch `/v1/me` and keep the company and user context for routing, instead of leaving that to callers.
3. **ProblemDetail mapping, and the 500-with-body quirk.** The API answers 4xx with RFC 7807 `application/problem+json`, and `tsanetgit/Connect-API-Code#122` shows it also answers 500 with a meaningful JSON body where a 4xx belongs. Outside the V2 attachment client (`tsanetgit/Connect_SDK#77`, which maps problem details for its own three calls only), nothing in `connect-library` or the apps reads the body: a validation failure reaches the console and the demo as a generic server error. Map problem details once, in the client layer, into a structured exception the facades throw and the console and demo render (type, title, detail); classify the 500-with-body case by inspecting the body, not the status. The V2 client's `translate` is the template.

## Acceptance
- The shipped default configuration contains no plaintext HTTP base URL; a non-https base URL is rejected at startup unless explicitly opted out.
- In password mode, an expired token or a 401 triggers one transparent re-login and retry, proven by a test with a clock and a mocked 401; a second failure surfaces a clear error, not a stack trace. After login, `/v1/me` has been fetched and the context is available to subsequent commands.
- A 4xx problem-details response renders its type, title and detail in the console and the demo; a 500 with a problem-shaped body renders the same way rather than as an unknown error, proven against the recorded shapes from `tsanetgit/Connect-API-Code#122`.

## References
- Evaluation note 2026-06-01 (sections 5.2, 5.4, 5.5, 5.7) and 2026-06-05 addendum.
- `tsanetgit/Connect_SDK#3`, `#5`, `#27`, `#77`; `tsanetgit/Connect-API-Code#122`; `tsanetgit/Member-Integrations#33`.

shawn-tsanet:
Closes #20. Connection hardening, re-scoped on 2026-09-07 to the three items still undone: the shipped plaintext default, password-mode token lifecycle plus `/v1/me` after login, and problem-details mapping.

## What this changes, by item

**1. TLS default.** The console app and the scripted demo shipped `base-url: "http://localhost:8080"`, a plaintext target for a bearer token. Both now default to the https BETA host, the same default the demo-ui already uses. The console rejects a non-https base URL at startup with the fix in the message, unless `tsanet.api.allow-insecure-http` is true; the test profile opts in for its local mock. No guard in the library, since library tests and members' mocks use http and the acceptance names the shipped app configuration. Sweep receipt: `grep -rn 'http://' */src/main/resources` now finds only the console's own local webhook-receiver address, which is not a Connect target.

**2. Password lifecycle and `/v1/me`.** Login returns the token and its `expiresIn`; the store keeps the expiry for password mode too, so the 60-second skew and `isAuthorized` now apply to both modes. The token manager renews on expiry in both modes. For password mode it re-logins transparently with the configured credentials, only when the session belongs to that user; otherwise it fails with a clear "log in again", because an interactively typed password is never retained. The 401 retry interceptor is wired for both modes and asks the token manager whether renewal is possible. Login rides its own unauthenticated client, without the bearer supplier or the 401 retry: on the session client a re-login on an expired token would recurse into itself through both. An expiry that cannot be renewed surfaces from the bearer supplier as a classified 401 rather than a bare exception, so the console and the demo render it like every other failure. Both login paths end in one method that fetches `/v1/me` through the session client, keeps the company and user on the session (`AuthFacade.currentUserContext()`), and on failure clears the store and rethrows: a login is not a login until a protected call has answered.

**Verified live this session:** on BETA, `/v1/me` answers a client-credentials token with HTTP 200 and the company (user null), so fetching it after every login is safe in both modes. The advisor had flagged this as the one thing that could brick client-credentials mode; the probe settles it.

**3. Problem-details mapping.** A public `ConnectApiException` (kind, status, type, title, detail, instance; value-free message; `isProblem(typeSuffix)`), thrown by a response error handler on the library's one `RestTemplate` factory and classified by the body alone: RFC 7807 fields mean `PROBLEM`; a `message` field means `LEGACY`, the API's default shape that arrives with a 500 where a 4xx belongs (`tsanetgit/Connect-API-Code#122`, with `validationErrors` folded into the detail); anything else is `OTHER` with a 200-character excerpt, the request URL and path scrubbed out of it first because intermediaries echo the request URI. A connectivity interceptor, installed first so it also wraps the 401 retry, turns transport failures into the same exception naming only the cause class: once an `IOException` reaches `RestTemplate` its message carries the request URL, and the URL carries the case token. The generated client already sends `Accept: application/json, application/problem+json` on every call except health (checked in the generated sources this session), which is why the API answers with the documented 4xx; nothing had parsed the answer. The V2 attachment client converts the classified exception first, so its typed contract holds in production. The console prints the exception's message, which reads `HTTP 422 Unprocessable Entity (case-update-error): OPEN cases cannot be closed.`; the demo's error handler passes the upstream status through with the API's title, detail, type and the kind.

**Docs.** The user guide, runbook, deploy skill and its references described a bad login as "Connect API returned 500 — Error processing request". Through the SDK that is now `HTTP 401 Authentication Failed` with the API's detail; the text says so and keeps the legacy 500 as what a raw client without the header sees.

## Verification

- Reactor, `mvn test`: **connect-library 156 run, 0 failures, 1 skipped** (25 new offline tests; the one skip is the env-gated attachments live test), **attachment-receiver 182 / 0 / 40**, **demo-ui 18 / 0** (3 new), **console 37 / 0** (4 new), **scripted demo 6 / 0**.
- `ConnectApiResponseErrorHandlerTest` (8), through the generated client on the library's own `RestTemplate`, every request path carrying a marker token: the #122 shapes (422 problem on closure, legacy 500 with message, validation errors folded, 401 problem on login), an HTML body as excerpt, an intermediary page that echoes the request path scrubbed, a closed port as connectivity, and the body-first rule on a problem body riding a 500. No message carries the marker or the host.
- `TokenManagerTest` (+6): expiry from `expiresIn`, none when absent, transparent re-login for the configured user, a different typed user refused, interactive login tracked, refresh refused without a matching session.
- `OAuth401RetryInterceptorTest` (3, the first for this class): retry once with the renewed bearer, pass-through when renewal is impossible, non-401 untouched.
- `DefaultTsaNetApiSessionLoginTest` (5): interactive, configured and client-credentials logins all fetch `/v1/me` and keep the context; a failing `/v1/me` clears the store and rethrows; logout forgets the context. `TsaNetApiRuntimeBearerTest` (3): the bearer supplier's classified 401.
- `ConnectApiAttachmentsV2ApiTest` (+1): the classified exception keeps its type through the attachment client on the runtime template. `ConnectFacadeConfigurationTest` (4) and `ApiErrorHandlerTest` (3) for the guard and the demo rendering.
- Live: the `/v1/me` client-credentials probe above. The error mapping was not run against BETA; the shapes are the ones recorded on #122 on 2026-07-10, which the same API still returns per the memory verified 2026-07-09.

| claim | line |
|---|---|
| login cannot recurse through re-login | `TsaNetApiRuntime`: `loginApiClient` on `ConnectApiRestTemplates.create()`, no bearer supplier, no 401 interceptor |
| password re-login only for the configured user | `TokenManager.canReloginWithConfiguredPassword`; `itNeverRenewsASessionOpenedByADifferentTypedUser` |
| a login is not a login until `/v1/me` answered | `DefaultTsaNetApiSession.completeLogin`; `aFailingCurrentUserCallFailsTheLoginAndClearsTheStore` |
| classification is by body, not status | `ConnectApiResponseErrorHandler.classify`; `classificationIsByBodyFirstEvenWhenAProblemBodyRidesOnAFiveHundred` |
| the URL never reaches a message | `scrub`, `ConnectApiException.connectivity`; every handler test's `doesNotContain("MARKER")` |
| the 401 retry fires before classification | interceptor chain order (connectivity, then retry) runs inside `execute`; the error handler runs on the returned response |
| the generated client sends the problem+json Accept | `localVarAccepts` in every generated API except `HealthApi` (checked in `target/generated-sources`) |
| the console renders type, title, detail | `ConsoleShellRunner` prints `ex.getMessage()`, which `ConnectApiException.format` builds from those fields |

- **Positional and interface contracts.** `AuthFacade` gained `currentUserContext()`; implementers by command, `git grep -ln "implements AuthFacade"`: the default session (via `TsaNetApiSession`'s facade set) and the account-scoped session's inner facade, both updated. `ConnectApiAuthGateway.login` changed return type; callers by command: `TokenManager` only in main (the session now goes through `tokenManager.loginWithPassword`), plus the gateway test and the token-manager test, all updated. `ConnectApiSessionStore.savePassword` kept its two-argument form, so the four test call sites are untouched.
- **Guard vs representation.** `requireHttps` is a guard on operator configuration; the representation alternative, a URL type that cannot hold http, would reject every local mock. The opt-out is the escape hatch, in config, on purpose.
- **Newly reachable failures.** Every non-2xx and every transport failure now surfaces as `ConnectApiException`; receiving handlers are the console's `Command failed:` line, the login command's `Login failed:` line, and the demo's `handleConnectApi`. The demo's `handleUpstream(RestClientResponseException)` is now mostly unreachable from library calls and stays as a fallback.
- **Abstraction introduced.** The error handler centralizes classification; old form swept by command, `grep -rn 'catch (RestClientException\|catch (HttpStatusCodeException\|catch (RestClientResponseException'` over main sources: the OAuth token gateway (its own client, untouched) and the V2 attachment client, whose arms now sit behind the new catch.
- **Behavior change to state plainly.** Anyone running the console or the scripted demo against a local http mock must now set `tsanet.api.allow-insecure-http=true` or startup fails, by design. Password-mode sessions now expire on the API's `expiresIn`, so `isAuthorized` can flip to false where it never did before.
- **Schema / stored shape.** None. The SQLite user-context table was already written by the user gateway; it is now written on every login.

## Gate dispositions (CONCERNS, first run)

- *Concurrent 401s each trigger a re-login; no single-flight.* Acknowledged as a follow-up. Same class as the pre-existing client-credentials refresh; password mode widens it to a live login per concurrent 401. Correct, wasteful under load; a single-flight around `refreshAccessToken` is the fix when it matters.
- *`completeLogin` clears the store on a transient connectivity failure to `/v1/me`.* Designed. The acceptance says the context is available after login; a login whose first protected call did not answer is not a login yet, and the caller retries the login. Stated here rather than softened.
- *Confirm the multi-module build.* The reactor above is that receipt: all five modules green, including the console's context test on the opted-in http mock.

- Single-flight on `refreshAccessToken` across concurrent 401s, and atomic writes in `ConnectApiSessionStore` (a concurrent reader can see a new token beside a stale username or expiry).
- The demo-ui and the scripted demo build their own sessions and have no https guard; both ship https values. A guard in each is a one-liner if wanted.
- A failure while reading a 2xx body happens after the interceptor chain, so it can still surface Spring's URL-bearing `ResourceAccessException`. Rare; same residual class the scrub closes for non-2xx.
- Optional retry of `/v1/me` on an unattended client-credentials startup, where a blip on one endpoint now turns a working start into a hard failure.
- Once `tsanetgit/Connect-API-Code#125` fixes the spec's crossed error modes, the generated models will carry `ProblemDetail`; the handler can then deserialize instead of reading the tree.

## Advisor and gate

QA round at `4174d45`: two blockers (context dropped on renewal; lifetime at or below the skew born expired) fixed in `d0cc011`, disposition per note in the PR comments. Library suite at `d0cc011`: 163 run, 0 failures, 1 skipped.

Advisor consulted twice. Orientation: the login-recursion trap and its structural fix (a separate unauthenticated client), fail the login when `/v1/me` fails, do not retain typed passwords, keep the guard out of the library, BETA as the shipped default. Pre-done: blocked on whether `/v1/me` answers a client-credentials token (settled by the live probe, plus a client-credentials login test), the unrenewable-expiry exception reaching the demo unclassified (fixed at the bearer supplier), and the excerpt echoing the request path (fixed with the scrub). Pre-PR gate: **CONCERNS**, dispositions above. Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
@@ -15,10 +15,31 @@ import org.springframework.context.annotation.Configuration;
 public class ConnectFacadeConfiguration {
     @Bean
     TsaNetApiSessionFactory tsaNetApiSessionFactory(ConnectFacadeProperties properties) {
+        requireHttps(properties.api());
         return TsaNetApi.sessionFactory(TsaNetApiConnectionSettings.of(
             properties.api().baseUrl(),
             properties.storage() != null ? properties.storage().sqlitePath() : "data.db"
         ));
+    }
+
+    /**
+     * A bearer over plain http is a credential on the wire. Fail at startup, with the fix in
+     * the message, rather than at the first request.
+     */
+    static void requireHttps(ConnectFacadeProperties.Api api) {
+        String baseUrl = api == null ? null : api.baseUrl();
+        if (baseUrl == null || baseUrl.isBlank()) {
+            throw new IllegalStateException("tsanet.api.base-url is required");
+        }
+        if (baseUrl.regionMatches(true, 0, "https://", 0, 8)) {
+            return;
+        }
+        if (api.insecureHttpAllowed()) {
+            return;
+        }
+        throw new IllegalStateException("tsanet.api.base-url must use https; the Connect API is HTTPS-only in every real "
+            + "environment and a bearer token must not travel in plain text. Set tsanet.api.allow-insecure-http=true "
+            + "only for a local mock.");
     }
 
     @Bean
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeProperties.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeProperties.java
@@ -13,7 +13,14 @@ public record ConnectFacadeProperties(
     Storage storage,
     List<ApplicationUserAccountConfig> accounts
 ) {
-    public record Api(String baseUrl) {
+    /**
+     * @param baseUrl           the Connect API base URL; https in every real environment
+     * @param allowInsecureHttp opt-out of the https requirement for a local mock; null means false
+     */
+    public record Api(String baseUrl, Boolean allowInsecureHttp) {
+        public boolean insecureHttpAllowed() {
+            return Boolean.TRUE.equals(allowInsecureHttp);
+        }
     }
 
     public record Auth(String username, String password) {
--- a/TSANet-integration-app/src/main/resources/application.yml
+++ b/TSANet-integration-app/src/main/resources/application.yml
@@ -10,7 +10,11 @@ spring:
 
 tsanet:
   api:
-    base-url: "http://localhost:8080"
+    # The Connect API is HTTPS-only in every real environment; this default is the BETA
+    # host, the same default the demo-ui uses. A plain-http base URL is rejected at startup
+    # unless allow-insecure-http is true, which exists for a local mock and nothing else.
+    base-url: "https://connect2.tsanet.net"
+    allow-insecure-http: false
   storage:
     sqlite-path: "${user.home}/.tsanet-client-demo"
   # No accounts are configured here. This is a public repository, so credentials
--- a/TSANet-integration-app/src/main/resources/application.yml
+++ b/TSANet-integration-app/src/main/resources/application.yml
@@ -12,7 +12,8 @@ tsanet:
   api:
     # The Connect API is HTTPS-only in every real environment; this default is the BETA
     # host, the same default the demo-ui uses. A plain-http base URL is rejected at startup
-    # unless allow-insecure-http is true, which exists for a local mock and nothing else.
+    # unless allow-insecure-http is true. That switch exists for a local mock; it admits any
+    # non-https host, so it stays false everywhere else.
     base-url: "https://connect2.tsanet.net"
     allow-insecure-http: false
   storage:
--- a/TSANet-integration-app/src/test/java/com/tsanet/facade/config/ConnectFacadeConfigurationTest.java
+++ b/TSANet-integration-app/src/test/java/com/tsanet/facade/config/ConnectFacadeConfigurationTest.java
@@ -1,0 +1,40 @@
+package com.tsanet.facade.config;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.jupiter.api.Test;
+
+/** The startup guard: no bearer over plain http unless the operator said so, in the config, on purpose. */
+class ConnectFacadeConfigurationTest {
+
+    @Test
+    void anHttpsBaseUrlPasses() {
+        assertThatCode(() -> ConnectFacadeConfiguration.requireHttps(
+            new ConnectFacadeProperties.Api("https://connect2.tsanet.net", null))).doesNotThrowAnyException();
+        assertThatCode(() -> ConnectFacadeConfiguration.requireHttps(
+            new ConnectFacadeProperties.Api("HTTPS://connect2.tsanet.net", false))).doesNotThrowAnyException();
+    }
+
+    @Test
+    void aPlainHttpBaseUrlFailsAtStartupWithTheFixInTheMessage() {
+        assertThatThrownBy(() -> ConnectFacadeConfiguration.requireHttps(
+            new ConnectFacadeProperties.Api("http://connect2.tsanet.net", null)))
+            .isInstanceOf(IllegalStateException.class)
+            .hasMessageContaining("https")
+            .hasMessageContaining("tsanet.api.allow-insecure-http");
+    }
+
+    @Test
+    void theOptOutAdmitsALocalMockOnly() {
+        assertThatCode(() -> ConnectFacadeConfiguration.requireHttps(
+            new ConnectFacadeProperties.Api("http://localhost:8080", true))).doesNotThrowAnyException();
+    }
+
+    @Test
+    void aMissingBaseUrlIsItsOwnError() {
+        assertThatThrownBy(() -> ConnectFacadeConfiguration.requireHttps(new ConnectFacadeProperties.Api(" ", true)))
+            .isInstanceOf(IllegalStateException.class)
+            .hasMessageContaining("required");
+    }
+}
--- a/TSANet-integration-app/src/test/resources/application.yml
+++ b/TSANet-integration-app/src/test/resources/application.yml
@@ -5,6 +5,8 @@ spring:
 tsanet:
   api:
     base-url: "http://localhost:8080"
+    # Test-only: the context test points at a local mock that speaks plain http.
+    allow-insecure-http: true
   storage:
     sqlite-path: "${java.io.tmpdir}/tsanet-test"
   accounts:
--- a/TSANet-integration-demo/src/main/resources/application.yml
+++ b/TSANet-integration-demo/src/main/resources/application.yml
@@ -13,7 +13,8 @@ server:
 
 tsanet:
   api:
-    base-url: "http://localhost:8080"
+    # HTTPS-only in every real environment; BETA is the shipped default (see the console app).
+    base-url: "https://connect2.tsanet.net"
   storage:
     sqlite-path: "${user.home}/.tsanet-client-demo"
   # No accounts are configured here. This is a public repository, so credentials
--- a/connect-library/src/main/java/com/tsanet/api/ConnectApiException.java
+++ b/connect-library/src/main/java/com/tsanet/api/ConnectApiException.java
@@ -21,7 +21,9 @@ package com.tsanet.api;
  * </ul>
  *
  * <p>The message is value-free by construction: status, kind, the API's own title and detail,
- * never the URL, a header or a token.
+ * never the request URL, a header or a token. The request URL and its path are scrubbed from
+ * any echoed body before an excerpt is taken; a body that spells out an identifier on its own
+ * is not recognized as such.
  *
  * <p>Note for anyone raising the generated client's retry count: its retry loop catches
  * Spring's {@code HttpServerErrorException}, which this handler no longer throws, so that loop
@@ -61,7 +63,11 @@ public class ConnectApiException extends RuntimeException {
         return kind;
     }
 
-    /** HTTP status of the answer, or 0 for {@link Kind#CONNECTIVITY}. */
+    /**
+     * The status the API asserted: for {@link Kind#PROBLEM} the body's own {@code status} when it
+     * carries one (which can differ from the wire status, see tsanetgit/Connect-API-Code#122),
+     * otherwise the wire status; 0 for {@link Kind#CONNECTIVITY}. Not a transport-level fact.
+     */
     public int status() {
         return status;
     }
--- a/connect-library/src/main/java/com/tsanet/api/ConnectApiException.java
+++ b/connect-library/src/main/java/com/tsanet/api/ConnectApiException.java
@@ -1,0 +1,128 @@
+package com.tsanet.api;
+
+/**
+ * A Connect API call failed, in a form the caller can act on. Thrown from every call the
+ * library makes through its HTTP client, replacing Spring's transport exceptions (whose
+ * messages carry the request URL, and the URL carries the case token for most endpoints).
+ *
+ * <p>{@link #kind()} says what the API answered with:
+ * <ul>
+ *   <li>{@link Kind#PROBLEM}: an RFC 7807 body ({@code type}, {@code title}, {@code status},
+ *       {@code detail}, {@code instance}), which the API returns for the documented 4xx codes
+ *       because the library sends {@code Accept: application/json, application/problem+json}.</li>
+ *   <li>{@link Kind#LEGACY}: the API's default {@code {"message": ...}} body, usually with a
+ *       500 where a 4xx belongs (tsanetgit/Connect-API-Code#122; kept for backward
+ *       compatibility). Classified by the body, not the status, so it never reads as an
+ *       unknown server error.</li>
+ *   <li>{@link Kind#CONNECTIVITY}: the API could not be reached; {@link #detail()} names the
+ *       failure's class only.</li>
+ *   <li>{@link Kind#OTHER}: a non-2xx with no recognizable body; {@link #detail()} carries at
+ *       most a short excerpt of it.</li>
+ * </ul>
+ *
+ * <p>The message is value-free by construction: status, kind, the API's own title and detail,
+ * never the URL, a header or a token.
+ *
+ * <p>Note for anyone raising the generated client's retry count: its retry loop catches
+ * Spring's {@code HttpServerErrorException}, which this handler no longer throws, so that loop
+ * is inert. Retries belong in the caller that knows the operation is idempotent.
+ */
+public class ConnectApiException extends RuntimeException {
+
+    public enum Kind { PROBLEM, LEGACY, CONNECTIVITY, OTHER }
+
+    private final Kind kind;
+    private final int status;
+    private final String type;
+    private final String title;
+    private final String detail;
+    private final String instance;
+
+    public ConnectApiException(Kind kind, int status, String type, String title, String detail, String instance,
+                               Throwable cause) {
+        super(format(kind, status, type, title, detail), cause);
+        this.kind = kind;
+        this.status = status;
+        this.type = type;
+        this.title = title;
+        this.detail = detail;
+        this.instance = instance;
+    }
+
+    public static ConnectApiException connectivity(Throwable cause) {
+        Throwable root = cause;
+        while (root.getCause() != null && root.getCause() != root) {
+            root = root.getCause();
+        }
+        return new ConnectApiException(Kind.CONNECTIVITY, 0, null, null, root.getClass().getSimpleName(), null, cause);
+    }
+
+    public Kind kind() {
+        return kind;
+    }
+
+    /** HTTP status of the answer, or 0 for {@link Kind#CONNECTIVITY}. */
+    public int status() {
+        return status;
+    }
+
+    /** The RFC 7807 {@code type} URI, or null. */
+    public String type() {
+        return type;
+    }
+
+    public String title() {
+        return title;
+    }
+
+    /** The API's {@code detail}, the legacy {@code message}, the cause class, or a body excerpt. */
+    public String detail() {
+        return detail;
+    }
+
+    public String instance() {
+        return instance;
+    }
+
+    /** Whether the problem {@code type} ends with the given segment, for example {@code authentication-error}. */
+    public boolean isProblem(String typeSuffix) {
+        return type != null && (type.equals(typeSuffix) || type.endsWith("/" + typeSuffix));
+    }
+
+    private static String format(Kind kind, int status, String type, String title, String detail) {
+        StringBuilder message = new StringBuilder();
+        switch (kind) {
+            case CONNECTIVITY -> message.append("cannot reach the Connect API (").append(detail).append(')');
+            case PROBLEM -> {
+                message.append("HTTP ").append(status);
+                if (title != null && !title.isBlank()) {
+                    message.append(' ').append(title);
+                }
+                if (type != null && !type.isBlank()) {
+                    message.append(" (").append(lastSegment(type)).append(')');
+                }
+                if (detail != null && !detail.isBlank()) {
+                    message.append(": ").append(detail);
+                }
+            }
+            case LEGACY -> {
+                message.append("HTTP ").append(status).append(" (legacy error)");
+                if (detail != null && !detail.isBlank()) {
+                    message.append(": ").append(detail);
+                }
+            }
+            case OTHER -> {
+                message.append("HTTP ").append(status).append(" from the Connect API");
+                if (detail != null && !detail.isBlank()) {
+                    message.append(": ").append(detail);
+                }
+            }
+        }
+        return message.toString();
+    }
+
+    private static String lastSegment(String type) {
+        int slash = type.lastIndexOf('/');
+        return slash < 0 || slash == type.length() - 1 ? type : type.substring(slash + 1);
+    }
+}
--- a/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiAttachmentsV2Api.java
+++ b/connect-library/src/main/java/com/tsanet/api/connectapi/internal/ConnectApiAttachmentsV2Api.java
@@ -4,6 +4,7 @@ import com.tsanet.api.attachments.v2.AttachmentCompleteRequest;
 import com.tsanet.api.attachments.v2.AttachmentCompleteResult;
 import com.tsanet.api.attachments.v2.AttachmentGrant;
 import com.tsanet.api.attachments.v2.AttachmentGrantRequest;
+import com.tsanet.api.ConnectApiException;
 import com.tsanet.api.attachments.v2.AttachmentV2Exception;
 import com.tsanet.api.generated.invoker.ApiClient;
 import java.util.List;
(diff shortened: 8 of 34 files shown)


shawn-tsanet (commented):
**Verdict:** CONCERNS

**Minimal-Diff Rule check:**

- `ConnectFacadeConfiguration.requireHttps` + `Api.allowInsecureHttp` + `application.yml` (app & demo) → item 1, TLS default. Traces.
- `application.yml` (test) opt-in → item 1, keeps local mock working. Traces.
- `ConnectApiException` (new) → item 3, problem-details mapping. Traces.
- `ConnectApiResponseErrorHandler` / `ConnectApiConnectivityInterceptor` / `ConnectApiRestTemplates` (new) → item 3. Traces.
- `ConnectApiAttachmentsV2Api` catch arm → item 3, typed contract holds. Traces.
- `ConnectApiAuthGateway.login` return-type change + `PasswordLogin` → item 2, `expiresIn`. Traces.
- `ConnectApiSessionStore` savePassword overload + userContext → items 2. Traces.
- `TokenManager` refresh/re-login → item 2. Traces.
- `AuthFacade.currentUserContext` + `AccountScopedTsaNetApiSession` + `DefaultTsaNetApiSession.completeLogin` → item 2, `/v1/me`. Traces.
- `TsaNetApiRuntime` two-client split + bearer supplier classification → item 2, recursion guard. Traces.
- `ApiErrorHandler.handleConnectApi` (demo) → item 3, rendering. Traces.
- Docs (RUNBOOK, USER_GUIDE, deploy skill + refs) → "Docs" section. Traces.

No untraced hunks. Every non-trivial change maps to an explicit item in the PR body.

**Correctness:**

Traced a bad-login through `identity.login` on the session client: response handler `classify` reads the 401 problem body → `ConnectApiException(PROBLEM, 401, ...)` → `completeLogin` never runs because login threw first. Correct.

Traced `completeLogin` on a client-credentials auth: `authenticate()` → token stored → `userGateway.getCurrentUser()` (verified 200 on BETA per the probe) → context saved. On a transient `/v1/me` failure the store is cleared and rethrown — designed, disposition acknowledged. Fine.

One behavior worth flagging (not blocking): `ensureValidAccessToken` now calls `refreshAccessToken()` for **both** modes when `isExpired` is true. For an interactively-typed password session whose token expired, `canReloginWithConfiguredPassword()` returns false (username mismatch or no configured password), so `refreshAccessToken` throws `IllegalStateException`. That's caught in `bearerTokenForRequest` and reshaped into a classified 401 — correct. But the same `IllegalStateException` is *not* caught inside `TokenManager.ensureValidAccessToken` when called directly (e.g. any non-supplier caller). `OAuth401RetryInterceptorTest` covers the interceptor path and `TsaNetApiRuntimeBearerTest` covers the supplier path, so the reachable production paths are tested. No gap in shipped paths.

`scrub` replaces `url.getRawPath()` then `url.getPath()`. If a body echoes the decoded path but not the raw one, the second replace catches it; if it echoes the raw, the first does. Both null-guarded. The excerpt runs after scrub, and the `<request-path>` test confirms it. Correct.

The concurrent-401 no-single-flight and transient-`/v1/me`-clears-store items are both acknowledged as intentional/follow-up in the PR — agreed, neither holds the merge.

**Security:**

No secrets in the diff. The `requireHttps` guard is the right call — rejects a bearer over plaintext at startup with a fix in the message. `ConnectApiException` is value-free by construction and every handler test asserts `doesNotContain("MARKER")` / no host — the URL-carries-token leak is closed at both the error-handler and connectivity paths. Interactively-typed passwords are never retained; `canReloginWithConfiguredPassword` gates re-login to the configured user only, with `itNeverRenewsASessionOpenedByADifferentTypedUser` proving it. No injection surface. Clean.

**Regression risk:**

- `ConnectApiAuthGateway.login` return type changed String→`PasswordLogin`. Callers enumerated in PR body (`TokenManager`, gateway test, token-manager test) — all updated in the diff; I confirmed `TokenManager.authenticatePassword` and both tests consume `login.accessToken()`. No stray caller in the diff.
- `AuthFacade` gained `currentUserContext()`; both implementers (`DefaultTsaNetApiSession`, `AccountScopedTsaNetApiSession` inner facade) updated. If any other `implements AuthFacade` exists outside this diff it would fail to compile — the reactor's green build (156/0/1 library, plus the four consuming modules) is the receipt that no such implementer was missed.
- `savePassword(username, token)` two-arg form retained, so the four existing test call sites are untouched — confirmed the overload delegates with `null` expiry.
- The 401 retry interceptor now installs for **both** modes (previously client-credentials only). This widens the retry surface to password sessions. `handleUpstream(RestClientResponseException)` in the demo becomes mostly unreachable but stays as fallback — acknowledged. This is the change with the widest blast radius; the reactor green across demo-ui/console/scripted-demo is the coverage that it didn't break existing renderers.

**Operational:**

- New config key `tsanet.api.allow-insecure-http` — this is a Spring `@ConfigurationProperties` record field, not an env var, and the shipped `application.yml` (app & test) both set it. No `.env.example` in this repo's scope for these modules. Fine.
- **Behavior change requiring operator awareness:** anyone running the console or scripted demo against a local http mock must now set `allow-insecure-http=true` or startup fails. Stated plainly in the PR. No sidecar/LaunchAgent/schema touched.
- No new dependency, no schema/migration change (user-context table pre-existed).

**Recommendation:** Merge; the two acknowledged items (single-flight on concurrent 401s, transient-`/v1/me` clearing the store) are correct-by-design follow-ups and none of them holds this diff.

---

shawn-tsanet (commented):
## QA review of connection hardening: classified API errors, password-mode renewal, /v1/me on login, https default (#20)

Separate-session QA pass at head `4174d45`.

**Verdict:** ❌ Needs changes

Two items, both small fixes in the token/session layer. Everything else below is a note, and the three claims this PR rests on that I could check independently all held.

Receipts: `mvn test` in the local clone at `4174d45`, whose merge-base with `origin/main` is main's head `85e3b5c`: **connect-library 156 run, 0 failures, 1 skipped**, attachment-receiver 182 / 0 / 40, demo-ui 18 / 0, console 37 / 0, scripted demo 6 / 0, all matching the body. Same JDK 26 caveat as the previous reviews.

### Blocking
1. **The user context is silently dropped on every token renewal, and nothing puts it back.** Traced by command rather than by eye:
   - `saveUserContext` has exactly one caller in main: `DefaultTsaNetApiSession:224`, inside `completeLogin`.
   - `completeLogin` has exactly two callers: `DefaultTsaNetApiSession:201` (`authenticate`) and `:211` (`login`).
   - `savePassword` / `saveOAuth` are called from `TokenManager:120` and `:113`, both reachable from `refreshAccessToken()`.
   - Both store methods null the field (`ConnectApiSessionStore:26` and `:44`).

   So a renewal — the 401 retry interceptor firing on any request, or the expiry branch in `ensureValidAccessToken` — empties `currentUserContext()` for a session that remains perfectly authenticated, and no path repopulates it. Issue 20's acceptance is "After login, `/v1/me` has been fetched and the context is available to subsequent commands"; a renewal happens *during* those subsequent commands. `DefaultTsaNetApiSessionLoginTest` asserts the context after login, after a failed login and after logout, never after a renewal, which is why the suite is green. The only reason this is not user-visible today is that nothing outside the library and its tests reads `currentUserContext()` yet, and it is public API on a published artifact. Clearing on a *different* identity is right, so the fix is to clear on identity change rather than on every save, or to refetch after a renewal.

2. **A token whose `expiresIn` is at or below the 60-second skew is born expired.** `TokenManager:14` sets `EXPIRY_SKEW = Duration.ofSeconds(60)`, and `ConnectApiSessionStore:75` returns expired when `now >= expiresAt - skew`. With `expiresIn <= 60` that is true the instant login returns. By mode: a configured-password or client-credentials session then renews on *every* request, one login per request, each new token equally expired; an interactively typed session gets `refreshAccessToken` throwing, so the first call after a **successful** login fails with "log in again". This is newly reachable for password mode, since before this PR password sessions stored no expiry at all. The PR's own test data sits exactly on the boundary (`new PasswordLogin("typed-token", 60)` in `TokenManagerTest`), and it passes only because it asserts the stored instant and never asks `isExpired`. Ignoring or flooring an `expiresIn` at or below the skew is a one-line guard; a test at 60 and just under it would pin it.

### Minimal-Diff Rule
- Pass. Thirty-three files, each traceable to one of the three items or to the docs sweep the error-shape change forces. The positional contracts check out by command: `implements AuthFacade` and `implements TsaNetApiSession` return only the default and account-scoped sessions, both updated, and the changed `login` return type has exactly the callers the body names.

### Correctness
- **Recursion is genuinely prevented.** The login `ApiClient` is built on its own `ConnectApiRestTemplates.create()` with no bearer supplier and no 401 interceptor, so a re-login triggered by an expired token cannot re-enter itself. This was the right structural fix rather than a flag.
- **Interceptor order matches the claim**: connectivity is added in the factory, the 401 retry is appended afterwards in the runtime, so connectivity wraps the retry.
- **Classification precedence** (problem, then legacy, then excerpt) is exercised through the real `RestTemplate` and the generated client, not a hand-rolled stub, and the scrub runs before the excerpt. Good tests.
- `ConnectApiException.status()` for a `PROBLEM` is the **body's** status when the body carries one, not the wire status. That is deliberate and tested, and it is what makes the V2 gateway stop retrying a 500-shaped 422. The getter's javadoc says "HTTP status of the answer", which reads as the wire status; worth a word so nobody uses it for transport decisions.
- `scrub` removes the full URL and its path, not a bare token appearing on its own in a body. An error page that names only the identifier still reaches the `OTHER` excerpt. Narrow, but the class javadoc's "never ... a token" is absolute.
- A failure while reading a 2xx body happens after the interceptor chain has returned, so it can still surface as Spring's URL-bearing `ResourceAccessException`. Same class of residual as above, rarer.

### Security
- **The premise of the whole error-mapping item is true, and I verified it independently** rather than taking the body's word: every generated API except `HealthApi` sends `"application/json", "application/problem+json"`, `IdentityApi.login` included. So the doc, runbook and skill rewrites that now promise a 401 for a bad login rest on something real.
- Typed passwords are never retained: `loginWithPassword` builds a throwaway config, the manager's own credentials are untouched, and `canReloginWithConfiguredPassword` requires the session's username to equal the configured one before any silent renewal. Both directions are tested.
- The https guard reads correctly and its message names the fix. One precision point: `allow-insecure-http=true` admits any non-https host, not only a local mock, despite the comment saying "for a local mock and nothing else". Fine as an escape hatch; the comment overstates what is enforced.
- The shipped default host checks out. `connect2.tsanet.net` resolves and answers `200` on `/v1/health`, it is the same BETA entry `demo-ui` already ships, and production `.org` is deliberately unlisted there. The claim "the same default the demo-ui already uses" is accurate.
- My blocking finding on `tsanetgit/Connect_SDK#77` was fixed before that PR merged, so I am not re-raising it: main's connectivity branch names only the root cause class, and this PR adds the classified branch in front of it.

### Regressions
- **The scripted demo got the https default but not the guard.** `TSANet-integration-demo` depends on `connect-library` only, not on the console module, and builds its session through its own `TsaNetApiConfigurationBean`, so `requireHttps` never runs there and the module has no `allow-insecure-http` key. Its shipped value is https, so the first half of issue 20's acceptance holds; an operator overriding it to http is accepted silently. The body scopes the guard to the console, so this is a gap against the acceptance rather than against the body. `demo-ui` is already listed as a follow-up.
- Widening the 401 retry to password mode is the change with the largest blast radius, and it is now the only class here with its own test file. Reasonable.
- Concurrent 401s each drive their own renewal; disclosed as a follow-up and I agree it is not for this PR. Adjacent and worth folding into the same follow-up: `savePassword` and `saveOAuth` write four or five volatile fields non-atomically, so a concurrent reader can observe a new token beside a stale username or expiry.

### Operational
- Behavior changes are stated plainly in the body: a local http mock now needs the opt-out, and password sessions can now report themselves unauthorized where they never did. Both are correct directions.
- A transient `/v1/me` failure now fails the whole login and clears the store. Dispositioned as designed, and I agree for an interactive login. Worth watching for unattended client-credentials integrations, where a blip on one endpoint now turns a working startup into a hard failure.

### Nits
- None beyond the javadoc wording noted above.

shawn-tsanet:
## Disposition of the QA reviews at `4174d45`

Fixed in `d0cc011`. Receipts: `mvn -pl connect-library test` at `d0cc011`: **163 run, 0 failures, 1 skipped** (was 156; seven tests added). Five of the seven fail against `4174d45` with the fix stashed, so they pin the defects rather than the implementation.

### Separate-session QA, blocking

**1. User context dropped on every renewal.** Fixed at the store. `savePassword` and `saveOAuth` now call `forgetUserContextUnlessSamePrincipal(mode, principal)` before assigning, so the context survives a renewed bearer for the same principal (configured-user re-login, fresh client-credentials token) and is dropped when the principal or the mode changes. `grep -rn 'userContext = null' connect-library/src/main` now returns exactly two sites: the helper and `clear()`. The 2-arg `savePassword` overload delegates to the 3-arg one, so it takes the same path. Pinned by `ConnectApiSessionStoreTest.aRenewedBearerForTheSamePrincipalKeepsTheUserContext` (both modes), `...aBearerForADifferentPrincipalDropsTheUserContext` (other user, and same name across a mode switch), `TokenManagerTest.aRenewalKeepsTheUserContextOfTheSamePrincipal` (through `ensureValidAccessToken`), and at the acceptance level `DefaultTsaNetApiSessionLoginTest.aTokenRenewalDuringTheSessionKeepsTheContext`. Chose clear-on-identity-change over refetch-after-renewal: the context describes the principal, not the token, and a refetch would couple `TokenManager` to the user gateway.

**2. Lifetime at or below the skew is born expired.** Fixed in `TokenManager.expiryFrom(Long)`: a stated lifetime at or below `EXPIRY_SKEW` is treated as unstated (no expiry tracked; the 401 path covers it), for both password and client-credentials tokens. Ignoring rather than flooring: a floor fabricates a lifetime the server did not state and, for any small floor, reproduces the login-per-request loop. Pinned at the boundary: `aLifetimeAtOrBelowTheSkewIsNotTrackedSoTheTokenIsNotBornExpired` (60 s typed login, first `ensureValidAccessToken` returns the token, one login total), `aLifetimeJustAboveTheSkewIsTrackedAndValidOnArrival` (61 s), `aShortLivedClientCredentialsTokenIsNotBornExpiredEither`. The existing `anInteractiveLoginTracksExpiryToo` moved from 60 s to 900 s; 60 is the value the guard now intentionally declines to track.

### Separate-session QA, notes

- **`status()` javadoc**: rewritten to say a `PROBLEM` carries the body's own status when present, otherwise the wire status, and that it is not a transport-level fact. Handler receipt: `classify` picks `bodyStatus` over the wire status when the body has an integer `status`.
- **"never a token" is absolute**: class javadoc now states the scrub removes the request URL and its path from any echoed body, and that an identifier spelled out on its own is not recognized. Behavior unchanged.
- **2xx body-read failure can still surface Spring's URL-bearing exception**: acknowledged, not addressed here. Same residual class as above; folded into the follow-up below.
- **`allow-insecure-http` comment overstates**: yml comment now says the switch admits any non-https host and stays false everywhere but a local mock.
- **Scripted demo has the https default but not the guard**: correct, the body scoped the guard to the console. Follow-up together with demo-ui (both build their own session, neither runs `requireHttps`).
- **Non-atomic store writes** (a concurrent reader can see a new token beside a stale username or expiry): follow-up, folded into the single-flight item already listed.
- **Transient `/v1/me` failing an unattended client-credentials startup**: designed, stands; noted for the follow-up so a retry-on-startup can be considered there.
- **Independent verification of the `Accept` header claim and the default host**: thank you, nothing to change.

### Auto-QA (CONCERNS, recommends merge)

- `ensureValidAccessToken` throwing `IllegalStateException` for a non-supplier caller: no such caller ships; the supplier and interceptor paths are the tested ones. Acknowledged.
- Widened 401 retry to password mode and the demo's `handleUpstream` fallback: acknowledged as stated.

### Follow-ups (one issue, to be filed after merge)

single-flight on concurrent renewals; atomic store writes; https guard for demo-ui and the scripted demo; residual URL-bearing exception on a 2xx body-read failure; optional retry of `/v1/me` on unattended startup.

shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
Pass. Every hunk traces to a prior blocking finding: `forgetUserContextUnlessSamePrincipal` (finding 1), `expiryFrom` (finding 2), and the javadoc/yml-comment tweaks to the four accepted notes.

**Correctness:**
Traced finding 1's fix: password renewal for the same user — `savePassword("api@test.com", "t2", ...)` sees `authMode==CONNECT1_PASSWORD` and `username=="api@test.com"`, so `forgetUserContextUnlessSamePrincipal` leaves `userContext` intact; a different user or a mode switch clears it. Confirmed the 2-arg `savePassword` delegates to the 3-arg form (`savePassword(username, token)` → 3-arg with null expiry), so it takes the same guard path. Correct, and `aBearerForADifferentPrincipalDropsTheUserContext` pins both the other-user and mode-switch branches.

Traced finding 2's fix: `expiryFrom(60)` with `EXPIRY_SKEW=60` → `60 <= 60` → returns null → no expiry tracked → `isExpired(NOW)` false → first `ensureValidAccessToken` returns the token, one login total. `expiryFrom(61)` → tracked, valid on arrival. Boundary pinned at 60 and 61 for password, and at the skew for client-credentials. The clear-on-identity-change over refetch choice keeps `TokenManager` decoupled from the user gateway — sound.

One note, not blocking: `expiryFrom` widens the `Long`→boxed handling but the client-credentials path passes a primitive `long` from `token.expiresInSeconds()`, autoboxed — non-null always, so the null branch there is dead but harmless. No behavior change.

**Security:**
None. Javadoc/comment edits are documentation-only; no secrets, no new input surface. The `allow-insecure-http` comment now accurately states it admits any non-https host.

**Regression risk:**
`forgetUserContextUnlessSamePrincipal` is the only new branch in `savePassword`/`saveOAuth`; both are reached from `TokenManager` renewal and from the 4 existing 2-arg test call sites (unaffected — same-principal or fresh store, context null anyway). No consumer of `saveOAuth`/`savePassword` outside the diff changes behavior for the first-login case (fresh store → `authMode` null → context cleared, same as before).

**Carried forward:**
- User context dropped on renewal — RESOLVED
- Lifetime ≤ skew born expired — RESOLVED
- status() javadoc misleading — RESOLVED
- "never a token" absolute vs scrub — RESOLVED (doc clarified, behavior unchanged)
- allow-insecure-http comment overstates — RESOLVED
- Scripted-demo/demo-ui https guard gap — FOLLOW-UP
- Non-atomic store writes / single-flight — FOLLOW-UP
- 2xx body-read URL-bearing exception — FOLLOW-UP
- Transient /v1/me on unattended startup — FOLLOW-UP

**Recommendation:** Merge; both prior blockers are fixed with boundary-pinning tests and the remaining items are the disclosed post-merge follow-ups.

---

PR #80, opened 2026-09-07
Connection hardening follow-ups: single-flight renewal, atomic session store, value-free 2xx read failure, /v1/me retry, https guard in both demos (#79)

### LINKED ISSUE #79: Connection hardening follow-ups from #78: single-flight renewal, atomic session store, https guard in both demos
Follow-ups deferred from `tsanetgit/Connect_SDK#78` (merged e39392e, closes #20). Each was raised in review and dispositioned as out of that PR's scope; none is a defect in shipped behavior today, and each is small.

## Items

- [ ] **Single-flight on renewal.** Concurrent 401s (or concurrent expiry checks) each drive their own `refreshAccessToken()`, so N in-flight requests can produce N logins. Serialize the renewal so callers waiting on it share one result. Applies to both client-credentials and password mode.
- [ ] **Atomic writes in `ConnectApiSessionStore`.** `savePassword` and `saveOAuth` assign five volatile fields one at a time; a concurrent reader can observe a new token beside a stale username or expiry. Hold the session in one immutable snapshot object behind a single volatile reference.
- [ ] **https guard in demo-ui and the scripted demo.** Both build their own sessions and never run the console's `requireHttps`; their shipped values are https, but an operator override to http is accepted silently. Same one-liner and `allow-insecure-http` opt-out as the console.
- [ ] **URL-bearing exception on a 2xx body-read failure.** A read failure after the interceptor chain has returned can still surface Spring's `ResourceAccessException`, whose message carries the request URL. Same residual class the error handler's scrub closes for non-2xx; wrap it value-free at the same boundary.
- [ ] **Optional retry of `/v1/me` on unattended startup.** A transient failure on the post-login `/v1/me` call fails the login and clears the store (by design for interactive use). For an unattended client-credentials integration a bounded retry before failing startup would avoid turning a blip into a hard failure.

## Acceptance

- One renewal per burst of concurrent 401s, proven by a test with N threads and one gateway call.
- No test can observe a token from one save beside a username or expiry from another.
- Both demos reject an http base URL at startup unless opted in, with the same message the console prints.
- A simulated 2xx body-read failure surfaces as `ConnectApiException(CONNECTIVITY)` naming only the failure class.

shawn-tsanet:
Closes #79. The five follow-ups deferred from `tsanetgit/Connect_SDK#78`, one commit each, in dependency order, plus one pre-PR review commit.

## What changed

1. **Atomic session store.** `ConnectApiSessionStore` holds one immutable `Snapshot` behind a single volatile reference; writes are synchronized and replace it whole, every getter reads one, `clear()` installs the empty one. A reader can no longer see a bearer beside another login's username or expiry, and `isAuthorized` judges a token against its own expiry. Public surface unchanged.
2. **Single-flight renewal.** `TokenManager` serializes renewals behind one lock. `ensureValidAccessToken` re-reads the snapshot under the lock and returns the token a previous caller stored. The 401 interceptor passes the bearer the rejected request carried (prefix stripped) to `renewUnlessAlreadyRenewed`: a different unexpired bearer already in the store answers the 401 with no network call; the same bearer is bad server-side whatever its expiry says, and one renewal runs.
3. **Value-free 2xx body-read failure.** The connectivity interceptor reads the response body inside its try. The library's buffering factory caches it there; the extractor and the error handler read the cached bytes; a mid-read failure is classified `CONNECTIVITY` naming only the failure's class, instead of Spring's `ResourceAccessException` with the request URL and the case token in it.
4. **`/v1/me` retry on unattended login.** `authenticate()` (configured credentials) retries the post-login `/v1/me` call on `CONNECTIVITY` failures only: three attempts, 250 ms then 500 ms, interrupt restored and failed at once. `login()` with a typed password does not retry, nothing the API answered is retried, the final failure still clears the store. No new config key.
5. **One https decision, three apps.** `ConnectApiBaseUrl.requireHttps(baseUrl, allowInsecureHttp, optOutSetting)` in the library: pure, no Spring, message names the caller's own setting. The console delegates to it (behavior and tests unchanged). The scripted demo gains `tsanet.api.allow-insecure-http` (false shipped, true in its test config) checked when the session factory bean is built. The demo-ui gains `tsanet.demo.allow-insecure-http` (false shipped) and checks every configured environment's `api-base-url` once in `EnvironmentService`'s constructor. The library's own request paths never call the helper, so tests and local mocks keep using http.
6. **Review round.** `refreshAccessToken` had no production caller left and is removed; the retry count derives from the backoff table's length; `clear()` documents the pre-existing logout-during-renewal window it does not close.

## Behavior changes an operator sees

- The scripted demo and the demo-ui now refuse a plain-http base URL at startup, with the fix in the message. Both ship https values; a local mock needs the opt-in.
- A service using configured credentials waits up to 750 ms longer before a login fails when `/v1/me` is unreachable.

## Receipts

Whole reactor at `016f6f8` (and connect-library again after the review commit):

| module | run | failures | skipped |
|---|---|---|---|
| connect-library | 181 | 0 | 1 |
| attachment-receiver | 182 | 0 | 40 |
| demo-ui | 19 | 0 | 0 |
| TSANet-integration-app | 37 | 0 | 0 |
| TSANet-integration-demo | 8 | 0 | 0 |

Red-first probes: the eight-caller renewal test fails with the lock removed (more than one login); the body-read test fails before the interceptor change with `ResourceAccessException: I/O error on POST request for "https://connect.example/v1/collaboration-requests/CASETOKEN-MARKER/closure"`, the exact residual the #78 QA named.

- **Abstraction introduced (`ConnectApiBaseUrl`), old form grepped:** `grep -rn 'regionMatches(true, 0, "https://"\|startsWith("https\|startsWith("http\|"https://"'` over every module's main code returns nothing outside the helper. The console's inline check was the only prior site and it now delegates.
- **Guard added, why not a representation change:** the https rule is policy over a configured string; there is no representation that makes http unrepresentable without also breaking the local-mock case, hence a guard with an explicit opt-in.
- **Newly reachable failures, receiving handler:** the scripted demo and demo-ui startup failures are `IllegalStateException` out of bean construction, the same shape the console already produced, surfaced by Spring Boot's startup failure report. The `CONNECTIVITY` classification of a body-read failure lands in the same handlers every other connectivity failure does (console renderer, demo `ApiErrorHandler.handleConnectApi` → 502).
- **Positional contracts:** `TsaNetApplicationProperties.Api` and `DemoProperties` each gained a component. `grep -rn 'TsaNetApplicationProperties.Api(\|new DemoProperties('` lists five sites, all tests, all updated in this diff; Spring binds both by name.
- **Prose claims, line that makes them true:** "the library's buffering factory caches it there" rests on `ConnectApiRestTemplates.create()` wrapping `SimpleClientHttpRequestFactory` in `BufferingClientHttpRequestFactory`, and `grep -rn 'ConnectApiConnectivityInterceptor\|new RestTemplate('` over main code shows that factory is the only place the interceptor is installed (the Entra token gateway's bare `RestTemplate` carries no interceptors). "No production caller" for `refreshAccessToken` rests on the same grep returning only two test `never()` verifications, both retargeted.
- **Schema/contract:** none; no stored shape changed. The `Snapshot` is in-memory.

## Not in this PR

- A renewal already past its network call when logout runs stores its token afterwards, and a straggling 401 on a client-credentials session logs an empty store in again. Both predate this PR; `clear()` and `renewUnlessAlreadyRenewed` state them, and the empty-store branch is pinned by test.
- `savePassword` carries the previous snapshot's `accountId` forward (pre-existing, pinned by test).
- A `DemoProperties` opt-in applies to all environments at once; per-environment opt-in was not asked for.

## Advisor and gate

QA round at `b32ce90` (separate session: merge with notes; auto-QA: looks good) answered in the PR comments; fixes in the seventh commit: `clear()` synchronized, two javadocs corrected, connectivity interceptor package-private, base-url setting name in the missing-URL message, empty-store 401 branch pinned. Reactor green, library 182.

Advisor consulted twice. Orientation: land the atomic store before single-flight (the observed-token comparison needs one coherent snapshot), strip the `Bearer ` prefix, make the body-read test also prove identical bytes reach the extractor and an empty 204 does not throw, keep the https helper pure and never called from the library's own paths, retry `/v1/me` only unattended and only on connectivity. Pre-done: sound on the two-monitor interplay and `saveUserContext` racing a renewal; asked for the four greps above, all clean, and the removal of the caller-less method. Pre-PR gate: **CONCERNS** on both runs (cap). Run 1: the buffering invariant and the positional enumerations, answered above; the backoff/attempt coupling, fixed in the review commit. Run 2 at `b32ce90`: the same buffering receipt (the `create()` line cited above), a console-side caller of the removed method (the grep above spans every module and finds none), the 750 ms worst-case wait and the startup failure on a plain-http config (both under behavior changes). Nothing new to change. Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
@@ -1,5 +1,6 @@
 package com.tsanet.facade.config;
 
+import com.tsanet.api.ConnectApiBaseUrl;
 import com.tsanet.api.TsaNetApi;
 import com.tsanet.api.TsaNetApiConnectionSettings;
 import com.tsanet.api.TsaNetApiSession;
@@ -24,22 +25,15 @@ public class ConnectFacadeConfiguration {
 
     /**
      * A bearer over plain http is a credential on the wire. Fail at startup, with the fix in
-     * the message, rather than at the first request.
+     * the message, rather than at the first request. The decision itself lives in the library
+     * ({@link ConnectApiBaseUrl}) so the console and both demos make it the same way.
      */
     static void requireHttps(ConnectFacadeProperties.Api api) {
-        String baseUrl = api == null ? null : api.baseUrl();
-        if (baseUrl == null || baseUrl.isBlank()) {
-            throw new IllegalStateException("tsanet.api.base-url is required");
-        }
-        if (baseUrl.regionMatches(true, 0, "https://", 0, 8)) {
-            return;
-        }
-        if (api.insecureHttpAllowed()) {
-            return;
-        }
-        throw new IllegalStateException("tsanet.api.base-url must use https; the Connect API is HTTPS-only in every real "
-            + "environment and a bearer token must not travel in plain text. Set tsanet.api.allow-insecure-http=true "
-            + "only for a local mock.");
+        ConnectApiBaseUrl.requireHttps(
+            api == null ? null : api.baseUrl(),
+            api != null && api.insecureHttpAllowed(),
+            "tsanet.api.allow-insecure-http"
+        );
     }
 
     @Bean
--- a/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
+++ b/TSANet-integration-app/src/main/java/com/tsanet/facade/config/ConnectFacadeConfiguration.java
@@ -32,6 +32,7 @@ public class ConnectFacadeConfiguration {
         ConnectApiBaseUrl.requireHttps(
             api == null ? null : api.baseUrl(),
             api != null && api.insecureHttpAllowed(),
+            "tsanet.api.base-url",
             "tsanet.api.allow-insecure-http"
         );
     }
--- a/TSANet-integration-app/src/test/java/com/tsanet/facade/config/ConnectFacadeConfigurationTest.java
+++ b/TSANet-integration-app/src/test/java/com/tsanet/facade/config/ConnectFacadeConfigurationTest.java
@@ -35,6 +35,6 @@ class ConnectFacadeConfigurationTest {
     void aMissingBaseUrlIsItsOwnError() {
         assertThatThrownBy(() -> ConnectFacadeConfiguration.requireHttps(new ConnectFacadeProperties.Api(" ", true)))
             .isInstanceOf(IllegalStateException.class)
-            .hasMessageContaining("required");
+            .hasMessageContaining("tsanet.api.base-url is required");
     }
 }
--- a/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApiConfigurationBean.java
+++ b/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApiConfigurationBean.java
@@ -1,6 +1,7 @@
 package com.tsanet.application.config;
 
 import com.tsanet.api.ApplicationUserAccountRegistry;
+import com.tsanet.api.ConnectApiBaseUrl;
 import com.tsanet.api.TsaNetApi;
 import com.tsanet.api.TsaNetApiConnectionSettings;
 import com.tsanet.api.TsaNetApiSession;
@@ -15,6 +16,12 @@ import org.springframework.context.annotation.Configuration;
 public class TsaNetApiConfigurationBean {
     @Bean
     TsaNetApiSessionFactory tsaNetApiSessionFactory(TsaNetApplicationProperties properties) {
+        // A bearer over plain http is a credential on the wire; refuse at startup, same rule as the console.
+        ConnectApiBaseUrl.requireHttps(
+            properties.api() == null ? null : properties.api().baseUrl(),
+            properties.api() != null && properties.api().insecureHttpAllowed(),
+            "tsanet.api.allow-insecure-http"
+        );
         return TsaNetApi.sessionFactory(TsaNetApiConnectionSettings.of(
             properties.api().baseUrl(),
             properties.storage() != null ? properties.storage().sqlitePath() : "data.db"
--- a/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApiConfigurationBean.java
+++ b/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApiConfigurationBean.java
@@ -20,6 +20,7 @@ public class TsaNetApiConfigurationBean {
         ConnectApiBaseUrl.requireHttps(
             properties.api() == null ? null : properties.api().baseUrl(),
             properties.api() != null && properties.api().insecureHttpAllowed(),
+            "tsanet.api.base-url",
             "tsanet.api.allow-insecure-http"
         );
         return TsaNetApi.sessionFactory(TsaNetApiConnectionSettings.of(
--- a/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApplicationProperties.java
+++ b/TSANet-integration-demo/src/main/java/com/tsanet/application/config/TsaNetApplicationProperties.java
@@ -13,7 +13,14 @@ public record TsaNetApplicationProperties(
     Storage storage,
     List<ApplicationUserAccountConfig> accounts
 ) {
-    public record Api(String baseUrl) {
+    /**
+     * @param baseUrl           the Connect API base URL; https in every real environment
+     * @param allowInsecureHttp opt-out of the https requirement for a local mock; null means false
+     */
+    public record Api(String baseUrl, Boolean allowInsecureHttp) {
+        public boolean insecureHttpAllowed() {
+            return Boolean.TRUE.equals(allowInsecureHttp);
+        }
     }
 
     public record Auth(String username, String password) {
--- a/TSANet-integration-demo/src/main/resources/application.yml
+++ b/TSANet-integration-demo/src/main/resources/application.yml
@@ -14,7 +14,10 @@ server:
 tsanet:
   api:
     # HTTPS-only in every real environment; BETA is the shipped default (see the console app).
+    # A plain-http base URL is rejected at startup unless allow-insecure-http is true. That
+    # switch exists for a local mock; it admits any non-https host, so it stays false elsewhere.
     base-url: "https://connect2.tsanet.net"
+    allow-insecure-http: false
   storage:
     sqlite-path: "${user.home}/.tsanet-client-demo"
   # No accounts are configured here. This is a public repository, so credentials
--- a/TSANet-integration-demo/src/test/java/com/tsanet/application/config/TsaNetApiConfigurationBeanTest.java
+++ b/TSANet-integration-demo/src/test/java/com/tsanet/application/config/TsaNetApiConfigurationBeanTest.java
@@ -1,0 +1,35 @@
+package com.tsanet.application.config;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import org.junit.jupiter.api.Test;
+
+/** The scripted demo refuses a plain-http base URL at startup unless the operator opted in, same rule as the console. */
+class TsaNetApiConfigurationBeanTest {
+
+    private static TsaNetApplicationProperties propertiesFor(String baseUrl, Boolean allowInsecureHttp) {
+        return new TsaNetApplicationProperties(
+            new TsaNetApplicationProperties.Api(baseUrl, allowInsecureHttp),
+            null,
+            new TsaNetApplicationProperties.Storage(System.getProperty("java.io.tmpdir") + "/tsanet-bean-test"),
+            null
+        );
+    }
+
+    @Test
+    void aPlainHttpBaseUrlFailsAtStartupWithTheFixInTheMessage() {
+        assertThatThrownBy(() -> new TsaNetApiConfigurationBean().tsaNetApiSessionFactory(propertiesFor("http://connect.example", null)))
+            .isInstanceOf(IllegalStateException.class)
+            .hasMessageContaining("https")
+            .hasMessageContaining("tsanet.api.allow-insecure-http=true");
+    }
+
+    @Test
+    void anHttpsBaseUrlAndAnOptedInHttpMockBothStart() {
+        assertThatCode(() -> new TsaNetApiConfigurationBean().tsaNetApiSessionFactory(propertiesFor("https://connect.example", null)))
+            .doesNotThrowAnyException();
+        assertThatCode(() -> new TsaNetApiConfigurationBean().tsaNetApiSessionFactory(propertiesFor("http://localhost:8080", true)))
+            .doesNotThrowAnyException();
+    }
+}
--- a/TSANet-integration-demo/src/test/resources/application.yml
+++ b/TSANet-integration-demo/src/test/resources/application.yml
@@ -1,6 +1,8 @@
 tsanet:
   api:
     base-url: "http://localhost:8080"
+    # The tests run against a local mock; the console's rule would otherwise refuse this URL.
+    allow-insecure-http: true
   storage:
     sqlite-path: "${java.io.tmpdir}/tsanet-test"
   accounts:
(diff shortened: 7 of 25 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `ConnectFacadeConfiguration.requireHttps` → delegates inline check to library helper — traces to change #5 (one https decision).
- `TsaNetApiConfigurationBean` https guard in bean → change #5 (scripted demo gains guard).
- `TsaNetApplicationProperties.Api` gains `allowInsecureHttp` + `insecureHttpAllowed()` → change #5, positional record component.
- `application.yml` (demo main + test) `allow-insecure-http` → change #5.
- `TsaNetApiConfigurationBeanTest` (new) → tests for change #5.
- `ConnectApiBaseUrl` (new) → change #5, the shared helper.
- `ConnectApiConnectivityInterceptor` body-read → change #3 (value-free 2xx body-read failure).
- `ConnectApiSessionStore` Snapshot rewrite → change #1 (atomic store).
- `OAuth401RetryInterceptor` `renewUnlessAlreadyRenewed` + `bearerSent` → change #2 (single-flight, prefix strip).
- `TokenManager` lock, `renew()`, `renewUnlessAlreadyRenewed`, `bearerOf` → change #2 + #6 (removal of `refreshAccessToken`).
- `DefaultTsaNetApiSession` backoff table + `completeLogin(boolean)` + `pause` → change #4 (/v1/me retry).
- All new/modified tests → trace to their respective changes.
- `DemoProperties` + `EnvironmentService` + demo-ui test setups → change #5 (demo-ui guard).

Every hunk traces. No scope creep found.

**Correctness:**

Traced the single-flight 401 path with two concurrent callers both holding bearer `stale`:
- Caller A enters `renewUnlessAlreadyRenewed("stale")`, takes lock. Store still has `stale` (expiry irrelevant per the `.equals(observedToken)` short-circuit), so `renew()` runs, stores `fresh`. Returns `fresh`.
- Caller B waits on lock, enters. `current.hasBearer()` true, `current.bearerToken()` = `fresh` ≠ `stale`, and `fresh` not expired → returns `fresh` with no network call. Correct — matches `aFourOhOneRenewalIsSkippedWhenAnotherCallerAlreadyRenewed`.

Edge in `bearerSent`: an `Authorization` header of exactly `"Bearer "` (trailing space, empty token) yields `substring(7)` = `""`, passed as `observedToken`. If the store also holds `""` (never happens — `hasBearer()` requires non-blank), no misfire. Benign.

`completeLogin` retry: `SocketTimeoutException`/`ConnectException` wrapped as `CONNECTIVITY` retry up to `UNATTENDED_CURRENT_USER_ATTEMPTS` (3), with `pause` restoring interrupt on `InterruptedException` and failing fast. A `PROBLEM`-kind (403) throws immediately and clears — verified by `aFailureTheApiAnsweredIsNotRetriedEvenUnattended`. The `catch (RuntimeException)` after `catch (ConnectApiException)` — since `ConnectApiException` is a `RuntimeException`, the more specific catch wins; non-Connect runtime exceptions clear-and-throw without retry. Correct ordering.

`saveUserContext` now synchronized and rebuilds the whole snapshot from `current` — preserves the logout-during-renewal window, which is explicitly documented on `clear()` and listed under "Not in this PR." Acceptable.

**Security:**

No secrets in the diff. The https guard is the security-relevant change and it hardens, not weakens: plain-http bearer transport now fails at startup unless an explicit opt-in is set. `insecureHttpAllowed()` defaults to false via `Boolean.TRUE.equals(null)`. Test configs set the opt-in only where a local mock is never actually contacted. The body-read change strengthens the case-token leak fix from #78 — the URL-bearing `ResourceAccessException` no longer escapes; verified by the test asserting the message contains neither `MARKER` nor `connect.example`.

**Regression risk:**

- `refreshAccessToken` removed. PR body's grep claims only two test `never()` verifications referenced it, both retargeted. The diff confirms `OAuth401RetryInterceptorTest`, `TokenManagerTest`, and `DefaultTsaNetApiSessionLoginTest` all updated to `renewUnlessAlreadyRenewed`. No production caller remains in the diff. Consistent with the enumeration.
- `TsaNetApplicationProperties.Api` and `DemoProperties` each gained a positional record component. PR body's grep lists five construction sites, all tests, all in this diff. The diff shows the demo-ui test constructors and the new bean test updated. Spring binds by name, so YAML binding is safe. The enumeration was produced and the sites decided — the positional-contract edge is swept.
- `ConnectApiSessionStore` public surface unchanged (getters return same Optionals); the added `snapshot()` is package-private for `TokenManager`. Consumers outside the package unaffected.
- `isExpired`/`isAuthorized` now judge against snapshot expiry — same semantics, now atomic. `itClearsAuthorizationOnLogout` extended to assert all fields clear.

**Recommendation:** Merge — every hunk traces, the concurrency and retry paths are correct and well-tested, and the security posture improves; Shawn's call on merge.

---

shawn-tsanet (commented):
## QA review of connection hardening follow-ups: single-flight renewal, atomic session store, value-free 2xx read failure, /v1/me retry, https guard in both demos (#79)

Separate-session QA pass at head `b32ce90`.

All five items land, and the concurrency work is the good kind: the tests fail first without the fix. Nothing below is new wrong behavior introduced here. The one thing I would fix before merge is cheap and it is a keyword; the rest is two javadocs that claim more than the code does.

Receipts: `mvn test` in the local clone at `b32ce90`, whose merge-base with `origin/main` is main's head `e39392e`: **connect-library 181 run, 0 failures, 1 skipped**, attachment-receiver 182 / 0 / 40, demo-ui 19 / 0, console 37 / 0, scripted demo 8 / 0. Every number matches the body's table. Same JDK 26 caveat as the previous reviews.

### Minimal-Diff Rule
- Pass. Twenty-four files over six commits, one per item plus the review commit, every hunk tracing to an item in issue 79. Removing `refreshAccessToken` is justified by its own grep and by the two retargeted test verifications.

### Correctness
- **`clear()` is the only mutator that does not take the monitor.** All three save methods are `synchronized`; `clear()` is a bare volatile assignment. A save's read-modify-write can straddle it: the save reads `current`, `clear()` installs `EMPTY`, the save writes a snapshot rebuilt from the pre-logout `current`, and the logout is undone. Adding `synchronized` closes exactly this interleaving and costs nothing.
- **Two javadocs added in this PR claim more than the code does.** Both describe the same area, so they are worth fixing together:
  - `OAuth401RetryInterceptor` says "the manager renews only if the store still holds that bearer". `renewUnlessAlreadyRenewed` renews in three cases: the store holds that bearer, the store is **empty**, or the store holds a different but **expired** bearer. The claim is right for one of the three.
  - `clear()` says a renewal "already past its network call" may repopulate the session. The window is wider in both directions. Narrower fix above aside, a 401 arriving *after* a logout on a client-credentials session starts a brand-new login: `supportsRefresh()` reads the injected config, not the store, so it stays true on an empty store, the interceptor calls through, `hasBearer()` is false, and the code falls to `renew()`. The logout is undone by a login that began after it. That behavior predates this PR (the old `refreshAccessToken` did the same), so it is not a regression, but the new comment reads as though it were covered. Either gate `renew()` on `hasBearer()` so the 401 simply passes through, or say plainly that a straggling 401 can re-authenticate a logged-out client-credentials session. No test covers the empty-store branch.
- **The single-flight logic is sound where I traced it.** Two callers that both see the token expired: the first renews under the lock, the second re-reads the snapshot, finds it unexpired and returns it with no second login. The 401 path keys on the bearer the rejected request actually carried, so a caller whose 401 was already answered by another renewal reuses that bearer, and a caller whose own bearer is still the stored one renews whatever the expiry claims. The eight-caller latch test with counted logins is the right proof, and the body reports it red with the lock removed. Lock ordering is consistently renewal lock then store monitor, never the reverse, so the two cannot deadlock.
- **Two bounded sharp edges in that comparison**, both worth a sentence of javadoc rather than a change. It keys on the token, not the principal, so a 401 raised under one principal can be answered with another's bearer; `supportsRefresh()` bounds this to a shared session where a login as the *configured* user overlaps an in-flight request issued under a different typed user, and the previous code reached the same outcome there by re-logging in, so nothing got worse. Separately, an identity provider that reissues the *same* JWT until it truly expires defeats the reuse check: each concurrent 401 caller finds the stored token equal to the one it sent and renews in turn, so the shared-renewal property degrades to one token fetch per caller in that case.
- The retry arithmetic is right: three attempts, indices 0 and 1 into the two-element table, and the third short-circuits before indexing. `ConnectApiException` is caught ahead of `RuntimeException`, so an API-answered failure never enters the retry.

### Security
- **The body-read change is correct and its precondition holds.** I checked the wiring rather than taking it from the body: `ConnectApiConnectivityInterceptor` is installed in exactly one place, and that place builds `new RestTemplate(new BufferingClientHttpRequestFactory(new SimpleClientHttpRequestFactory()))`. The test drives the real template and pins both edges, the extractor still receiving the whole body and an empty 204 not throwing.
- **No memory regression from the earlier read**, which is worth stating because it looks like one: that factory already buffered every response into a byte array, so the interceptor only moves the copy earlier in the same request. Large responses were being buffered before this PR too.
- The failure mode if the pairing were ever broken is silent rather than loud, since an unbuffered stream would be consumed and every body would arrive empty. The invariant lives only in prose on a public class. Making it package-private, or asserting the factory, would make it enforceable instead of remembered.
- The https helper is pure, its message names each caller's own opt-in, and all three apps now make one decision. The opt-in admitting any non-https host is stated in the message and in all three yml comments, which is the honest framing.

### Regressions
- The two record components added for the opt-in bind by name, and every construction site is a test, all updated here; the green reactor across five modules is the receipt that none was missed.
- Both demos now fail startup on a plain-http base URL, and `EnvironmentService` checks every configured environment rather than only the selected one, so an operator with one http entry cannot start even if they never choose it. That is the stricter and better reading; worth knowing it is the behavior.

### Operational
- The two operator-visible changes are stated in the body: both demos refuse plain http, and an unattended login can take up to 750 ms longer to fail.

### Nits
- The shared helper's missing-URL message lost the property name: the console used to say `tsanet.api.base-url is required` and now says `Connect API base URL is required`. The opt-out branch is parameterized but this one is not, so the message that exists to name the fix no longer names the key. No test asserts the old string, so nothing breaks. Passing the base-url property name alongside `optOutSetting` would restore it for all three callers.
- `savePassword` carries the previous snapshot's `accountId` forward, so a password login after a client-credentials session keeps the old account id beside the new principal. Pre-existing, now explicit and deliberately pinned by a test. Reads oddly next to the mode change, nothing more.

shawn-tsanet:
## Disposition of the QA reviews at `b32ce90`

Addressed in `2c67e95`. Receipts: whole reactor at `2c67e95`: **connect-library 182 run, 0 failures, 1 skipped** (one test added), attachment-receiver 182/0/40, demo-ui 19/0, console 37/0, scripted demo 8/0.

- **`clear()` does not take the monitor.** Fixed: `clear()` is `synchronized` like the three saves, so a save's read-modify-write can no longer straddle a logout. No deterministic test for the interleaving is possible without a hook inside the synchronized methods; the fix is the one-word one you named.
- **`OAuth401RetryInterceptor` javadoc claims one renewal case of three.** Fixed: it now names all three (rejected bearer still stored, empty store, stored bearer itself expired).
- **`clear()` javadoc reads as though the straggling-401 case were covered.** Fixed by stating it plainly rather than gating: the javadoc now says a straggling 401 on a client-credentials session finds an empty store and logs in again, and `renewUnlessAlreadyRenewed` documents the empty-store branch as the behavior the previous path also had. Chose documenting over gating `renew()` on `hasBearer()` because that would change what a request on a never-authenticated session does on a published artifact, which is outside #79. New test `aFourOhOneOnAnEmptyStoreLogsInAgainAsThePreviousPathDid` pins the branch so it is a decision, not an accident.
- **Two bounded sharp edges of keying on the token.** Both now in the `renewUnlessAlreadyRenewed` javadoc: the cross-principal answer bounded by `supportsRefresh()`, and the same-JWT reissue degrading shared renewal to one fetch per caller, never a failure.
- **The buffering invariant lives only in prose on a public class.** Fixed: `ConnectApiConnectivityInterceptor` is package-private, so only `ConnectApiRestTemplates.create()`, which installs the buffering factory, can add it. The javadoc says why.
- **Both demos fail startup on any http environment, not only the selected one.** Confirmed as the intended reading; stated in the PR body.

### Nits

- **Missing-URL message lost the property name.** Fixed: `requireHttps` takes the base-url setting name as well, so the console again says `tsanet.api.base-url is required`; the scripted demo says the same and the demo-ui names `tsanet.demo.environments.<env>.api-base-url`. The console test asserts the full string now.
- **`savePassword` carries the previous `accountId` forward.** Left as is; pre-existing and pinned by test. Noted in the body under "Not in this PR".

### Auto-QA (LOOKS GOOD)

Nothing to change.

PR #83, opened 2026-09-09
Publish attachment-receiver alongside connect-library on release (#82)

### LINKED ISSUE #82: Publish the attachment-receiver module in the next connect-library release
Part of epic #46. Unblocks `tsanetgit/Connect_Gateway#35` and `tsanetgit/Connect_Gateway#36`, which consume this module as a published artifact.

**Today.** `.github/workflows/publish.yml` deploys the parent pom and `connect-library` only (`mvn -pl .,connect-library deploy`), by design: the app and demo modules are never published. `attachment-receiver` (the `AttachmentStorage` SPI, the S3, Azure Files, Azure Blob and GCS adapters, the tenant config model and the go-live verifier; all merged on `main`) falls under that rule, so no release has ever carried it and GitHub Packages holds no `com.tsanet:attachment-receiver`. The gateway's attachments-only profile cannot build against it until one does.

**Change.**
- Add `attachment-receiver` to the reactor selection in the publish step, so a release deploys the parent pom, `connect-library` and `attachment-receiver`. The module has no dependency on `connect-library`, so nothing else moves.
- Update the workflow's header comment and the root pom's deploy-skip comment, which both state that only the library is published.
- The module's live adapter tests are gated on environment variables and skip in CI, so `deploy`'s full lifecycle gates the release on its unit and contract tests, the same way it gates on the library's.
- `skills/connect-sdk-deploy/references/consume-artifact.md` gains the second coordinate and one line on what the module is for (the gateway is the intended consumer; a member deploying its own receiver is the other).

**Acceptance.** The next release publishes `com.tsanet:attachment-receiver:<version>` alongside `com.tsanet:connect-library:<version>`, resolvable from GitHub Packages with a `read:packages` token; the release notes name the module and the four adapters; the gateway can declare the dependency at that version.

**Timing.** Rides the next release cut, which `main` already needs for the V2 attachment client (#77), the classified errors and renewal (#78, #80) and the Azure Blob adapter (#72); eight-plus commits sit on `main` past v1.0.0.

shawn-tsanet:
Closes #82. Unblocks `tsanetgit/Connect_Gateway#35` and `#36`, which consume the receiver as a published artifact.

## What changed

- **`publish.yml`**: the deploy step's reactor selection is `-pl .,connect-library,attachment-receiver`. The header comment, the step name and the step comment say the receiver is published too, that its environment-gated live tests skip in CI while its unit and contract tests gate the release, and that the app and demo modules are still never published.
- **Root pom**: the `distributionManagement` comment names the new selection.
- **Receiver pom description** (ships inside the published pom, and package versions are immutable): no longer calls the HTTPS ingest a gated concern under `Connect-API-Code#140`; says it is built in `tsanetgit/Connect_Gateway#35`, and that this is a plain library with no main class.
- **Docs**: `consume-artifact.md` gains `com.tsanet:attachment-receiver` (same repository and version, plain library, gateway as intended consumer, no dependency on `connect-library`); the deploy skill's module map row for the receiver says "Library only" instead of "Spring Boot jar" and points the ingest endpoint at the gateway.

Same version and same release as the library. No sources or javadoc jars, matching the library.

## Receipts

Rehearsed the workflow's two steps locally, exactly as they run in CI, against a file-based deployment target:

```
mvn versions:set -DnewVersion=0.0.1-TEST -DgenerateBackupPoms=false
mvn -pl .,connect-library,attachment-receiver deploy -Dmaven.deploy.skip=false -DaltDeploymentRepository=local::file://<scratch>/repo
```

- The file repository received `tsanet-client-parent`, `connect-library` and `attachment-receiver`, each at `0.0.1-TEST`, pom + jar + checksums.
- The receiver's deployed pom carries `<parent><version>0.0.1-TEST` (the workflow's `versions:set` has no module selection, so every module's parent reference is bumped) and the new description.
- The receiver jar: 29 classes, `META-INF/MANIFEST.MF` only under `META-INF`, no `BOOT-INF`, no `Main-Class`. A plain library jar.
- Receiver tests in that deploy run: **182 run, 0 failures, 40 skipped** (the environment-gated live adapter tests).
- Version restored afterward; `git diff` is the five files above and nothing else.

- **Prose claims, line that makes each true.** "Does not depend on `connect-library`": `grep -n connect-library attachment-receiver/pom.xml` returns nothing. "No main class": `grep -rln '@SpringBootApplication\|public static void main' attachment-receiver/src/main` returns nothing, and no `spring-boot-maven-plugin` in its pom. "Live tests skip in CI": the 40 skipped in the rehearsal, all under `@EnabledIfEnvironmentVariable`. "`versions:set` bumps the parent reference": the deployed receiver pom above.
- **Left-behind sweep** for statements that only the library is published: `git grep -n -i -e 'never published' -e 'library only' -e '-pl \.,connect-library' -e 'published alongside'`. Hits are the two files this PR updates, `verify.yml`'s "ones never published" (still true of the app and demo modules), a dated historical line in the root README about v0.1.0, a build-command comment in the skill, and the consume-artifact line about the parent pom (still true). Nothing stale left.
- **Guard / representation / schema / newly reachable failure**: not applicable; no code, no contract. The one new failure mode is intended: a receiver test failure now fails the release, the same rule the library already had.

## QA round at `a010000`

Separate-session QA: merge with notes; auto-QA: looks good. Second commit `bda310b` drops the stale `httpclient5` override (the factory test proves the platform pin links, and fails on the old 5.3 line with the exact `TlsSocketStrategy` error), adds the dependency-footprint sentence to the consumption reference, and records in the pom why all four adapters stay required. Disposition per note in the PR comments.

## Follow-up (not in this PR)

Per-provider packaging of the receiver (one artifact per adapter over a core SPI artifact, or classpath discovery of adapters), so the gateway profile takes only the backends its tenants use. File once `tsanetgit/Connect_Gateway#35` has a consumer shape.

## Advisor and gate

Advisor consulted twice. Orientation: verify `versions:set` carries no module selection (it does not), rehearse the exact deploy selection to a file repository, same version and release as the library, no sources/javadoc, and fix both the skill row and the pom description now because the description ships inside an immutable package version. Pre-done: three blocks resolved before opening: the dependency claim (grep receipt above), a present-tense contradiction between the pom description and the skill row's "not built yet" (both now say the endpoint is built in the gateway, with the issue number), and the left-behind sweep (above). Pre-PR gate: **LOOKS GOOD**, first run.

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -3,9 +3,10 @@ name: publish
 # A published GitHub Release is what cuts a version, deliberately rather than on
 # every push: consumers pin a version, so a library change reaches them only when
 # someone decides it should. Same pattern as Connect_Gateway's connector-core
-# publish. Publishes the root pom (connect-library's parent, which consumers must
-# be able to resolve) and connect-library ONLY; the app and demo modules are
-# never published.
+# publish. Publishes the root pom (the parent, which consumers must be able to
+# resolve), connect-library, and attachment-receiver (the storage SPI and adapters
+# the gateway's attachments-only profile will consume, tsanetgit/Connect_Gateway#35;
+# a plain library jar) ONLY; the app and demo modules are never published.
 on:
   release:
     types: [published]
@@ -92,14 +93,16 @@ jobs:
           mvn -B --no-transfer-progress versions:set \
             -DnewVersion="${{ steps.v.outputs.version }}" -DgenerateBackupPoms=false
 
-      # `deploy` runs the full lifecycle, so connect-library's tests gate the
-      # release: a broken client cannot be published. Reactor selection publishes
-      # the parent pom and the library only.
-      - name: Build, test and publish connect-library
+      # `deploy` runs the full lifecycle, so each published module's tests gate
+      # the release: a broken client cannot be published. Reactor selection
+      # publishes the parent pom, the library and the receiver only. The
+      # receiver's live adapter tests are gated on environment variables and skip
+      # here; its unit and contract tests run.
+      - name: Build, test and publish connect-library and attachment-receiver
         working-directory: Connect_SDK
         env:
           MAVEN_USERNAME: ${{ github.actor }}
           MAVEN_TOKEN: ${{ secrets.GITHUB_TOKEN }}
         # -Dmaven.deploy.skip=false arms the deploy that the root pom disarms by
         # default, so this workflow is the only path that publishes.
-        run: mvn -B --no-transfer-progress -pl .,connect-library deploy -Dmaven.deploy.skip=false
+        run: mvn -B --no-transfer-progress -pl .,connect-library,attachment-receiver deploy -Dmaven.deploy.skip=false
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -20,15 +20,6 @@
         tsanetgit/Connect_Gateway#35), and nothing in this module may assume its wire shape.
     </description>
 
-    <properties>
-        <!-- The Boot parent pins httpclient5 5.3.x; the AWS SDK's apache5 HTTP client
-             needs TlsSocketStrategy, which exists only from httpclient5 5.4. Without
-             this override every real S3Client construction dies with
-             NoClassDefFoundError - found by the live acceptance, invisible offline. -->
-        <httpclient5.version>5.6.1</httpclient5.version>
-        <httpcore5.version>5.4.2</httpcore5.version>
-    </properties>
-
     <dependencyManagement>
         <dependencies>
             <dependency>
@@ -55,6 +46,12 @@
         </dependencies>
     </dependencyManagement>
 
+    <!-- All four adapters are required, not optional: RegisteringStorageFactory links every
+         one of them, so a consumer takes all four cloud SDKs (about 170 runtime artifacts).
+         A per-provider split is a deliberate follow-up, not something to do by marking
+         these optional. The AWS SDK's apache5 client needs httpclient5 5.4 or later for
+         TlsSocketStrategy; the Boot 4 parent pins 5.6.x, so no override is needed here, and
+         RegisteringStorageFactoryTest constructs a real S3Client to prove the class links. -->
     <dependencies>
         <dependency>
             <groupId>software.amazon.awssdk</groupId>
--- a/attachment-receiver/pom.xml
+++ b/attachment-receiver/pom.xml
@@ -13,9 +13,11 @@
     <artifactId>attachment-receiver</artifactId>
     <name>attachment-receiver</name>
     <description>
-        Receive side of TSANet Connect case attachments: the AttachmentStorage SPI and its
-        storage adapters. The normalized HTTPS ingest endpoint is a separate, gated concern
-        (tsanetgit/Connect-API-Code#140) and nothing in this module may assume its wire shape.
+        Receive side of TSANet Connect case attachments: the AttachmentStorage SPI, its
+        storage adapters (S3, Azure Files, Azure Blob, GCS), the tenant config model and the
+        go-live verifier. A plain library with no main class. The HTTPS ingest endpoint that
+        feeds it is built in tsanetgit/Connect_Gateway (its attachments-only profile,
+        tsanetgit/Connect_Gateway#35), and nothing in this module may assume its wire shape.
     </description>
 
     <properties>
--- a/pom.xml
+++ b/pom.xml
@@ -38,8 +38,9 @@
     </properties>
 
     <!-- Where `mvn deploy` publishes. The workflow deploys the root pom (this file is
-         connect-library's parent, so consumers need it resolvable) plus connect-library
-         only, via reactor selection (-pl .,connect-library); the app and demo modules
+         the parent, so consumers need it resolvable) plus connect-library and
+         attachment-receiver only, via reactor selection
+         (-pl .,connect-library,attachment-receiver); the app and demo modules
          are never published. The server id `github` is wired to credentials by the
          publish workflow's setup-java step. -->
     <distributionManagement>
--- a/skills/connect-sdk-deploy/SKILL.md
+++ b/skills/connect-sdk-deploy/SKILL.md
@@ -33,7 +33,7 @@ the release version, branch names, and module list against the live repository.
 | `TSANet-integration-app` | Console reference app exposing the library as CLI commands (`login`, `requests`, `notes add`, `webhooks create`, ...) | Spring Boot jar |
 | `TSANet-integration-demo` | Scripted demo scenarios over the library | Spring Boot jar |
 | `demo-ui` | The branded web demo: dashboard, partner search, dynamic process forms, full case lifecycle from the browser. This is "the SDK demo" most people mean. | Spring Boot jar, port 8090 |
-| `attachment-receiver` | The storage half of receiving pushed files: a streaming storage SPI with AWS S3, Azure Files (backend id `azure`), Azure Blob Storage (backend id `azure_blob`), and Google Cloud Storage adapters, an encrypted per-tenant config store, and a go-live verifier. The HTTPS endpoint that accepts the push is not built yet; no standalone docs yet. | Spring Boot jar |
+| `attachment-receiver` | The storage half of receiving pushed files: a streaming storage SPI with AWS S3, Azure Files (backend id `azure`), Azure Blob Storage (backend id `azure_blob`), and Google Cloud Storage adapters, an encrypted per-tenant config store, and a go-live verifier. The HTTPS endpoint that accepts the push is being built in tsanetgit/Connect_Gateway's attachments-only profile (tsanetgit/Connect_Gateway#35), not here; no standalone docs yet. | Library only (published alongside connect-library) |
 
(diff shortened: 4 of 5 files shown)


shawn-tsanet (commented):
**Verdict:** `LOOKS GOOD`

**Minimal-Diff Rule check:**
- `publish.yml` header comment (L3-9): traces to "the header comment ... say the receiver is published too." ✓
- `publish.yml` deploy step name/comment + reactor selection (L92-108): traces to the core change "`-pl .,connect-library,attachment-receiver`." ✓
- `pom.xml` distributionManagement comment (L38-43): traces to "Root pom: the `distributionManagement` comment names the new selection." ✓
- `attachment-receiver/pom.xml` description (L13-20): traces to "Receiver pom description ... no longer calls the HTTPS ingest a gated concern ... plain library with no main class." ✓
- `SKILL.md` module map row (L33): traces to "the deploy skill's module map row ... says 'Library only.'" ✓
- `consume-artifact.md` new paragraph (L8+): traces to "`consume-artifact.md` gains `com.tsanet:attachment-receiver`." ✓

Every hunk traces. Six files claimed in prose ("five files" in the description body vs. stats "5 files" — but the diff shows 5 files: publish.yml, attachment-receiver/pom.xml, pom.xml, SKILL.md, consume-artifact.md). Consistent.

**Correctness:**
This is a docs+CI-config change with no runtime code. The one behavioral change is the reactor selection `-pl .,connect-library,attachment-receiver` (L108). Tracing the release path: on `release: published`, `versions:set` (no `-pl`, so all module parent refs bump) → `deploy` on the three-module reactor. The author rehearsed exactly this against a file repo and reported the receiver pom, jar (no BOOT-INF/Main-Class), and checksums landed at the test version, with 182 tests run / 40 skipped. The 40 skips being `@EnabledIfEnvironmentVariable` live tests is the intended CI behavior. The one newly-reachable failure — a receiver test failure now fails the release — is named and is the same rule the library already carries. No edge cases in a comment/selection change.

None.

**Security:**
No secrets in the diff. `MAVEN_TOKEN: ${{ secrets.GITHUB_TOKEN }}` is unchanged from prior. Publishing an additional module widens what lands in the public/registry package, but the author confirmed the receiver jar contains no BOOT-INF and no embedded config — 29 classes + MANIFEST only. No credential or tenant-secret material is compiled into the jar (the config store is encrypted-at-rest per-tenant, not baked in). No injection surface.

None.

**Regression risk:**
- The `versions:set` step (L92) is unchanged and already bumps every module's parent reference regardless of reactor selection — so the receiver's parent version was already being set correctly before this PR; the only new thing is that the receiver now also gets *deployed*. Confirmed by the rehearsal's deployed-pom `<parent><version>0.0.1-TEST`.
- Consumers of the publish workflow: this is the single publish path (armed only by `-Dmaven.deploy.skip=false`). Adding a module to the reactor cannot un-publish connect-library or the parent — both remain in the selection. App/demo modules stay excluded (`verify.yml`'s "ones never published" still true, per the sweep).
- Left-behind vocabulary sweep for "never published" / "library only" / "-pl .,connect-library" was run and every hit decided — the two updated files, plus stale-but-correct historical/parent-pom lines. That's the exact edge (vocabulary gained a member) and it was swept.
- Immutable-package concern handled: the pom description ships inside the version and can't be edited post-release, so fixing it in the same PR that first publishes it is correct sequencing.

None.

**Recommendation:** Merge — this is a clean, fully-traced docs+reactor change with a rehearsed deploy receipt, and the one new failure mode is intended.

Advisory only — I do not merge. Note this is still marked **Draft**; flip it out of draft before merging.

---

shawn-tsanet (commented):
## QA review of Publish attachment-receiver alongside connect-library on release (#82)

Separate-session QA pass at head `a010000`.

The reactor change is right and the docs sweep is thorough. Two things about the module's pom are worth settling before this release, because a published version is immutable and this PR is already editing that pom.

Receipts: I read the poms at the PR head and measured the two claims below in a throwaway worktree; I did not repeat the deploy rehearsal, so the file-repository receipt in the body stands as the author's. Reading the poms confirms its preconditions: the receiver declares no dependency on `connect-library`, has no `<build>` section and so no `spring-boot-maven-plugin`, which is what makes a plain library jar rather than a Boot fat jar, and sets no `maven.deploy.skip` of its own, so it inherits the root's disarm and the workflow's `-Dmaven.deploy.skip=false` arms it.

### Minimal-Diff Rule
- Pass. Five files, each traceable to an item in issue 82: the reactor selection and its three comments, the receiver's description, and the two doc surfaces.

### Correctness
- **The `httpclient5` override in the receiver's pom is stale, and it now downgrades rather than upgrades.** Its comment says "The Boot parent pins httpclient5 5.3.x ... Without this override every real S3Client construction dies". Measured on this head: `help:evaluate` on a module *without* the override resolves `httpclient5.version` to **5.6.4**, which is what Boot 4.1.1 pins; with the override the receiver resolves **5.6.1**, and the dependency tree confirms `httpclient5:5.6.1` arriving under `software.amazon.awssdk:apache5-client`. Both are well above the 5.4 the comment cares about, so the failure it describes cannot happen under this repo's Boot version, and the override's only remaining effect is to hold the module one patch line behind the platform. The Boot 3.3.x the comment describes is two majors ago. Either drop the override, or say what it now does. This is the moment because the pom ships inside the package.
- **There is no consumer trap here, which is worth stating** because the comment reads like there is one. A dependency's property does not travel: a consumer resolves transitive versions from its own dependency management, so a Boot 4 consumer gets 5.6.4 and a non-Boot consumer gets whatever the AWS SDK asks for. Nothing about publishing this module obliges a consumer to repeat the override.
- The three BOM imports ship in the published pom, which is what lets the version-less dependency entries resolve for a consumer. That is the correct shape for this module and needs no change.

### Regressions
- `versions:set` carries no module selection, so every module's parent reference bumps; the body's deployed-pom receipt shows that. The selection `-pl .,connect-library,attachment-receiver` is coherent because the receiver depends on neither of the other two.
- The one new failure mode is intended and named: a receiver test failure now fails the release.

### Operational
- **Publishing this module hands every consumer a 171-artifact runtime tree.** Measured with `dependency:list -DincludeScope=runtime` on this head: 171 entries, including all four cloud stacks at once, 37 Netty artifacts, 30 AWS SDK, 17 gRPC, 9 OpenTelemetry, 8 Azure and 8 Google Cloud. The gateway's attachments-only profile, the intended consumer, will need whichever backends its tenants actually use rather than all four.
- The new `consume-artifact.md` paragraph gives the coordinates, says there is no main class and names the intended consumer, which is the right content, but the footprint is the single most useful fact for someone deciding to take the dependency and it is not there. One sentence would cover it.
- Whether to mark the four adapter dependencies `<optional>` is a real design call rather than an obvious fix, since `RegisteringStorageFactory` names all four adapters and a consumer omitting one changes what that class can link. Worth deciding deliberately before the first published version sets the expectation, rather than after.

### Nits
- None.

shawn-tsanet:
## Disposition of the QA reviews at `a010000`

Addressed in `bda310b`. Receipts: `mvn -pl attachment-receiver test` at `bda310b`: **182 run, 0 failures, 40 skipped** (the environment-gated live tests); `dependency:list -DincludeScope=runtime` now shows `httpclient5:5.6.4` and `httpcore5:5.4.3`, the platform's pins.

- **The `httpclient5` override is stale and downgrades.** Confirmed by measurement (`help:evaluate` on the receiver: 5.6.1 with the override, 5.6.4 without, matching your numbers) and removed, together with `httpcore5.version` and the Boot 3.3 comment. The proof it can go: `RegisteringStorageFactoryTest.s3WithRegionBucketAndStaticKeysBuildsS3Adapter` constructs a real `S3Client`, which is where the original failure lived. Pinned back to the 5.3 line it fails with `NoClassDefFoundError: org/apache/hc/client5/http/ssl/TlsSocketStrategy`, the exact failure the old comment described; on the platform pin it passes. So the test discriminates the hazard and the platform pin satisfies it. A comment on the dependencies block now records the 5.4 floor and points at that test, so the next reader does not reintroduce the override.
- **No consumer trap.** Agreed and worth having in the record; the comment says a consumer resolves its own versions.
- **The BOM imports ship in the published pom.** Agreed, unchanged.
- **171-artifact runtime tree; footprint absent from the consumption reference.** Measured 170 here (compile plus runtime lines), same picture: all four cloud stacks, 37 Netty, 30 AWS SDK, 17 gRPC, 9 OpenTelemetry. One sentence added to `consume-artifact.md` stating the footprint and why (the factory links every adapter).
- **Whether to mark the adapters optional: decide deliberately before the first published version.** Decided: not optional in this release. `RegisteringStorageFactory` imports and constructs all four adapters, so a consumer that omitted one would fail to link that class rather than lose a feature. The dependencies comment states this, and a per-provider split (separate artifacts, or a factory that resolves adapters by presence) is recorded as the follow-up below. The first published version therefore sets the expectation "all four, required", which is the honest one for the code as it is.

### Auto-QA (LOOKS GOOD)

Nothing to change. The draft flag is Shawn's to flip at merge.

### Follow-up (not in this PR)

Per-provider packaging of the receiver: either one artifact per adapter with the SPI in a core artifact, or a factory that discovers adapters present on the classpath, so the gateway's attachments-only profile takes only the backends its tenants use. To be filed once `tsanetgit/Connect_Gateway#35` has a concrete consumer shape to size it against.

PR #84, opened 2026-09-09
Remove the always-test createRequest shims deprecated for removal since 0.1.0 (2.0.0)

shawn-tsanet:
For the 2.0.0 major. Removes the two `CollaborationRequestsFacade.createRequest` default overloads that silently passed `testSubmission=true`, deprecated `forRemoval` since 0.1.0 and promised for removal at the next major, plus the test that pinned that behavior. Both forms of `createRequest` now take the explicit per-call flag only. The deploy skill and the consumption reference say the shims are gone rather than pending.

## Receipts

- Every caller in this repository already passes the flag: console `CreateRequestExecutor`, demo-ui `CollaborationRequestsController` (both forms), scripted demo `AcmeCreateCollaborationRequestsScenario`; each call read, each ends in `true`.
- `mvn test-compile` over the whole reactor succeeds without the shims (main and test sources of all five modules).
- connect-library: **181 run, 0 failures, 1 skipped** (one fewer than main: the removed shim test).
- `tsanetgit/Connect_Gateway`'s engine never calls `createRequest` (grep of `gateway-engine` main); `tsanetgit/Fin-Intercom_App` consumes the engine, not the library's facade.
- Stragglers: `git grep -i 'always-test\|shim\|forRemoval'` over skills, connect-library, docs and README returns only the two updated doc lines. No other `forRemoval` member exists in the library.
- The facade's `Map` and `CollaborationRequestFormTemplateDto` imports are still used by the remaining six-argument overload.

- **Vocabulary lost a member; old form grepped:** the four-argument and five-argument-without-flag `createRequest` calls. In-repo: none (compile receipt). Known consumers: none (above). An un-inventoried consumer that passed no flag fails at compile against 2.0.0, which is the promised outcome of a `forRemoval` deprecation honored at its boundary.
- **Prose claims:** "deprecated since 0.1.0" per the 0.1.0 release notes and the javadoc removed here; "promised for removal at the next release boundary" is the javadoc's own words.
- Guard, representation, schema, newly reachable failure: not applicable.

## Follow-up (not in this PR)

`CollaborationRequestStatusDto` keeps a plain `@Deprecated` compatibility constructor from before the `testCase` field existed, whose javadoc also says "scheduled for removal at the next release boundary". Left in: it is not `forRemoval`, so consumers were never machine-warned; its old-arity uses have not been enumerated across the reactor's tests (14 files construct the DTO) or consumers; and it degrades safely (null `testCase` reads as unknown, never as real). To be filed: mark it `@Deprecated(forRemoval = true, since = "0.1.0")` with 3.0.0 as the target, and enumerate the call sites, so the javadoc stops promising a boundary that passed.

## Advisor and gate

Advisor consulted once, pre-done (the change is two deletions and two doc lines): ship as scoped, keep the DTO constructor out with a filed follow-up, run the straggler grep (clean, above). Pre-PR gate: **LOOKS GOOD**, first run. Release notes for 2.0.0 carry this as breaking change 5.

### DIFF
--- a/connect-library/src/main/java/com/tsanet/api/facade/CollaborationRequestsFacade.java
+++ b/connect-library/src/main/java/com/tsanet/api/facade/CollaborationRequestsFacade.java
@@ -54,41 +54,6 @@ public interface CollaborationRequestsFacade {
         boolean testSubmission
     );
 
-    /**
-     * Compatibility form from before the flag was per-call; keeps the pre-existing
-     * behavior of always creating a TEST submission.
-     *
-     * @deprecated pass {@code testSubmission} explicitly; scheduled for removal at
-     *     the next release boundary.
-     */
-    @Deprecated(forRemoval = true)
-    default CollaborationRequestStatusDto createRequest(
-        long receiverCompanyId,
-        String caseNumber,
-        String summary,
-        String description
-    ) {
-        return createRequest(receiverCompanyId, caseNumber, summary, description, true);
-    }
-
-    /**
-     * Compatibility form from before the flag was per-call; keeps the pre-existing
-     * behavior of always creating a TEST submission.
-     *
-     * @deprecated pass {@code testSubmission} explicitly; scheduled for removal at
-     *     the next release boundary.
-     */
-    @Deprecated(forRemoval = true)
-    default CollaborationRequestStatusDto createRequest(
-        CollaborationRequestFormTemplateDto formTemplate,
-        String caseNumber,
-        String summary,
-        String description,
-        Map<Long, String> customFieldValues
-    ) {
-        return createRequest(formTemplate, caseNumber, summary, description, customFieldValues, true);
-    }
-
     CollaborationRequestStatusDto fetchRequestByToken(String caseToken);
 
     void syncAllDetails();
--- a/connect-library/src/test/java/com/tsanet/api/internal/DefaultTsaNetApiSessionCreateRequestTest.java
+++ b/connect-library/src/test/java/com/tsanet/api/internal/DefaultTsaNetApiSessionCreateRequestTest.java
@@ -107,13 +107,6 @@ class DefaultTsaNetApiSessionCreateRequestTest {
         assertThat(capturedForm().getTestSubmission()).isTrue();
     }
 
-    @Test
-    void theDeprecatedOverloadKeepsTheOldAlwaysTestBehavior() {
-        session.createRequest(template(), "CASE-1", "s", "d", Map.of());
-
-        assertThat(capturedForm().getTestSubmission()).isTrue();
-    }
-
     private CollaborationRequestDTO capturedForm() {
         ArgumentCaptor<CollaborationRequestDTO> captor = ArgumentCaptor.forClass(CollaborationRequestDTO.class);
         verify(collaborationGateway).createCollaborationRequest(captor.capture());
--- a/skills/connect-sdk-deploy/SKILL.md
+++ b/skills/connect-sdk-deploy/SKILL.md
@@ -112,7 +112,9 @@ These bite regardless of path, and none of them are visible in the OpenAPI schem
 - **Test-mode asymmetry.** You *write* the `testSubmission` flag when creating a case
   but *read* it back as `testCase`. The library's create API takes an explicit
   per-call `testSubmission` flag with no silent default; the older always-test
-  signatures, deprecated in 0.1.0, were removed in 2.0.0.
+  signatures, deprecated in 0.1.0, were removed in 2.0.0. A consumer that bumps to
+  2.0.0 without recompiling hits `NoSuchMethodError` at the create call, which is that
+  removal, not a platform fault.
 - **Engineer emails must be on the member's registered domain.** The platform rejects
   collaboration requests whose engineer email is off the member company's registered
   domain. This is a business rule, not a schema rule, and the error is not
--- a/skills/connect-sdk-deploy/SKILL.md
+++ b/skills/connect-sdk-deploy/SKILL.md
@@ -111,9 +111,8 @@ These bite regardless of path, and none of them are visible in the OpenAPI schem
   member burn an afternoon debugging their network for what is a typo in a password.
 - **Test-mode asymmetry.** You *write* the `testSubmission` flag when creating a case
   but *read* it back as `testCase`. The library's create API takes an explicit
-  per-call `testSubmission` flag with no silent default; older always-test method
-  signatures remain as deprecated shims (still present in 1.0.0) and will be removed
-  at a later release boundary.
+  per-call `testSubmission` flag with no silent default; the older always-test
+  signatures, deprecated in 0.1.0, were removed in 2.0.0.
 - **Engineer emails must be on the member's registered domain.** The platform rejects
   collaboration requests whose engineer email is off the member company's registered
   domain. This is a business rule, not a schema rule, and the error is not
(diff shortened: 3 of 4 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- **Hunk 1** (`CollaborationRequestsFacade.java`, −35): removes the two `@Deprecated(forRemoval = true)` default `createRequest` overloads. Traces directly to the PR's core claim — remove the always-test shims for 2.0.0.
- **Hunk 2** (`DefaultTsaNetApiSessionCreateRequestTest.java`, −7): removes `theDeprecatedOverloadKeepsTheOldAlwaysTestBehavior`, the test that pinned the removed behavior. Traces to "plus the test that pinned that behavior."
- **Hunk 3** (`connect-sdk-deploy/SKILL.md`): rewrites the test-mode-asymmetry note from "shims still present in 1.0.0, removed at a later boundary" to "removed in 2.0.0." Traces to "the deploy skill … say the shims are gone rather than pending."
- **Hunk 4** (`consume-artifact.md`): same doc-status flip from pending to removed. Traces to "the consumption reference." Note the prose still reads "**Known deprecation:**" is dropped and reworded to a present-tense removal statement — consistent with the change.

All four hunks trace. No incidental tidying.

**Correctness:**

The two deleted methods were pure `default` delegations passing `true`. The five-arg-with-flag and six-arg-with-flag overloads they delegated to remain (the diff context shows the six-arg form ending at line 55). Trace: a caller invoking the four-arg form (`createRequest(id, case, summary, desc)`) no longer resolves to any method — it fails at compile against 2.0.0. That is the promised `forRemoval` outcome. In-repo compile receipt (`mvn test-compile` green, 181 run) confirms no surviving in-tree caller used the removed arity. The removed test asserted `getTestSubmission().isTrue()` on the now-deleted overload; deleting the test alongside the method is correct — no dangling reference.

Import-retention claim ("`Map` and `CollaborationRequestFormTemplateDto` still used by the six-arg overload") is plausible but not visible in the diff; a stale-import warning would surface at compile, and compile is green, so this is consistent.

**Security:** None. Deletions only; no secrets, no input handling, no auth surface touched.

**Regression risk:**

- This is a **breaking API change** for external consumers passing no flag. The PR enumerates known consumers — console `CreateRequestExecutor`, demo-ui controller, scripted scenario (all pass `true`), plus `Connect_Gateway`'s engine (grep: never calls `createRequest`) and `Fin-Intercom_App` (consumes engine, not facade). An un-inventoried external caller on the old arity would fail at compile against 2.0.0 — the intended, documented boundary behavior. This is correctly gated behind the major version.
- **Vocabulary member removed** — the four-arg and five-arg-no-flag call forms. Old form grepped in-repo (compile receipt) and across known consumers; clean. Straggler grep (`always-test|shim|forRemoval`) returns only the two updated doc lines and no other `forRemoval` member. Edge swept.
- The follow-up `CollaborationRequestStatusDto` compatibility constructor is correctly held out (not `forRemoval`, degrades safely) with a filed follow-up.

**Recommendation:** Merge — a correctly-scoped breaking removal behind the 2.0.0 boundary, with in-repo and consumer receipts backing every claim; the DTO-constructor follow-up is properly deferred, not a blocker.

One operational note for Shawn (does not move the verdict): this must land in the 2.0.0 release and the release notes must carry it as a breaking change — the PR states it's breaking change 5, so that's tracked.

---

shawn-tsanet (commented):
## QA review of Remove the always-test createRequest shims deprecated for removal since 0.1.0 (2.0.0)

Separate-session QA pass at head `efd20bb`.

The removal is correct and honors a real deprecation. Every factual claim in the body checked out. The notes are all on the consumer-facing upgrade note, which is the deliverable half of a PR whose code half is a deletion.

Receipts, all run against this head:

- Full reactor green: **connect-library 181 run, 0 failures, 1 skipped**, one fewer than main exactly as stated, plus attachment-receiver 182 / 0 / 40, demo-ui 19 / 0, console 37 / 0, scripted demo 8 / 0.
- "Deprecated since 0.1.0" holds. `git log -S forRemoval` puts the annotation's introduction at `66f44fe`, and `git merge-base --is-ancestor 66f44fe v0.1.0` confirms it shipped in that release, six hours before the tag.
- "No other `forRemoval` member exists in the library" holds: the grep over `connect-library/src/main` now returns nothing.
- The follow-up's premise holds too: exactly one plain `@Deprecated` remains in the library, `CollaborationRequestStatusDto:26`, the compatibility constructor the body files for later.

### Minimal-Diff Rule
- Pass. Four files: the two shims, the test that pinned them, and the two doc surfaces that described them as pending.

### Correctness
- The removed methods were pure `default` delegations passing `true`, and the explicit-flag overloads they delegated to remain. Removing the test alongside the method is right; it asserted only the deleted behavior.
- **The annotation never carried `since`.** It was a bare `@Deprecated(forRemoval = true)`, so a consumer's compiler and IDE warned "deprecated, marked for removal" with no version attached. The claim "since 0.1.0" is true historically and is stated here for the first time. Worth noting because the body's own follow-up prescribes `since = "0.1.0"` for the DTO constructor, which is the right lesson drawn from this one.

### Regressions
- None in the repository: the reactor compiles and passes without the shims, so no in-tree caller used the removed arity.
- **Removing a `default` interface method breaks binary compatibility as well as source.** A consumer who bumps the version without recompiling gets a `NoSuchMethodError` when the call runs, not a compile error. The upgrade note says the signatures are removed, which is accurate, but naming the runtime symptom is what saves someone the debugging trip. That is the stated purpose of the surrounding gotcha list.

### Operational
- **The docs name 2.0.0 before any tag exists.** The publish workflow takes the version from the release tag, so nothing in the repository makes 2.0.0 true; these two lines are a forward promise. `tsanetgit/Connect_SDK#83` rides the same release, so the two are coupled: whoever cuts it must use 2.0.0 or both PRs' doc changes become wrong. A sequencing dependency rather than a defect, worth stating so it is not discovered at tag time.

### Nits
- **The new Upgrading entry is the only one not keyed by its version.** The section reads as a migration index: `**0.2.0** added ...`, `**1.0.0** moved to ...`. The new bullet opens with prose and reaches 2.0.0 mid-sentence, so a reader scanning for their target version does not find it. Leading with `**2.0.0**` matches the neighbours and costs nothing.

shawn-tsanet:
## Disposition of the QA reviews at `efd20bb`

Addressed in `046d03d` (docs only; the code half is unchanged).

- **Binary as well as source compatibility.** Right, and the more useful half of the note: a consumer that bumps without recompiling gets `NoSuchMethodError` at the create call. Both doc surfaces now say so, the consumption reference's Upgrading bullet and the skill's gotcha list, with "recompile against 2.0.0" as the action.
- **The annotation never carried `since`.** Acknowledged; the "since 0.1.0" claim is historical (your `git log -S` receipt is the one I would cite) and is now stated in the docs rather than the code. The DTO-constructor follow-up prescribes `since` for exactly that reason.
- **The docs name 2.0.0 before a tag exists.** Acknowledged as a sequencing dependency, not a defect. The release draft's checklist records it: the cut that carries #83 and this PR must be tagged `v2.0.0`, or these lines become false. The version proposal is already 2.0.0 on independent grounds (the `ConnectApiException` change), so the dependency is satisfied by the plan rather than by luck.
- **Upgrading bullet not keyed by version.** Fixed: it now opens with `**2.0.0**` like its neighbours.

### Auto-QA (LOOKS GOOD)

Nothing to change. The release-notes coupling it names is the same dependency above.

PR #85, opened 2026-09-09
security-audit: read unquoted config values and short secrets in the credentials check (#63)

### LINKED ISSUE #63: security-audit: the embedded-credentials check cannot see unquoted config values, so committed secrets in .yml/.properties pass silently
Source: QA review of `tsanetgit/Connect_SDK#62`, 2026-08-14. Follow-up to `tsanetgit/Connect_SDK#61` acceptance item 5, refrained after probing the detector: the miss that let #61 ship is not specific to CRaSH, and the narrow check #61 asks for would not close it.

## Problem

`scripts/security-audit.py`'s embedded-credentials check reports PASS over committed credentials that sit in the config formats it already scans.

**Receipt.** Run the audit against `origin/main` at `eca8225`, which carries `auth-password: crash` in `TSANet-integration-app/src/main/resources/application.yml` and `crash.auth-password=crash` in `tsanet-demo-crash-ssh.properties`:

```
PASS  [credentials] no embedded credentials in source
0 fail, 2 warn, 5 pass, 0 skip
```

**This is not a file-coverage gap.** `SOURCE_EXT` (`security-audit.py:35`) already includes `.properties`, `.yaml`, and `.yml`. Both files were opened and scanned. The pattern is what failed.

The keyword pattern requires the value to be **quoted** and **at least 8 characters**:

```python
r"(?i)(password|client[_-]?secret|api[_-]?key|apikey)\s*[=:]\s*[\"']([^\"'{$<\n][^\"'\n]{7,})[\"']"
```

Probing that regex directly isolates two independent causes:

```
MISS   | main application.yml (real)        | auth-password: crash
MISS   | main properties file (real)        | crash.auth-password=crash
MISS   | same value, but quoted             | auth-password: "crash"
MATCH  | quoted AND >=8 chars               | auth-password: "crashcrash"
MISS   | unquoted AND >=8 chars             | auth-password: crashcrash
MISS   | realistic yaml secret, unquoted    | client-secret: Xq7bTn2LpR9dWv4A
MATCH  | same secret, quoted                | client-secret: "Xq7bTn2LpR9dWv4A"
```

1. **Unquoted values are invisible.** The `["']` requirement means the detector only sees quoted literals, but unquoted is the native convention for `.properties`, `.yml`, and `.yaml`: three of the seven extensions it scans. The last two rows are the load-bearing pair. A realistic, high-entropy `client-secret` committed in YAML passes silently, and only adding quotes around the identical value makes the detector fire. Nothing about that miss involves CRaSH; any secret in a Spring config file is equally invisible.

2. **The 8-character floor exempts short secrets.** `{7,}` after one leading character sets a minimum length of 8, so `crash` (5) slips under even when quoted. Short, guessable credentials are the higher-severity ones, so the floor is inverted with respect to risk.

## Why not the narrow check

`#61`'s acceptance item 5 asks for a rule that FAILs on a shipped `crash.enabled=true` alongside a committed `crash.auth-password`. That would satisfy the checkbox and leave the class open: every unquoted YAML or properties credential in the repo stays undetected.

It would also be the **third** narrow patch to this one detector. The comment at `security-audit.py:33` records the first, adding `.sh` to `SOURCE_EXT` after `scripts/simple-test.sh` "carried a real credential past every prior run of this audit". That earlier fix widened *which files* are read; this issue is that widening the file list does not help when the value matcher cannot parse the format. Patching per-instance a third time predicts a fourth.

## Fix

Make the value matcher format-aware rather than quote-dependent:

- Accept unquoted values in `key=value` (`.properties`) and `key: value` (`.yml`, `.yaml`) form, terminated by end-of-line or a comment marker, in addition to the current quoted form.
- Drop or substantially lower the length floor for these matches. The floor exists to suppress false positives, but the existing `placeholders` regex and the `names_itself` guard are the filters actually suited to that job, and length is what let `crash` through.

Expect this to surface more candidates on `.yml` and `.properties` than the current pattern does. That is the point, but the tuning belongs in `placeholders` / `names_itself`, not in a length threshold, so the acceptance below pins both directions: it must catch the real cases and must not start failing the clean tree.

## Acceptance

- [ ] `auth-password: crash`, unquoted in a `.yml` fixture, FAILs the credentials check.
- [ ] `client-secret: Xq7bTn2LpR9dWv4A`, unquoted in a `.yml` fixture, FAILs the credentials check.
- [ ] A regression fixture carrying the `#61` pair (`crash.enabled=true` plus `crash.auth-password=crash`) FAILs. This satisfies `#61`'s acceptance item 5 as a consequence of fixing the class rather than as a CRaSH-specific rule.
- [ ] Every credential shape the current pattern already catches still FAILs, so widening loses no existing coverage.
- [ ] The audit still reports `0 fail` against `Connect_SDK` at `main` once `#62` has merged, so the widened matcher does not fire on the real tree.
- [ ] `security-audit.yml` continues to pass in CI on a clean branch.

shawn-tsanet:
Closes #63. The embedded-credentials check matched only a quoted value of eight or more characters, so a secret written the native way in a `.properties`, `.yml`, `.yaml` or `.sh` file passed silently, and so did any short secret even when quoted. That is how #61's `crash`/`crash` shipped past every run of this audit.

## What changed

- **Two keyword forms replace the one quoted-only regex.** The quoted form applies to every scanned file with the length floor dropped: length never separated a secret from a placeholder, and the floor exempted exactly the short, guessable secrets that matter most. The unquoted form applies to the config formats only (`CONFIG_EXT`): an optional YAML list dash, an optional `export` or `local`, a possibly dotted or prefixed key, the value to end of line or a `#` / `!` comment, with a first character that is never a quote, `$`, `{`, `<` or `!`, so references, templates, placeholders and YAML tags do not match at all rather than being filtered afterward.
- **One new guard** beside the existing placeholder and self-naming guards: a bare `ALL_CAPS` identifier with at least one underscore names a variable (setup-java's `server-password: MAVEN_TOKEN`) and carries no entropy. The underscore is required on purpose so an all-caps secret like `HUNTER2` is still reported; an all-caps secret with an underscore is silenced by design, and the comment says so.
- **The placeholder test reads the value, not the whole match**, for the keyword forms. The unquoted form's whole match includes the key prefix and any trailing comment, so `test.password=realsecret` and `password: realsecret  # test env` must not be silenced by a word that is not part of the secret. The token, PAT and private-key patterns still test the whole match.
- **`scripts/test_security_audit.py`**, 24 probes on `unittest`, importing the hyphenated script and running the check on fixture trees in temp dirs: the issue's three acceptance shapes; the other misses (quoted short secret, shell `export` and `local`, dotted key, YAML list item, value before a comment, placeholder word in the comment or key prefix); every shape the old matcher already caught (quoted Java literal, Slack token, GitHub PAT, private key); the references and placeholders that must stay quiet (`${...}`, `$VAR`, `!vault`, `{{ }}`, the env-var name, a key naming itself, `"test-password"`, Java's unquoted assignment); and one subprocess run pinning exit code 1, which the workflow keys on. The Slack fixture is built from adjacent literals so the committed source never holds the contiguous token shape.
- **`security-audit.yml`** runs the probes before the audit, so a detector regression fails CI the way a finding does.

## Receipts

- Probes: **24 pass** on this script; **13 fail** against the script on `main` (the seven miss shapes, `HUNTER2`, the two placeholder-scope cases, the list item, `local`, and the exit-code probe), the other eleven pass on both (retained shapes and quiet cases).
- The final regexes over the whole tree: four candidates, all silenced (three by the old guards, one by the new), zero fire. `python3 scripts/security-audit.py` on this branch: `PASS [credentials] no embedded credentials in source`, `0 fail, 3 warn` (the three warns are pre-existing: unpinned actions in the other two workflows, credential-shaped names near log calls, non-local `http://` in tests and the receiver's https guard).
- Every shape in #63's probe table now matches: `auth-password: crash`, `crash.auth-password=crash`, `client-secret: Xq7bTn2LpR9dWv4A`, and the quoted `"crash"`.
- No document describes the old check (`git grep` for the floor, "quoted", "embedded credential" and "security-audit" over markdown returns nothing), so nothing is left behind.

## Acceptance, item by item

1. `auth-password: crash` unquoted in a `.yml` fixture FAILs: `test_unquoted_short_password_in_yaml_fails`.
2. `client-secret: Xq7bTn2LpR9dWv4A` unquoted in a `.yml` fixture FAILs: `test_unquoted_high_entropy_client_secret_in_yaml_fails`.
3. The #61 pair in `.properties` FAILs: `test_the_61_pair_in_properties_fails`, as a consequence of the class fix, not a CRaSH rule.
4. Every previously caught shape still FAILs: the four `still_fails` probes.
5. `0 fail` against `main`'s tree: the run above (this branch is `main` plus this change).
6. `security-audit.yml` passes on a clean branch: this PR's own run.

- **Guard added, why not a representation change**: the guard is the detector; the representation change (parsing YAML and properties structurally) would trade a regex the reviewer can read for a parser with its own gaps, for no coverage the two forms do not already give.
- **Third fix to one class**: this is the second change to the detector after the `.sh` widening, and the issue asked for the class fix rather than a third narrow patch; the two forms are that.
- **Newly reachable failure**: a FAIL from this check exits 1 and fails the workflow; the probe step fails the same job on a detector regression. Both end in the job's error annotation.
- **Prose claims**: "four candidates, zero fire" is the tree scan with the shipped regexes; "13 fail on main's script" is the run against `git show origin/main:scripts/security-audit.py`.
- Schema/contract: none.

## Known gaps, pre-existing, not this PR

**Key vocabulary.** The keys are `password`, `passwd`, `pwd`, `client-secret`, `api-key` and `apikey`. Bare `secret`, `token`, `auth-token`, `access-key` and `credentials` are not matched, so a high-entropy value under one of those names passes. Each needs a noise check on real trees before it joins the list. Same shape of gap as #63 one level over: #63 fixed value syntax; this is key vocabulary. One follow-up issue with the JSON gap below, on ask.

**JSON.** JSON's `"password": "x"` never matched the keyword form (the closing quote after the key blocks `\s*[=:]`) and still does not; `.json` is scanned only for the token, PAT and key patterns. Filed separately if wanted.

## Advisor and gate

Advisor consulted twice. Orientation: drop the floor entirely, require an underscore in the variable-name guard, add `export`, exclude `$`/`{`/`<` at the first character, re-run the tree scan with the final regexes (all done). Pre-done: one block, the placeholder scan reading the whole match (fixed, two probes), plus defusing the Slack literal and the list-item/`local` forms (done). QA round at `7022d7c` (separate session: merge with notes; auto-QA: looks good) answered in `475cde9`: placeholder words anchored to word edges, `passwd` and `pwd` added, three probes; 27 probes, 15 red on main. Pre-PR gate: **LOOKS GOOD** on the first run, before the pre-done edits; **CONCERNS** on the second, at this head, with one ask: confirm `security-audit.py` accepts `--repo-root`, which the exit-code probe depends on. It does: `main()` defines `--repo-root` (default: the repository root), and the probe passes on that flag. One correction to the commit message: it says twenty probes and nine red on `main`; the counts after the pre-done edits are 24 and 13, as stated above. Review is CI plus a QA-persona pass from a separate session.

### DIFF
--- a/.github/workflows/security-audit.yml
+++ b/.github/workflows/security-audit.yml
@@ -36,6 +36,12 @@ jobs:
       - name: Check out the repository
         uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
 
+      # The detector's own probes run first: a regression in what the audit
+      # can see must fail CI the same way a finding does, or a silent miss
+      # (the class #63 closed) ships as a green run.
+      - name: Probe the audit's detectors
+        run: python3 -m unittest -v scripts/test_security_audit.py
+
       - name: Run mechanical security audit
         shell: bash
         run: |
--- a/scripts/security-audit.py
+++ b/scripts/security-audit.py
@@ -154,8 +154,11 @@ def names_a_variable(match):
     return bool(VARIABLE_NAME.match(match.group(2).strip()))
 
 
-# The credential-shaped keys, shared by both keyword forms below.
-CREDENTIAL_KEY = r"(password|client[_-]?secret|api[_-]?key|apikey)"
+# The credential-shaped keys, shared by both keyword forms below. `passwd` and
+# `pwd` are the two spellings that cost no precision; `secret`, `token`,
+# `access-key` and `credentials` on their own are a wider vocabulary that needs
+# a noise check on real trees first, and they are disclosed as a gap in #63's PR.
+CREDENTIAL_KEY = r"(password|passwd|pwd|client[_-]?secret|api[_-]?key|apikey)"
 
 # Quoted form, any scanned file. The value must be single-line: without
 # excluding newlines a prompt such as System.out.print("Password: ") matches
@@ -188,7 +191,14 @@ def check_no_embedded_secrets(root):
         (re.compile(r"-----BEGIN (RSA |EC )?PRIVATE KEY-----"), "private key"),
     ]
     config_patterns = [(UNQUOTED_CREDENTIAL, "credential-shaped config value")]
-    placeholders = re.compile(r"(?i)your|example|changeme|placeholder|xxx|\.\.\.|\$\{|<.+>|test|dummy|sample")
+    # Placeholder WORDS are anchored to word edges: `test-password` and
+    # `your-secret` are placeholders, `Xq7test2LpR9dWv4A` is a secret that
+    # happens to contain one. With the length floor gone this guard is what
+    # separates a secret from a placeholder, so it must not match inside one.
+    # The reference shapes (`...`, `${`, `<...>`) have no word edges to anchor.
+    placeholders = re.compile(
+        r"(?i)(?<![a-z0-9])(your|example|changeme|placeholder|xxx|test|dummy|sample)(?![a-z0-9])"
+        r"|\.\.\.|\$\{|<.+>")
     hits = []
     for rel, body in walk_sources(root):
         applicable = patterns + (config_patterns if rel.endswith(CONFIG_EXT) else [])
--- a/scripts/security-audit.py
+++ b/scripts/security-audit.py
@@ -33,6 +33,9 @@ POM_NS = {"m": "http://maven.apache.org/POM/4.0.0"}
 # login to sit, and omitting them was not a neutral gap: scripts/simple-test.sh
 # carried a real credential past every prior run of this audit.
 SOURCE_EXT = (".java", ".xml", ".properties", ".yaml", ".yml", ".json", ".sh")
+# Formats whose native convention is an UNQUOTED value: key=value, key: value,
+# export KEY=value. A quote-dependent matcher is blind to all three (#63).
+CONFIG_EXT = (".properties", ".yaml", ".yml", ".sh")
 SKIP_DIRS = {".git", "target", "node_modules", ".idea"}
 
 
@@ -128,7 +131,7 @@ def names_itself(match):
     stripped keeps those quiet without weakening the check, because any real
     secret differs from its own key name.
 
-    Only applies to the keyword pattern, which is the one with two groups.
+    Only applies to the keyword patterns, which are the ones with two groups.
     """
     if match.re.groups < 2:
         return False
@@ -136,26 +139,72 @@ def names_itself(match):
     return normalize(match.group(1)) == normalize(match.group(2))
 
 
+# A value that is a bare ALL_CAPS identifier with at least one underscore names
+# a variable (setup-java's `server-password: MAVEN_TOKEN`, an `export X=$Y`
+# chain), and carries no entropy of its own. The underscore is required on
+# purpose: an all-caps secret with no underscore (HUNTER2) is still reported,
+# and an all-caps secret WITH an underscore is silenced by design; that shape
+# is far more often a reference than a credential.
+VARIABLE_NAME = re.compile(r"^[A-Z][A-Z0-9]*(?:_[A-Z0-9]+)+$")
+
+
+def names_a_variable(match):
+    if match.re.groups < 2:
+        return False
+    return bool(VARIABLE_NAME.match(match.group(2).strip()))
+
+
+# The credential-shaped keys, shared by both keyword forms below.
+CREDENTIAL_KEY = r"(password|client[_-]?secret|api[_-]?key|apikey)"
+
+# Quoted form, any scanned file. The value must be single-line: without
+# excluding newlines a prompt such as System.out.print("Password: ") matches
+# across the following lines and reports a literal that does not exist. No
+# length floor: the old {7,} exempted exactly the short, guessable secrets
+# that matter most, and length is not what separates a secret from a
+# placeholder; the guards below are (#63).
+QUOTED_CREDENTIAL = re.compile(
+    r"(?i)" + CREDENTIAL_KEY + r"\s*[=:]\s*[\"']([^\"'{$<\n][^\"'\n]*)[\"']")
+
+# Unquoted form, CONFIG_EXT only: an optional YAML list dash, an optional
+# `export` or `local`, a key that may be dotted or prefixed
+# (crash.auth-password, TSANET_API_PASSWORD), the value running to end of line
+# or to a `#` / `!` comment. The first character
+# rejects a quote (the quoted form's job), `$` (a shell or Spring reference),
+# `{` (a template) and `<` (a placeholder), and `!` (a YAML tag such as
+# `!vault`). In sh and YAML `!` is not a comment marker, so a value containing
+# one is reported truncated rather than missed.
+UNQUOTED_CREDENTIAL = re.compile(
+    r"(?im)^[ \t]*(?:-[ \t]+)?(?:(?:export|local)[ \t]+)?(?:[\w.\-]*[._-])?" + CREDENTIAL_KEY
+    + r"[ \t]*[=:][ \t]*([^\s\"'#!${<][^#\n]*?)[ \t]*(?:[#!].*)?$")
+
+
 def check_no_embedded_secrets(root):
     cat = "credentials"
     patterns = [
-        # The value must be single-line: without excluding newlines a prompt
-        # such as System.out.print("Password: ") matches across the following
-        # lines and reports a literal that does not exist.
-        (re.compile(r"(?i)(password|client[_-]?secret|api[_-]?key|apikey)\s*[=:]\s*[\"']([^\"'{$<\n][^\"'\n]{7,})[\"']"),
-         "credential-shaped literal"),
+        (QUOTED_CREDENTIAL, "credential-shaped literal"),
         (re.compile(r"xox[baprs]-[A-Za-z0-9-]{10,}"), "Slack token"),
         (re.compile(r"ghp_[A-Za-z0-9]{20,}"), "GitHub PAT"),
         (re.compile(r"-----BEGIN (RSA |EC )?PRIVATE KEY-----"), "private key"),
     ]
+    config_patterns = [(UNQUOTED_CREDENTIAL, "credential-shaped config value")]
     placeholders = re.compile(r"(?i)your|example|changeme|placeholder|xxx|\.\.\.|\$\{|<.+>|test|dummy|sample")
     hits = []
     for rel, body in walk_sources(root):
-        for pat, label in patterns:
+        applicable = patterns + (config_patterns if rel.endswith(CONFIG_EXT) else [])
+        for pat, label in applicable:
             for m in pat.finditer(body):
-                if placeholders.search(m.group(0)):
+                # The placeholder test reads the VALUE for the keyword forms. The
+                # unquoted form's whole match also holds the key prefix and any
+                # trailing comment, and `test.password=realsecret` or
+                # `password: realsecret  # test env` must not be silenced by a
+                # word that is not part of the secret.
+                candidate = m.group(2) if m.re.groups >= 2 else m.group(0)
+                if placeholders.search(candidate):
                     continue
                 if names_itself(m):
+                    continue
+                if names_a_variable(m):
                     continue
                 hits.append(f"{rel}: {label}")
     if hits:
--- a/scripts/test_security_audit.py
+++ b/scripts/test_security_audit.py
@@ -89,6 +89,22 @@ class EmbeddedCredentialsProbe(unittest.TestCase):
     def test_a_placeholder_word_in_the_key_prefix_does_not_silence_the_value(self):
         self.assertFails("app.properties", "test.password=Xq7bTn2LpR9dWv4A\n")
 
+    def test_a_placeholder_word_inside_a_secret_does_not_silence_it(self):
+        # The guard is what separates a secret from a placeholder now that the
+        # length floor is gone, so it must match placeholder WORDS, not substrings.
+        self.assertFails("application.yml", "client-secret: Xq7test2LpR9dWv4A\n")
+        self.assertFails("application.yml", "client-secret: myXXXsecretvalue1\n")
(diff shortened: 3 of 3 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**

- `security-audit.yml` new step (lines +39–44): probes run before the audit. Traces to "runs the probes before the audit, so a detector regression fails CI."
- `CONFIG_EXT` addition (security-audit.py +36–38): traces to the "unquoted form applies to config formats only" claim.
- Docstring "keyword pattern"→"keyword patterns" (line 131): traces to there now being two keyword forms. Trivial, in scope.
- `VARIABLE_NAME` + `names_a_variable` (+142–152): the "one new guard" for `ALL_CAPS_WITH_UNDERSCORE`. Traces.
- `CREDENTIAL_KEY`, `QUOTED_CREDENTIAL`, `UNQUOTED_CREDENTIAL` (+155–182): the "two keyword forms replace the one quoted-only regex." Traces.
- `check_no_embedded_secrets` body rewrite (+189–208): `applicable` gating on `CONFIG_EXT`, `candidate` value-vs-whole-match placeholder test, new guard call. All trace to described changes.
- `test_security_audit.py` (new, 161 lines): traces to "24 probes."

Every hunk traces. No scope creep.

**Correctness:**

Traced `test.password=Xq7bTn2LpR9dWv4A` in a `.properties` (the placeholder-in-key-prefix case, the one that's easy to get wrong):
- `UNQUOTED_CREDENTIAL` matches: prefix `test.` consumed by `(?:[\w.\-]*[._-])?`, key `password` in group 1, value `Xq7bTn2LpR9dWv4A` in group 2.
- `candidate = m.group(2)` = `Xq7bTn2LpR9dWv4A` — the `test` in the prefix is not in group 2, so `placeholders.search` does not silence it. Correct — this is exactly the bug the PR claims to fix, and `test_a_placeholder_word_in_the_key_prefix_does_not_silence_the_value` pins it.

Traced `server-password: MAVEN_TOKEN`: unquoted matches, group 2 = `MAVEN_TOKEN`, `VARIABLE_NAME` matches (`MAVEN` + `_TOKEN`), silenced by `names_a_variable`. Correct.

One edge I checked on the `candidate` logic: for the Slack/PAT/private-key patterns `m.re.groups` is 0, so `candidate = m.group(0)` (whole match) — placeholder test still runs on the whole token. Unchanged behavior for those. Good.

The `names_a_variable` guard requires `match.re.groups < 2` → return False, mirroring `names_itself`. Both keyword forms have 2 groups, the others have 0/1, so the guard only touches keyword hits. Correct.

The exit-code probe depends on `--repo-root`; PR body confirms `main()` defines it and the probe passes. Consistent with the diff (`subprocess.run([..., "--repo-root", root])`).

**Security:** None. Fixtures use throwaway values (`hunter2`, `Xq7bTn2LpR9dWv4A`) written at runtime; the Slack token is split across adjacent literals so the committed source never holds the contiguous shape. No real credentials in the diff. This PR tightens a security detector.

**Regression risk:**

Consumers of `check_no_embedded_secrets`: the `credentials` check in `security-audit.py`, invoked by `security-audit.yml`. The change only *widens* what fires and adds guards to suppress references. Risk is false positives on the existing tree — PR reports the full-tree scan yields "0 fail, 3 warn (pre-existing)" and "four candidates, all silenced." The three retained token/PAT/key patterns are untouched (still group-0), so no regression there.

The known gap (`.json` `"password":"x"` still unmatched by the keyword form) is pre-existing, explicitly declared, and out of scope. Not a regression introduced here.

Blast-radius edges: positional-contract change (group indexing now used via `m.group(2)` for both keyword forms) is asserted by probes covering both quoted and unquoted forms; the newly reachable failure (FAIL→exit 1) is pinned by `ExitCodeContract`. Swept.

**Operational:** No env var, no schema, no new dependency, no LaunchAgent restart. CI gains a `python3 -m unittest` step ahead of the audit. `.env.example`/`pyproject.toml`/`schema.sql` unaffected — none needed.

**Nits:** None worth holding.

**Recommendation:** Merge — the diff traces cleanly, the placeholder-scope and variable-name guards are correct and each pinned by a probe, and no consumer regresses.

---

shawn-tsanet (commented):
## QA review of security-audit: read unquoted config values and short secrets in the credentials check (#63)

Separate-session QA pass at head `7022d7c`.

The fix is correct, the probe suite genuinely discriminates the defect, and running the probes ahead of the audit in CI is the right structural move. Two residual blind spots are worth naming, and one of them matters more after this change than before it.

Receipts, all run by me against this head:

- The 24 probes pass on this script.
- **Red-first verified**: the same 24 probes against `origin/main`'s `security-audit.py` produce **13 failures**, exactly the count the body claims. The suite pins the defect rather than the implementation.
- The new audit over the real tree at this head: `PASS [credentials] no embedded credentials in source`, `0 fail, 3 warn, 4 pass`, which is acceptance item 5 and matches the body's three pre-existing warns.

### Minimal-Diff Rule
- Pass. Three files: the two matcher forms plus the new guard, the probe suite, and the workflow step that runs it.

### Correctness
- Splitting one quote-dependent regex into a quoted form for every scanned file and an unquoted form for the config formats is the right shape, and excluding the reference and template first characters inside the pattern rather than filtering afterward is what keeps the guards from having to carry that work.
- Reading the placeholder test from the value rather than the whole match is a real fix in its own right: without it the unquoted form's key prefix and trailing comment could silence a live secret, and the two probes for that are the ones I would have asked for.
- **The placeholder guard matches substrings anywhere in the value, and this change leans on it harder.** Removing the length floor is right, and the stated reasoning is that the guards, not length, separate a secret from a placeholder. That relocation makes the guard's precision load-bearing. It is an unanchored substring match over six words, so a high-entropy secret that merely contains one of them is silent. Demonstrated on this script:

  | fixture | verdict |
  |---|---|
  | `client-secret: Xq7test2LpR9dWv4A` | PASS |
  | `client-secret: myXXXsecretvalue1` | PASS |
  | `password: sampleXq7bTn2LpR9` | PASS |
  | `password: Xq7bTn2LpR9dWv4A` | FAIL |

  The last row is the same shape without the embedded word, so the word is what silences the first three. Anchoring the test to the whole value, or to word boundaries, would close it without touching anything else this PR did.

### Regressions
- None: the four previously caught shapes still fail, pinned by their own probes, and I confirmed the real tree stays at `0 fail`.

### Operational
- **The key vocabulary is now the larger gap, and it is unnamed.** The value syntax is format-aware after this change while the key list is still four literals, so realistic key names carry a high-entropy value straight through. All of these PASS on this script:

  `secret: <value>`, `token: <value>`, `auth-token: <value>`, `db.passwd=<value>`, `db.pwd=<value>`, `access-key: <value>`, `credentials: <value>`

  This is pre-existing and outside `tsanetgit/Connect_SDK#63`'s scope, so it does not belong in this diff. It belongs in the body next to the JSON gap, which is disclosed the same way. Worth saying because the issue's own argument was that patching one instance predicts a fourth patch: the class fixed here is value syntax, and this is the same shape of gap one level over. Adding `passwd` and `pwd` is nearly free; `secret` and `token` alone would need a noise check first.
- Confirmed as documented, no action: the ALL_CAPS guard does silence a real secret that carries an underscore (`password: S3CRET_VALUE_X9` passes), and JSON's quoted form is still unmatched. Both are stated in the body, and I verified both rather than assuming.

### Nits
- Building the Slack fixture from adjacent literals so the committed source never holds the contiguous token shape is a good detail, and the comment explaining why is worth keeping.

shawn-tsanet:
## Disposition of the QA reviews at `7022d7c`

Addressed in `475cde9`. Receipts: **27 probes** pass on this script, **15** fail against `main`'s (your 13 plus the two new probe methods that cover the notes below); the audit over the tree at `475cde9` still reports `0 fail, 3 warn`.

- **The placeholder guard matches substrings anywhere in the value, and this change leans on it harder.** Right, and the sharper point is that dropping the length floor made the guard load-bearing. Fixed: the placeholder WORDS (`your`, `example`, `changeme`, `placeholder`, `xxx`, `test`, `dummy`, `sample`) are anchored to word edges with lookarounds, so `test-password` and `your-secret-here` stay quiet and a secret that merely contains one of them is reported. The reference shapes (`...`, `${`, `<...>`) keep their unanchored match since they have no word edges. Your four rows on this script: the three embedded-word values now FAIL and the plain value still FAILs; `"test-password"` still PASSes. Pinned by `test_a_placeholder_word_inside_a_secret_does_not_silence_it` (your three fixtures verbatim) and `test_placeholder_words_at_word_edges_still_stay_quiet`. One consequence stated on purpose: a value like `testing123` is no longer silenced, since `test` inside `testing` is not a placeholder word; the tree carries no such fixture.
- **The key vocabulary is the larger gap, unnamed.** Agreed on both halves. `passwd` and `pwd` are added (probe: `test_passwd_and_pwd_spellings_fail`; the tree stays at 0 fail). `secret`, `token`, `auth-token`, `access-key` and `credentials` on their own stay out of this PR and are now disclosed in the body beside the JSON gap, with the reason: each needs a noise check on real trees before it joins, or the check starts firing on `token: ${...}`-shaped configuration that is not a credential. Both gaps go into one follow-up issue when Shawn asks for it.
- **The ALL_CAPS-with-underscore guard and the JSON quoted form**: confirmed as documented, no change; thank you for verifying both rather than assuming.

### Auto-QA (LOOKS GOOD)

Nothing to change.

shawn-tsanet:
The placeholder-substring note from the QA review is filed as `tsanetgit/Connect_SDK#86` (security, 09-2026 Release, matching `tsanetgit/Connect_SDK#63`), so it does not ride this PR. Nothing here needs to change for it: narrowing the placeholder test from the whole match to the value is what this PR did, and it is strictly better. The issue covers the residual, that the vocabulary half is still an unanchored substring match, and that removing the length floor is what makes the guard's precision the check's precision.

PR #88, opened 2026-09-10
Build the two library mappers with JsonMapper and test the primitive-null setting in both apps

### LINKED ISSUE #87: Jackson 3 follow-ups: build the two library mappers with JsonMapper, and test the primitive-null setting in both apps
## Why

Two follow-ups from the Jackson 3 migration (`tsanetgit/Connect_SDK#67`, released as connect-library 1.0.0) were carried on `tsanetgit/Fin-Intercom_App#76` as "tracked elsewhere" with no issue behind them. This issue is that home. Both are small and both are about the same thing: Jackson 3 defaults that the migration worked around rather than adopted.

## 1. `new ObjectMapper()` to `JsonMapper` in library main code

Jackson 3's builder-first mapper is `tools.jackson.databind.json.JsonMapper`; `ObjectMapper` remains as the runtime type but `new ObjectMapper()` is the Jackson 2 idiom and bypasses the builder where per-mapper configuration belongs. On `main` today, library main code has two:

```
connect-library/src/main/java/com/tsanet/api/connectapi/internal/OAuthTokenGateway.java:18
connect-library/src/main/java/com/tsanet/api/webhook/WebhookPayloadParser.java:8
```

Move both to `JsonMapper.builder()...build()`, with any feature the code relies on stated on the builder rather than inherited from the default. Consumers are not affected (both mappers are private statics). The Fin connector has zero occurrences in main code (five in tests, left alone).

## 2. A test that asserts the primitive-null setting

Jackson 3 flipped `FAIL_ON_NULL_FOR_PRIMITIVES` to true. The migration turned it back off in `TSANet-integration-app` and `TSANet-integration-demo` `application.yml` (`fail-on-null-for-primitives: false`, with a comment on why) and `CollaborationRequestFormRepository` handles the same default in code. The QA round on #67 caught a sibling flip (`FAIL_ON_UNKNOWN_PROPERTIES`) only because a test failed; the primitive-null setting has no test, so a future Boot or Jackson bump that changes how the property binds would pass silently. Add a test per app that deserializes a body with a `null` primitive through the app's injected mapper and asserts it binds to the default rather than throwing.

## Acceptance

- [ ] `git grep 'new ObjectMapper()' -- connect-library/src/main` returns nothing
- [ ] connect-library test suite green with the two mappers rebuilt
- [ ] A test in each of the two apps that fails when `fail-on-null-for-primitives` is removed from that app's `application.yml` (red-probed once, then committed green)

## References

- Origin: `tsanetgit/Fin-Intercom_App#76`; migration: `tsanetgit/Connect_SDK#67`

shawn-tsanet:
Closes tsanetgit/Connect_SDK#87. The two Jackson 3 follow-ups carried on tsanetgit/Fin-Intercom_App#76.

## What

- **connect-library**, `OAuthTokenGateway` and `WebhookPayloadParser`: `new ObjectMapper()` becomes `JsonMapper.builder().build()`, initializer and import only. Field names, types and visibility unchanged.
- **TSANet-integration-app** and **TSANet-integration-demo**: one new `@SpringBootTest` each, `JacksonPrimitiveNullConfigTest`, that deserializes `{"count":null}` into an `int` record component through the Boot-injected `tools.jackson.databind.ObjectMapper` and asserts `0`. That pins the `fail-on-null-for-primitives: false` line in each main `application.yml`, which nothing tested.

No yml, dependency, or existing-test change. 4 files, +89/-2.

## The design point a reviewer should check

Both apps ship a `src/test/resources/application.yml`. On the classpath it shadows the main one, so a plain `@SpringBootTest` in either app never loads the main file, and a test written the obvious way would pin nothing. The new tests set

```
spring.config.location=file:./src/main/resources/application.yml,file:./src/test/resources/application.yml
```

Later locations override earlier ones, so every test override (mock base URL, tmp storage, CLI/setup/sync off) still applies, and the main file supplies the setting under test. `spring.config.location` replaces the default search locations entirely; that is intended and stated in the javadoc. The `file:` paths resolve against the module directory, surefire's default working directory in a reactor or `-pl` run; a location that does not resolve fails the context loudly (`ConfigDataLocationNotFoundException`), never a false pass.

## Receipts (this session, JDK 21, reactor build)

| Claim | Receipt |
|---|---|
| Green | `mvn -pl connect-library,TSANet-integration-app,TSANet-integration-demo -am test`: BUILD SUCCESS; connect-library 181 tests (1 skipped, pre-existing), app 9, demo 38; both new tests ran and passed. Re-run on the committed state after the last javadoc edit: green. |
| Red probe, app | Main yml line removed, test run through the reactor: `MismatchedInput Cannot map null into type int (set DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES to 'false' to allow)`. File restored, `git status` clean on it. |
| Red probe, demo | Same procedure, same error. Restored. |
| Acceptance grep | `git grep 'new ObjectMapper()' -- connect-library/src/main` is empty. No occurrence remains anywhere in the repo's Java. |
| Library change is behavior-neutral | Both classes call only `readTree` (grep: one call each, no `readValue`/`convertValue`/`treeToValue`). `WebhookPayloadDto` is four `String` components; the token gateway takes `expires_in` via `asLong(3600)` on the tree node. No primitive is ever bound through these mappers, so the Jackson 3 defaults they now carry unchanged, including `FAIL_ON_NULL_FOR_PRIMITIVES` on, cannot fire. No builder feature set, on purpose. |
| Existing coverage of the two classes still green | `WebhookPayloadParserTest` 2, `OAuthTokenGatewayTest` 1, `TokenManagerTest` in the green run. |

## Blast radius (generated)

`scripts/pre-pr-artifacts.py --base origin/main`: no detector fired (no positional contract, vocabulary, guard, check or duplication change). Judgment lines:

- **Guard / third+ fix / abstraction introduced:** none.
- **Newly reachable failure:** none in production code. In tests, a missing main yml line now fails `JacksonPrimitiveNullConfigTest` with the Jackson message above; the receiving handler is surefire, the outcome is a red build.
- **Schema, contract, or stored shape:** none.
- **Prose claims, with the line that makes each true:** "test-resources application.yml shadows the main one": both files exist at those paths and Spring loads the first classpath match. "Later locations override earlier ones": Spring Boot config-data semantics for `spring.config.location`; the red/green pair above is the empirical receipt. "Jackson 2 default": `FAIL_ON_NULL_FOR_PRIMITIVES` was off in 2.x, on in 3.x, as the main yml comment records. "Remove that line and this test goes red": the two red-probe rows.

## Pre-PR gate

`scripts/qa-review.py --local`: LOOKS GOOD. Its one unverified edge, whether the main yml is self-sufficient when loaded as the base under the test overrides, is answered by the green run: both contexts boot with exactly that layering.

## Not done here

- Test webEnvironment differs between the apps (`MOCK` default in the app, `NONE` in the demo) because each mirrors its module's existing `@SpringBootTest` convention; the app's test yml also sets `web-application-type: none`, so the effective environment is the same.
- IDE runs whose working directory is not the module directory will fail the new tests at context start; that is the loud failure mode, and surefire is the contract.

### DIFF
--- a/TSANet-integration-app/src/test/java/com/tsanet/facade/config/JacksonPrimitiveNullConfigTest.java
+++ b/TSANet-integration-app/src/test/java/com/tsanet/facade/config/JacksonPrimitiveNullConfigTest.java
@@ -1,0 +1,42 @@
+package com.tsanet.facade.config;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import tools.jackson.databind.ObjectMapper;
+
+/**
+ * Pins the Jackson 3 default flip that the Boot 4 upgrade worked around: main
+ * application.yml sets {@code fail-on-null-for-primitives: false} so an explicit
+ * JSON null on a primitive still binds to the Jackson 2 default instead of
+ * rejecting the payload. Nothing else exercises that line (Connect_SDK#87).
+ *
+ * <p>The test-resources application.yml shadows the main one on the classpath,
+ * so a plain {@code @SpringBootTest} would never load the setting under test.
+ * {@code spring.config.location} lists the main file first and the test file
+ * second: later locations override earlier ones, so the test overrides (mock
+ * base URL, tmp storage, CLI off) still apply while the main file supplies the
+ * Jackson setting. Remove that line from the main file and this test goes red.
+ * The {@code file:} paths resolve against the module directory, surefire's default
+ * working directory; {@code spring.config.location} replaces the default search
+ * locations entirely, which is intended here.
+ */
+@SpringBootTest(properties = {
+    "spring.config.location=file:./src/main/resources/application.yml,file:./src/test/resources/application.yml"
+})
+class JacksonPrimitiveNullConfigTest {
+
+    record Probe(int count) {
+    }
+
+    @Autowired
+    private ObjectMapper mapper;
+
+    @Test
+    void anExplicitNullOnAPrimitiveBindsToTheDefaultInsteadOfThrowing() {
+        Probe probe = mapper.readValue("{\"count\":null}", Probe.class);
+        assertThat(probe.count()).isZero();
+    }
+}
--- a/TSANet-integration-demo/src/test/java/com/tsanet/application/config/JacksonPrimitiveNullConfigTest.java
+++ b/TSANet-integration-demo/src/test/java/com/tsanet/application/config/JacksonPrimitiveNullConfigTest.java
@@ -1,0 +1,43 @@
+package com.tsanet.application.config;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import tools.jackson.databind.ObjectMapper;
+
+/**
+ * Pins the Jackson 3 default flip that the Boot 4 upgrade worked around: main
+ * application.yml sets {@code fail-on-null-for-primitives: false} so an explicit
+ * JSON null on a primitive still binds to the Jackson 2 default instead of
+ * rejecting the payload. Nothing else exercises that line (Connect_SDK#87).
+ *
+ * <p>The test-resources application.yml shadows the main one on the classpath,
+ * so a plain {@code @SpringBootTest} would never load the setting under test.
+ * {@code spring.config.location} lists the main file first and the test file
+ * second: later locations override earlier ones, so the test overrides (mock
+ * base URL, tmp storage, setup and sync off) still apply while the main file
+ * supplies the Jackson setting. Remove that line from the main file and this
+ * test goes red.
+ * The {@code file:} paths resolve against the module directory, surefire's default
+ * working directory; {@code spring.config.location} replaces the default search
+ * locations entirely, which is intended here.
+ */
+@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.NONE, properties = {
+    "spring.config.location=file:./src/main/resources/application.yml,file:./src/test/resources/application.yml"
+})
+class JacksonPrimitiveNullConfigTest {
+
+    record Probe(int count) {
+    }
+
+    @Autowired
+    private ObjectMapper mapper;
+
+    @Test
+    void anExplicitNullOnAPrimitiveBindsToTheDefaultInsteadOfThrowing() {
+        Probe probe = mapper.readValue("{\"count\":null}", Probe.class);
+        assertThat(probe.count()).isZero();
+    }
+}
--- a/connect-library/src/main/java/com/tsanet/api/connectapi/internal/OAuthTokenGateway.java
+++ b/connect-library/src/main/java/com/tsanet/api/connectapi/internal/OAuthTokenGateway.java
@@ -2,6 +2,7 @@ package com.tsanet.api.connectapi.internal;
 
 import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.json.JsonMapper;
 import com.tsanet.api.auth.ClientCredentialsAuthConfig;
 import com.tsanet.api.auth.OAuthAccessToken;
 import java.util.List;
@@ -15,7 +16,7 @@ import org.springframework.web.client.RestClientResponseException;
 import org.springframework.web.client.RestTemplate;
 
 public class OAuthTokenGateway {
-    private static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
+    private static final ObjectMapper OBJECT_MAPPER = JsonMapper.builder().build();
 
     private final RestTemplate restTemplate;
 
--- a/connect-library/src/main/java/com/tsanet/api/webhook/WebhookPayloadParser.java
+++ b/connect-library/src/main/java/com/tsanet/api/webhook/WebhookPayloadParser.java
@@ -2,10 +2,11 @@ package com.tsanet.api.webhook;
 
 import tools.jackson.databind.JsonNode;
 import tools.jackson.databind.ObjectMapper;
+import tools.jackson.databind.json.JsonMapper;
 import com.tsanet.api.connectapi.dto.WebhookPayloadDto;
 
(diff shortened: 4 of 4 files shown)


shawn-tsanet (commented):
**Verdict:** LOOKS GOOD

**Minimal-Diff Rule check:**
- Hunk 1 (`TSANet-integration-app/.../JacksonPrimitiveNullConfigTest.java`, +42): new test pinning the `fail-on-null-for-primitives: false` line — traces to "one new `@SpringBootTest` each" in the PR body. ✅
- Hunk 2 (`TSANet-integration-demo/.../JacksonPrimitiveNullConfigTest.java`, +43): same for the demo app, with `WebEnvironment.NONE` and the demo-specific javadoc (setup/sync off vs CLI off) — traces to the same requirement. ✅
- Hunk 3 (`OAuthTokenGateway.java`): import add + `new ObjectMapper()` → `JsonMapper.builder().build()` — traces to "OAuthTokenGateway ... becomes JsonMapper.builder().build()". ✅
- Hunk 4 (`WebhookPayloadParser.java`): same two-line change — traces to "WebhookPayloadParser ... becomes JsonMapper.builder().build()". ✅

All four hunks trace. No incidental tidying. Import ordering is unchanged from the existing convention (new import placed adjacent to the sibling `tools.jackson.databind` imports).

**Correctness:**
- Trace: `mapper.readValue("{\"count\":null}", Probe.class)` where `Probe(int count)`. With `FAIL_ON_NULL_FOR_PRIMITIVES` off (as the main yml sets), Jackson binds the explicit null to the primitive default `0`; `assertThat(probe.count()).isZero()` passes. Remove the yml line and Jackson 3's default (feature on) throws `MismatchedInput` — the red probes in the PR confirm both apps go red. The test genuinely pins the line rather than trivially passing.
- The `spring.config.location` design is the load-bearing risk. The concern — that the test-resources yml shadows the main one so a plain `@SpringBootTest` would pin nothing — is real, and the mitigation (explicit ordered locations, later overrides earlier) is Spring Boot config-data behavior. The green run with all test overrides still applied (mock URL, tmp storage) is the receipt that the layering works. I have not independently re-run the reactor build this session; I am relying on the cited receipts.
- Library change: both mappers call only `readTree` (per the PR grep), so no primitive binding path runs through them and the `FAIL_ON_NULL_FOR_PRIMITIVES`-on default carried unchanged cannot fire. `JsonMapper.builder().build()` with no feature calls is the documented Jackson 3 equivalent of the default `ObjectMapper`. Behavior-neutral as claimed.
- Edge: `readValue` in the tests declares checked-exception-free signature in Jackson 3 (unchecked `JacksonException`), so no `throws` clause is needed — consistent with the test compiling green.

**Security:**
None. No secrets, no external input paths changed, no logging added. The `spring.config.location` values are static literals scoped to test working directories.

**Regression risk:**
- Consumers of the two library mappers: `WebhookPayloadParser` (used by webhook ingestion) and `OAuthTokenGateway` (token acquisition). Both are static fields swapped to an equivalent constructor; existing `WebhookPayloadParserTest` (2), `OAuthTokenGatewayTest` (1), `TokenManagerTest` are cited green. The `readTree`-only usage means the swap is inert.
- The new tests are additive; they cannot break existing tests. Their one non-obvious failure mode — IDE runs with a non-module working directory — fails loudly at context start (`ConfigDataLocationNotFoundException`), never a false pass, and surefire is the stated contract. That's an accepted operational note, not a regression.
- No positional contract, vocabulary, schema, or guard change — pre-PR artifacts detector fired nothing, consistent with my read.

**Recommendation:** Merge — the change is minimal, traces cleanly, and the one load-bearing design point (config-location layering) is backed by red/green receipts.

Note: PR is still marked Draft — mark ready before merging. I have not re-run the build this session, so my correctness read rests on the cited receipts rather than an independent run.

---

shawn-tsanet (commented):
## QA review of Build the two library mappers with JsonMapper and test the primitive-null setting in both apps

**Verdict:** ✅ Ready to merge

Separate-session QA against `4c130cd` vs `origin/main` `fa9cd89`. Every receipt is a check run this session, independent of the PR body's.

### Minimal-Diff Rule
- Pass. Four hunks: two initializer-plus-import swaps in the library, two new test classes. Each traces to `tsanetgit/Connect_SDK#87`'s two items. No yml, dependency or existing-test change.

### Correctness
- **Green and red reproduced.** On a fresh archive of the PR head, `mvn -pl connect-library,<app> -am test -Dtest=JacksonPrimitiveNullConfigTest` passes in both apps. With the `fail-on-null-for-primitives` line deleted from each main `application.yml`, the same reactor run fails on exactly that test with `MismatchedInput: Cannot map null into type int (set DeserializationFeature.FAIL_ON_NULL_FOR_PRIMITIVES to 'false' to allow)`, app and demo alike. Files restored. So the test pins the line through the Boot-injected mapper, not a proxy.
- **The `spring.config.location` layering does not activate anything the plain test context leaves off.** Key-by-key diff of main vs test yml on the PR head: in the app, every main key that could act (`webhook.enabled`, `api.base-url`, `storage`, `cli`) is overridden by the test yml, and `WebhookReceiverController` is `@ConditionalOnProperty` on `tsanet.webhook.enabled`. In the demo, `demo.setup`, `scenarios.run-on-startup` and `sync` are overridden to off; `PollingConfiguration` is conditional on `tsanet.sync.enabled`; the one main-only key, `tsanet.connect-db`, is consumed solely by `ConnectApiDatabaseConfiguration`, which is `@ConditionalOnProperty` on `tsanet.demo.setup.enabled=true`, so the whole configuration is skipped. `server.port` is inert under `web-application-type: none` (app test yml) and `WebEnvironment.NONE` (demo test).
- **No profile-specific or imported config exists** for the explicit locations to change: each resources dir holds one `application.yml`, no `application-*.yml`, no `config/` dir, no `spring.config.import`.
- **Library swap is inert.** Both classes call only `readTree` (one call each at `WebhookPayloadParser.java:16` and `OAuthTokenGateway.java:59`); `WebhookPayloadDto` is four `String` components; `expires_in` is read via `asLong(3600)` on the tree. `JsonMapper.builder().build()` with no features set carries Jackson 3 defaults, same as `new ObjectMapper()`. `git grep 'new ObjectMapper()'` on the PR head is empty across all Java.

### Security
- No concerns.

### Regressions
- None. The new tests are additive; the two library fields keep name, type and visibility. The IDE working-directory caveat fails loudly at context start, as the PR states.

### Operational
- None. The test-side `@SpringBootTest` boots one extra context per module; build-time cost only.

### Nits
- The PR is still a draft. Mark it ready before merging.

